PR #3, opened 2021-08-20
task done

illiarerun:
Could've added shape initializations outside of constructors, but linter didn't allow using "useless constructors"

### DIFF
--- a/src/figuresClasses.ts
+++ b/src/figuresClasses.ts
@@ -1,19 +1,70 @@
+enum Color {
+  Red = 'red',
+  Green = 'green',
+  Blue = 'blue'
+}
+
+enum Shape {
+  Triangle = 'triangle',
+  Circle = 'circle',
+  Rectangle = 'rectangle'
+}
+
 export interface Figure {
-
+  shape: Shape,
+  color: Color,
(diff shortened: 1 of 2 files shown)


VeronikaLushchik (approved):
Great!

src/figuresClasses.ts
```diff
@@ -1,19 +1,70 @@
...
-export class Rectangle {
+export class Circle implements Figure {
+  shape: Shape;
+
+  constructor(public color: Color, public radius: number) {
+    this.shape = Shape.Circle;
+  }
 
+  getArea(): number {
+    return 3.14 * (this.radius ** 2);
```
VinnieJ-2k20, line 47: JS has Math.PI constant for more accurate calculations.

VinnieJ-2k20 (approved):
Good job 👍

src/figuresClasses.ts
illiarerun: Test didn't work with Math.PI (that was the first thing i tried 😄), i guess test were developed with exactly 3.14.

VinnieJ-2k20: 🤕

PR #15, opened 2021-08-23
Solution

Anna-Tkachenko:


src/figuresClasses.ts
```diff
@@ -1,19 +1,87 @@
...
 }
 
-export class Triangle {
+export interface Figure {
+  shape: Shape;
+  color: Color;
+  getArea(): number;
+}
 
+function checkSizes(...sizes: number[]): void {
```
TarasHoliuk, line 19: smart solution 👍

src/figuresClasses.ts
```diff
@@ -1,19 +1,87 @@
...
+  constructor(
+    public color: Color,
+    public width: number,
+    public height: number,
+  ) {
+    checkSizes(width, height);
+  }
+
+  getArea(): number {
+    return this.width * this.height;
```
TarasHoliuk, line 80: I guess you should round down to hundreds all areas of the figures.
If we change input values we can get results like 45,3333333

src/figuresClasses.ts
```diff
@@ -1,19 +1,87 @@
-export interface Figure {
+enum Shape {
```
TarasHoliuk, line 1: good decision to use enum here 👍

TarasHoliuk, line 1: good decision to use enum here 👍

TarasHoliuk (approved):
good job 🥇

src/figuresClasses.ts
Anna-Tkachenko: I did it in getInfo function

src/figuresClasses.ts
```diff
@@ -1,19 +1,87 @@
...
+
+  constructor(
+    public color: Color,
+    public radius: number,
+  ) {
+    checkSizes(radius);
+  }
 
+  getArea(): number {
+    return 3.14 * (this.radius ** 2);
```
VinnieJ-2k20, line 64: JS has Math.PI constant for more accurate calculations.

VinnieJ-2k20 (approved):
Good job 👍

PR #17, opened 2021-08-23
sollution

dima-tkachenko:


src/figuresClasses.ts
```diff
@@ -1,19 +1,88 @@
...
+    public color: Color,
+    public radius: number,
+  ) {
+    if (radius <= 0) {
+      throw new Error('Radius should be positive');
+    }
+  }
+
+  getArea(): number {
+    const pi = +Math.PI.toFixed(2);
```
sasha-morozov, line 59: In my opinion, the variable "pi" might be redundant here. You may use '3.14' instead since you're rounding up to hundreds.

sasha-morozov (approved):
Overall, the code is good. Left one suggestion above)

src/figuresClasses.ts
```diff
@@ -1,19 +1,88 @@
...
+    public c: number,
+  ) {
+    if (a <= 0 || b <= 0 || c <= 0) {
+      throw new Error('Side should be positive');
+    }
+
+    if (a + b <= c
+      || a + c <= b
+      || c + b <= a) {
+      throw new Error('Your sides is not create triangle');
```
VinnieJ-2k20, line 31: Use correct English
```suggestion
      throw new Error('Your sides don/'t create triangle');
```

src/figuresClasses.ts
VinnieJ-2k20: Agree, it is better to round the number just once in the end.

VinnieJ-2k20 (approved):
Good job 👍 
Added some comments on code improvement.

PR #19, opened 2021-08-23
solution

Timoszkin:


src/figuresClasses.ts
```diff
@@ -1,19 +1,68 @@
-export interface Figure {
+type Shape = 'triangle' | 'circle' | 'rectangle';
+type Color = 'red' | 'green' | 'blue';
```
vkryvytskyy, line 2: create enums with these properties

vkryvytskyy (approved):

PR #20, opened 2021-08-24
solution

DmytroMykoliv:


sasha-morozov (approved):
The code looks good! Seems that there is nothing redundant.

src/figuresClasses.ts
```diff
@@ -1,19 +1,76 @@
...
 
+export interface Figure {
+  shape: Shape;
+  color: Color;
+  getArea(): number;
 }
 
-export class Triangle {
+export class Triangle implements Figure {
+  shape: Shape ='triangle';
```
VinnieJ-2k20, line 11:
```suggestion
  shape: Shape = 'triangle';
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,76 @@
...
+    public color: Color,
+    public radius: number,
+  ) {
+    if (this.radius <= 0) {
+      throw new Error('Invalid value');
+    }
+  }
+
+  getArea(): number {
+    return +(3.14 * this.radius ** 2).toFixed(2);
```
VinnieJ-2k20, line 52: For more accurate calculations JS has Math.PI constant
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/PI

VinnieJ-2k20 (approved):
Good job 👍

src/figuresClasses.ts
DmytroMykoliv: I used Math.PI but it does a wrong calculation and I changed it to 3.14

src/figuresClasses.ts
DmytroMykoliv: Thank you, missed it ))

PR #34, opened 2021-08-26
Figures classes

kristof22122:


src/figuresClasses.ts
```diff
@@ -1,19 +1,91 @@
...
+  getArea(): number {
+    const halfArea: number = (this.x + this.y + this.z) / 2;
+    const area: number = (
+      halfArea
+      * (halfArea - this.x)
+      * (halfArea - this.y)
+      * (halfArea - this.z)
+    ) ** (1 / 2);
+
+    return +area.toFixed(2);
```
varseniuk, line 50:
```suggestion
```
Instead of this, you can simply use just `Math.sqrt` method

src/figuresClasses.ts
```diff
@@ -1,19 +1,91 @@
...
+    public color: Color,
+    public radius: number,
+  ) {
+    if (this.radius <= 0) {
+      throw new Error('size less than 0');
+    }
+  }
 
+  getArea(): number {
+    return +(3.14 * (this.radius) ** 2).toFixed(2);
```
varseniuk, line 67:
```suggestion
    return 3.14 * (this.radius) ** 2);
```
You can use `toFixed` only once in the end in your `getInfo` function

varseniuk (approved):
Good job! 
Left some comments

PR #40, opened 2021-08-27
TypeScript - Classes - Figures

george-w-boo:


src/figuresClasses.ts
```diff
@@ -1,19 +1,75 @@
 export interface Figure {
-
+  color: string;
+  shape: string;
```
varseniuk, line 3:
```suggestion
  color: string;
  shape: string;
```
Would be great to create enums for colors and shapes

src/figuresClasses.ts
```diff
@@ -1,19 +1,75 @@
...
 }
 
-export class Triangle {
+export class Triangle implements Figure {
+  constructor(
+    public color: string,
+    public a: number,
+    public b: number,
+    public c: number,
+    public shape: string = 'triangle',
```
varseniuk, line 13: Use enums instead of strings

src/figuresClasses.ts
```diff
@@ -1,19 +1,75 @@
...
+    return this.a * this.b;
+  }
 }
 
-export function getInfo(figure) {
+export function getInfo(figure: Figure): string {
+  const result = `A ${figure.color} ${figure.shape} - `
+  + `${(figure.getArea() % 1 === 0)
+    ? figure.getArea()
+    : figure.getArea().toFixed(2)}`;
```
varseniuk, line 72:
```suggestion
  const result = `A ${figure.color} ${figure.shape} - ${figure.getArea().toFixed(2)}`
```
I think you can simply it here without ternaries

varseniuk (changes requested):
All is good, but let's use enums

src/figuresClasses.ts
george-w-boo: It's a great idea, thank you :)

src/figuresClasses.ts
george-w-boo: Yes-либы )) Take a look at the [screenshot](https://clip2net.com/s/4dafMRl) below:

varseniuk: I suppose it's because you need number (add `+` sign)
```
const result = `A ${figure.color} ${figure.shape} - ${+figure.getArea().toFixed(2)}`
```

varseniuk (approved):
👍 
Left one comment

src/figuresClasses.ts
george-w-boo: Wow, you were right. Thank you :)

PR #44, opened 2021-08-29
Solution Figures Classes

AndriyNikorych:


Andrii-Pinchuk (approved):
👍

src/figuresClasses.ts
```diff
@@ -1,19 +1,80 @@
 export interface Figure {
-
+  shape: string;
+  color: string;
```
varseniuk, line 3:
```suggestion
  shape: string;
  color: string;
```
Use enums for color and shape

src/figuresClasses.ts
```diff
@@ -1,19 +1,80 @@
...
+  getArea(): number;
 }
 
-export class Triangle {
+export class Triangle implements Figure {
+  public shape = 'triangle';
```
varseniuk, line 8: Use enum instead of string

src/figuresClasses.ts
```diff
@@ -1,19 +1,80 @@
...
+  }
 
+  getArea(): number {
+    const p: number = (this.a + this.b + this.c) / 2; // half of perimeter
+    const area: number = (p
+      * (p - this.a)
+      * (p - this.b)
+      * (p - this.c)) ** (1 / 2); // Heron's formula
+
+    return Number(area.toFixed(2));
```
varseniuk, line 35:
```suggestion
    return Number(area);
```
You can use `toFixed` only once at the end in your `getInfo` function

varseniuk (changes requested):
All is good but let's optimise it a bit

varseniuk (approved):
👍

PR #45, opened 2021-08-29
solution

RomSmile:


src/figuresClasses.ts
```diff
@@ -1,19 +1,73 @@
-export interface Figure {
 
+export interface Figure {
+  shape: 'triangle' | 'circle' | 'rectangle';
+  color: 'red' | 'green' | 'blue';
```
varseniuk, line 4:
```suggestion
  shape: 'triangle' | 'circle' | 'rectangle';
  color: 'red' | 'green' | 'blue';
```
Use enums for shape and color

src/figuresClasses.ts
```diff
@@ -1,19 +1,73 @@
...
 export class Triangle {
+  shape = 'triangle';
+
+  constructor(
+    public color: string,
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    if (
```
varseniuk, line 17: How about the check if some of the sides is <=0?

src/figuresClasses.ts
```diff
@@ -1,19 +1,73 @@
...
+      || this.c + this.a <= this.b
+    ) {
+      throw new Error('Error');
+    }
+  }
 
+  getArea(): number {
+    const peremeter: number = (this.a + this.b + this.c) / 2;
+
+    // eslint-disable-next-line max-len
```
varseniuk, line 29:
```suggestion
```
Let's not disable es-lint and try to apply the proper formatting

src/figuresClasses.ts
```diff
@@ -1,19 +1,73 @@
...
+    return Math.floor(Math.sqrt(peremeter * (peremeter - this.a) * (peremeter - this.b) * (peremeter - this.c)) * 100) / 100;
+  }
 }
 
 export class Circle {
+  shape = 'circle';
```
varseniuk, line 35:
```suggestion
  shape = 'circle';
```
Use enum instead of string

src/figuresClasses.ts
```diff
@@ -1,19 +1,73 @@
...
+  }
 
+  getArea(): number {
+    return Math.floor(this.width * this.height);
+  }
 }
 
-export function getInfo(figure) {
+export function getInfo(figure: Figure): string {
+  const s = figure.getArea();
```
varseniuk, line 70:
```suggestion
  const s = figure.getArea();
```

s? What `s` suppose to mean?

varseniuk (changes requested):

src/figuresClasses.ts
```diff
@@ -1,19 +1,90 @@
-export interface Figure {
+enum ShapeTypes {
+  triangle,
+  circle,
+  rectangle,
+}
```
varseniuk, line 5:
```suggestion
enum ShapeTypes {
  triangle = 'triangle,
  circle = 'circle',
  rectangle = 'rectangle',
}
```
String enums should be written like this.
Then you can write, for example:
```
shape = ShapeTypes.triangle
```

varseniuk (changes requested):
Let's use enums properly

varseniuk (approved):

PR #96, opened 2021-10-07
solution

Bogdansgit:


src/figuresClasses.ts
```diff
@@ -1,19 +1,77 @@
-export interface Figure {
 
+type Color = 'red' | 'green' | 'blue';
+
+export interface Figure {
+  shape: string;
```
aliceblunt8, line 5: create enum for Shape

```suggestion
  shape: Shape;
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,77 @@
...
+  color: Color;
+  getArea: () => number;
```
aliceblunt8, line 7:
```suggestion
  getArea(): number;
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,77 @@
...
 }
 
-export class Triangle {
+export class Triangle implements Figure {
+  public shape: string = 'triangle';
```
aliceblunt8, line 11:
```suggestion
  public shape: Shape = Shape.Triangle;
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,77 @@
...
 
-}
+  constructor(
+    public color: Color,
+    private a: number,
+    private b: number,
+    private c: number,
+  ) {
+    if (a <= 0 || b <= 0 || c <= 0) {
+      throw new Error();
```
aliceblunt8, line 20: add error messages everywhere
```suggestion
      throw new Error('error message');
```

src/figuresClasses.ts
aliceblunt8: not fixed

PR #114, opened 2021-10-29
done

DammneD:


src/figuresClasses.ts
```diff
@@ -1,19 +1,86 @@
...
+
+  if (args.length === 3) {
+    if (a < (b + c) && b < (a + c) && c < (a + b)) {
+      return false;
+    }
 
+    return true;
+  }
+
+  return args.some((side: number) => side <= 0);
```
VinnieJ-2k20, line 21: function is composed with different pieces of logic. Better to perform checks in place - in the constructors. Or separate into separate function for triangle.

VinnieJ-2k20 (changes requested):
Almost done, some small fixes left.

VinnieJ-2k20 (approved):

PR #118, opened 2021-10-31
Task was solved

Igor-MateAcademy:


ivarovskiy (approved):
solution looks great! 🔥

src/figuresClasses.ts
```diff
@@ -1,19 +1,100 @@
...
+  public shape: string;
 
-}
+  sizes: number[];
+
+  constructor(
+    public color: string,
+    ...args: number[]
+  ) {
+    this.shape = 'triangle';
```
solaryasha, line 18: You could use it without `this.shape`
just `public shape = 'triangle'`

src/figuresClasses.ts
```diff
@@ -1,19 +1,100 @@
...
+    this.color = color;
```
solaryasha, line 19: You already done this, when call this field in constructor. This is code duplication

solaryasha (changes requested):

src/figuresClasses.ts
```diff
@@ -1,19 +1,96 @@
...
 
-export class Rectangle {
+export class Circle implements Figure {
+  public shape = 'circle';
+
+  sizes: number[];
+
+  constructor(
+    public color: Color,
+    ...args: number[]
```
VinnieJ-2k20, line 53: No need to use rest operator here, there's only `radius` argument.

src/figuresClasses.ts
```diff
@@ -1,19 +1,96 @@
...
-export class Triangle {
+export class Triangle implements Figure {
+  public shape = 'triangle';
 
-}
+  sizes: number[];
+
+  constructor(
+    public color: Color,
+    ...args: number[]
```
VinnieJ-2k20, line 18: Don't use rest here, there are fixed number of arguments for every shape.

VinnieJ-2k20 (changes requested):
Remove usage of rest operator.

src/figuresClasses.ts
```diff
@@ -1,19 +1,92 @@
...
+export class Circle implements Figure {
+  shape = 'circle';
+
+  constructor(
+    public color: Color,
+    public radius: number,
+  ) {
+    this.radius = radius;
+
+    if (!(radius > 0)) {
```
VinnieJ-2k20, line 53:
```suggestion
    if (radius <= 0) {
```

VinnieJ-2k20 (approved):

PR #136, opened 2021-11-20
Implement three figures classes

dmtrhrytsak:


src/figuresClasses.ts
```diff
@@ -1,19 +1,75 @@
-export interface Figure {
+type Shape = 'triangle' | 'circle' | 'rectangle';
+type Color = 'red' | 'green' | 'blue';
```
illiarerun, line 2: Use enum instead

src/figuresClasses.ts
```diff
@@ -1,19 +1,75 @@
...
 
-}
+  constructor(
+    public color: Color,
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    if (a <= 0 || b <= 0 || c <= 0) {
+      throw new Error('The lenght of the side cannot be negative');
```
illiarerun, line 20: ))0)
```suggestion
      throw new Error('The length of the side cannot be negative');
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,75 @@
...
+    }
+
+    if (a + b <= c || c + b <= a || c + a <= b) {
+      throw new Error('Such triangle doesnot exist');
```
illiarerun, line 24: 🔥 
```suggestion
      throw new Error('Such triangle does not exist');
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,75 @@
...
+export class Rectangle implements Figure {
+  shape: Shape = 'rectangle';
+
+  constructor(
+    public color: Color,
+    public width: number,
+    public height: number,
+  ) {
+    if (width <= 0 || height <= 0) {
+      throw new Error('The lenght of the side cannot be negative');
```
illiarerun, line 64:
```suggestion
      throw new Error('The length of the side cannot be negative');
```

illiarerun (approved):
Good job 🔥

PR #147, opened 2021-11-30
Solution

Veronika-Korona:


src/figuresClasses.ts
```diff
@@ -1,19 +1,81 @@
-export interface Figure {
-
+enum Shape {
+  Triangle = 'triangle',
+  Circle = 'circle',
+  Rectangle = 'rectangle',
 }
 
-export class Triangle {
+type Color = 'red' | 'green' | 'blue';
```
Vitaliy-0, line 7: you don't need this types, what if color will be pink?

src/figuresClasses.ts
```diff
@@ -1,19 +1,81 @@
...
+      throw new Error('Triangle side can not be less than 0');
+    }
+
+    if (a >= (b + c) || b >= (a + c) || c >= (a + b)) {
+      throw new Error('Sides 1, 2 and 3 can not form a triangle');
+    }
+  }
 
+  getArea():number {
+    const halfPerimeter = (this.a + this.b + this.c) / 2;
```
Vitaliy-0, line 34: halfSquare will be better

Vitaliy-0 (approved):
Good work

src/figuresClasses.ts
Veronika-Korona: Thanks for review.
But in task is written about 3 specific colors. I will wait for mentor opinion))

aionova: yes, it's true. Colors are limited to these 3. You can use type as you did, or use enum instead

aionova (approved):
GJ!

PR #152, opened 2021-11-30
solution

mariia-bila:


volodymyr-soltys97 (approved):
LGTM!

src/figuresClasses.ts
```diff
@@ -1,19 +1,75 @@
-export interface Figure {
+type Shape = 'triangle' | 'circle' | 'rectangle';
+type Color = 'red' | 'green' | 'blue';
```
aionova, line 2: it would be great to make these enums

aionova (approved):
GJ!

PR #158, opened 2021-11-30
Solution

alinaShchavinska:


dlugash (approved):
It is not obligatory to write 'public' because properties are public by default :)

dlugash (approved):
It is not obligatory to write 'public' because properties are public by default :)

dlugash (approved):
It is not obligatory to write 'public' because properties are public by default :)

src/figuresClasses.ts
```diff
@@ -1,19 +1,88 @@
...
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    if (a <= 0 || b <= 0 || c <= 0) {
+      throw new Error('Invalid value! Please, write another positive number');
+    }
+
+    if (a + b <= c || c + b <= a || c + a <= b) {
+      throw new Error('Incorrect value! Please, write another values');
```
ihor-jpeg, line 34:
```suggestion
      throw new Error('Incorrect value! Please, write other values');
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,88 @@
...
+    }
+  }
+
+  getArea(): number {
+    const p: number = (this.a + this.b + this.c) / 2;
+    const squareTr: number
+    = Math.sqrt(p * (p - this.a) * (p - this.b) * (p - this.c));
```
ihor-jpeg, line 41: do not break line before or after `=`
```suggestion
    const squareTr: number = Math.sqrt(
      p * (p - this.a) * (p - this.b) * (p - this.c)
    );
```

ihor-jpeg (approved):
Good work. Let's fix those few comments 👍

PR #162, opened 2021-11-30
solution

MagicBox15:


SerhiiKirik (approved):
Well done 🚀

src/figuresClasses.ts
```diff
@@ -1,19 +1,81 @@
-export interface Figure {
+type Shape = 'triangle' | 'circle' | 'rectangle';
+type Color = 'red' | 'green' | 'blue';
```
aionova, line 2: you could also use enum instead

aionova (approved):
GJ!

PR #163, opened 2021-11-30
Solution

aldmpr:


src/figuresClasses.ts
```diff
@@ -1,19 +1,75 @@
 export interface Figure {
-
+  shape: 'triangle' | 'circle' | 'rectangle';
```
IrynaHerasymchuk, line 2: You can also use enum

IrynaHerasymchuk (approved):
Pretty good!

src/figuresClasses.ts
ihor-jpeg: enum makes a bit more sense here

src/figuresClasses.ts
```diff
@@ -1,19 +1,75 @@
...
+  color: 'red' | 'green' | 'blue';
+  getArea(): number;
 }
 
-export class Triangle {
+export class Triangle implements Figure {
+  shape: Figure['shape'] = 'triangle';
```
ihor-jpeg, line 8: if you use enum you can make this a lot simpler

ihor-jpeg (approved):
well done! try using enum

PR #169, opened 2021-12-01
solution: ts_figures_classes 

lanebx:


Veronika-Korona (approved):
GJ!

src/figuresClasses.ts
```diff
@@ -1,19 +1,73 @@
...
+type ColorType = 'red' | 'green' | 'blue';
 
+export interface Figure {
+  shape: string,
+  color: ColorType,
+  getArea(): number;
 }
 
 export class Triangle {
+  shape = 'triangle';
```
ihor-jpeg, line 10: enum is much better when you need to hardcode values, because in the case when you need to change the value, you need to do it in only one place

ihor-jpeg (approved):
Very good work! Nice error massages, they are very informative 🔥

PR #170, opened 2021-12-01
completed figures

i-stanko:


Veronika-Korona (approved):
GJ!

src/figuresClasses.ts
```diff
@@ -1,19 +1,75 @@
...
 
+export interface Figure {
+  shape: string;
+  color: string;
+  getArea(): number;
 }
 
-export class Triangle {
+export class Triangle implements Figure {
+  shape = 'triangle';
```
ihor-jpeg, line 10: make an enum for this. try to avoid using hardcoded values

src/figuresClasses.ts
```diff
@@ -1,19 +1,75 @@
...
+    }
+  }
 
-export class Circle {
+  getArea(): number {
+    const semiPerimetr = (this.a + this.b + this.c) / 2;
+    const area = Math.sqrt(semiPerimetr
+      * (semiPerimetr - this.a)
+      * (semiPerimetr - this.b)
+      * (semiPerimetr - this.c));
```
ihor-jpeg, line 32:
```suggestion
    const area = Math.sqrt(
      semiPerimetr
      * (semiPerimetr - this.a)
      * (semiPerimetr - this.b)
      * (semiPerimetr - this.c)
    );
```

ihor-jpeg (changes requested):

aionova (approved):

PR #199, opened 2022-01-11
Develop

Bohdan-Maksymiuk:


src/figuresClasses.ts
```diff
@@ -1,19 +1,79 @@
...
+    if (width <= 0 || height <= 0) {
+      throw new Error('Width and height should be a positive numbers');
+    }
+  }
+
+  getArea(): number {
+    const square: number = this.width * this.height;
+
+    return square;
+  }
```
andrii-koptiev, line 74: I think you should add rounding down to the hundredth also here

andrii-koptiev (approved):

src/figuresClasses.ts
```diff
@@ -1,19 +1,79 @@
-export interface Figure {
+type Shape = 'triangle' | 'circle' | 'rectangle';
```
danheim, line 1: it has to be an enum

src/figuresClasses.ts
```diff
@@ -1,19 +1,79 @@
...
+      * (semiPerimeter - this.b) * (semiPerimeter - this.c);
+    const square: number = Math.sqrt(underTheRoot);
+
+    return Math.floor(square * 100) / 100;
+  }
 }
 
-export class Rectangle {
+export class Circle implements Figure {
+  shape: Shape = 'circle';
```
danheim, line 39: use Enum here, avoid strings

src/figuresClasses.ts
```diff
@@ -1,19 +1,79 @@
...
+    public color: Color,
+    public radius: number,
+  ) {
+    if (radius <= 0) {
+      throw new Error('Radius should be a positive numbers');
+    }
+  }
 
+  getArea(): number {
+    const square: number = Math.PI * this.radius ** 2;
```
danheim, line 51: avoid types here, because the result of your calc would be a number
```suggestion
    const square = Math.PI * this.radius ** 2;
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,79 @@
...
+    public width: number,
+    public height: number,
+  ) {
+    if (width <= 0 || height <= 0) {
+      throw new Error('Width and height should be a positive numbers');
+    }
+  }
+
+  getArea(): number {
+    const square: number = this.width * this.height;
```
danheim, line 71: same

danheim (changes requested):

danheim (approved):

PR #220, opened 2022-01-11
solution_v1.0

Rozdorozhnii:


src/figuresClasses.ts
```diff
@@ -1,19 +1,77 @@
-export interface Figure {
+type Shape = 'triangle' | 'circle' | 'rectangle';
+type Color = 'red' | 'green' | 'blue';
```
Mikedacheka, line 2: more colors?
In my opinion, better leave it with string type.

src/figuresClasses.ts
```diff
@@ -1,19 +1,77 @@
-export interface Figure {
+type Shape = 'triangle' | 'circle' | 'rectangle';
```
Mikedacheka, line 1: Same situation colleague.
Maybe ,it will be better, when Class have it by own field.
In this case , for better code support - enum.
Will discuss it later, comrade.

Mikedacheka (approved):
Great solution!

src/figuresClasses.ts
Olexandr6: Agree, it's better to use enum here

src/figuresClasses.ts
```diff
@@ -1,19 +1,77 @@
...
+    if ((this.a + this.b) <= this.c
+      || (this.a + this.c) <= this.b
+      || (this.b + this.c) <= this.a
+    ) {
+      throw new Error('Entered data is invalid');
+    }
+  }
 
+  getArea(): number {
+    const p: number = (this.a + this.b + this.c) / 2;
```
Olexandr6, line 33: What is `p`? Rename it

src/figuresClasses.ts
```diff
@@ -1,19 +1,77 @@
...
+      * (p - this.a) * (p - this.b) * (p - this.c)) * 100) / 100;
+  }
 }
 
 export class Circle {
+  readonly shape: Shape = 'circle';
+
+  constructor(
+    readonly color: Color,
+    private a: number,
```
Olexandr6, line 45:
```suggestion
    private radius: number,
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,77 @@
...
+  ) {
+    if (!(this.a > 0)) {
```
Olexandr6, line 47: You can simplify this condition
```suggestion
    if (this.a < 0) {
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,77 @@
...
 export class Rectangle {
+  readonly shape: Shape = 'rectangle';
 
-}
+  constructor(
+    readonly color: Color,
+    private a: number,
+    private b: number,
+  ) {
+    if (!(this.a > 0) || !(this.b > 0)) {
```
Olexandr6, line 65: Here is the same case...

src/figuresClasses.ts
```diff
@@ -1,19 +1,77 @@
...
+export class Triangle implements Figure {
+  readonly shape: Shape = 'triangle';
+
+  constructor(
+    readonly color: Color,
+    private a: number,
+    private b: number,
+    private c: number,
+  ) {
+    if (!(this.a > 0) || !(this.b > 0) || !(this.c > 0)) {
```
Olexandr6, line 20: Simplify this condition

src/figuresClasses.ts
```diff
@@ -1,19 +1,77 @@
-export interface Figure {
+type Shape = 'triangle' | 'circle' | 'rectangle';
+type Color = 'red' | 'green' | 'blue';
+type GetArea = () => number;
 
+export interface Figure {
+  shape: Shape;
+  color: Color;
+  getArea: GetArea;
```
Olexandr6, line 8: You can wtire it sorten
```suggestion
export interface Figure {
  shape: Shape;
  color: Color;
  getArea(): number;
```

Olexandr6 (changes requested):
Check comments 👇

src/figuresClasses.ts
Rozdorozhnii: I wrote this harder code to train himself read and write a bit difficult code. I'm not sure will I need it in future, hope not, but think better to be ready for it

src/figuresClasses.ts
Rozdorozhnii: I wrote this harder code to train himself read and write a bit difficult code. I'm not sure will I need it in future, hope not, but think better to be ready for it

src/figuresClasses.ts
Rozdorozhnii: I wrote this harder code to train himself read and write a bit difficult code. I'm not sure will I need it in future, hope not, but think better to be ready for it

Olexandr6 (approved):

PR #223, opened 2022-01-12
Solution

YevheniiKorzun:


Mikedacheka (approved):
awesome

src/figuresClasses.ts
```diff
@@ -1,19 +1,115 @@
+type Shape = 'triangle' | 'circle' | 'rectangle';
+type Color = 'red' | 'green' | 'blue';
```
danheim, line 2: Create enum for Shape and Color

danheim (changes requested):

danheim (approved):

PR #235, opened 2022-02-04
first try

krasnobokydima:


013g (approved):
Well done

src/figuresClasses.ts
```diff
@@ -1,19 +1,101 @@
 export interface Figure {
-
+  shape: string
+  color: string
+  getArea(): number | never
+  sides?: number[]
+  radius?: number
```
Anna-Tkachenko, line 6: In Figure interface you should specify only 3 main properties, sides and radius should be created for specific figures (sides only for Triangles, radius only for Circles)

src/figuresClasses.ts
```diff
@@ -1,19 +1,101 @@
...
+    let p = a + b + c;
 
+    const errorValue = sides.some((elem: number) => elem <= 0
+        || p - elem <= elem);
+
+    if (errorValue) {
+      throw new Error("sides 1, 2 and 3 can't form a triangle");
+    }
+
+    this.getArea = (): number | never => {
```
Anna-Tkachenko, line 33: getArea method should always return the number of area, why did you use never type here?

src/figuresClasses.ts
```diff
@@ -1,19 +1,101 @@
...
+  constructor(color: string, radius: number) {
+    this.shape = 'circle';
+    this.color = color;
+    this.radius = radius;
+
+    if (radius <= 0) {
+      throw new Error('radius must be greater than 0');
+    }
+
+    this.getArea = (): number | never => {
```
Anna-Tkachenko, line 61: Same here with never

Anna-Tkachenko (changes requested):

src/figuresClasses.ts
```diff
@@ -1,19 +1,97 @@
...
+
+type Shape = 'triangle' | 'circle' | 'rectangle';
+type Color = 'blue' | 'green' | 'red';
 
+export interface Figure {
+  shape: Shape
+  color: Color
+  getArea(): number
+  radius?: number
+  sides?: number[]
```
Anna-Tkachenko, line 11: In Figure interface you should specify only those properties that should be for all figures
Remove radius and sides from this interface and add it to the needed figures

Anna-Tkachenko (approved):

Anna-Tkachenko (changes requested):
Classes should implement Figure interface

Anna-Tkachenko (approved):

PR #236, opened 2022-02-07
solution

dmytroosypiuk:


src/figuresClasses.ts
```diff
@@ -1,19 +1,85 @@
-export interface Figure {
+type Color = 'red' | 'green' | 'blue';
 
+export interface Figure {
+  shape: string;
```
Anna-Tkachenko, line 4: It's better to create some type for defining values that can be in shape property

src/figuresClasses.ts
```diff
@@ -1,19 +1,85 @@
...
+  public shape: string;
 
-}
+  constructor(
+    public color: Color,
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    this.shape = this.constructor.name.toLowerCase();
```
Anna-Tkachenko, line 18: It's redundant

Anna-Tkachenko (approved):
Good job
Just need some fixes with shape property

PR #240, opened 2022-02-08
figures

Oleksii-Pinchuk:


mtctxd (approved):
Ніюи все класно

src/figuresClasses.ts
```diff
@@ -1,19 +1,84 @@
-export interface Figure {
+type Color = 'red' | 'green' | 'blue';
 
+export interface Figure {
+  shape: string;
```
Anna-Tkachenko, line 4: It's better to create some type for defining values for shape property

src/figuresClasses.ts
```diff
@@ -1,19 +1,84 @@
...
+      throw new Error('Radius length should be more than 0');
+    }
+  }
+
+  getArea(): number {
+    return Math.floor(Math.PI
+      * this.radius
+      * this.radius
+      * 100)
+      / 100;
```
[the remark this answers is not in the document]
Anna-Tkachenko: agree

src/figuresClasses.ts
```diff
@@ -1,19 +1,84 @@
...
+    this.shape = 'rectangle';
 
-export function getInfo(figure) {
+    if (a <= 0 || b <= 0) {
+      throw new Error('Sides length should be more than 0');
+    }
+  }
+
+  getArea(): number {
+    return Math.floor(this.a * this.b * 100) / 100;
```
Anna-Tkachenko, line 78: Don't repeat yourself: Math.floor( ... * 100) / 100 repeated 3 times in your code, you can create separate function for this operation

Anna-Tkachenko (approved):

PR #241, opened 2022-02-09
ez

013g:


pogrebnyakp (approved):

src/figuresClasses.ts
```diff
@@ -1,19 +1,83 @@
...
 export interface Figure {
-
+  shape:Shape;
+  color:Color;
+  getArea(): number;
+  a?:number;
+  b?:number;
+  c?:number;
+  radius?:number;
 }
```
Anna-Tkachenko, line 11: Figure interface should contain only those properties that should be in every figure (shape, color, getArea)

src/figuresClasses.ts
```diff
@@ -1,19 +1,83 @@
...
+
+  ) {
+    this.shape = 'triangle';
 
-export class Circle {
+    if ((a >= b + c)
+    || (b >= a + c)
+    || (c >= b + a)
+    || (c <= 0 || a <= 0 || b <= 0)) {
+      throw new Error('Error');
```
Anna-Tkachenko, line 29: Check point 3 in checklist

src/figuresClasses.ts
```diff
@@ -1,19 +1,83 @@
...
+
+  constructor(
+    public color:Color,
+    public a:number,
+    public b:number,
+  ) {
+    this.shape = 'rectangle';
+
+    if (a <= 0 || b <= 0) {
+      throw new Error('Error');
```
Anna-Tkachenko, line 72: Check point 3 in checklist

src/figuresClasses.ts
Anna-Tkachenko: It's better to create some types for defining the values that can be in shape and color properties

Anna-Tkachenko (changes requested):

src/figuresClasses.ts
```diff
@@ -1,19 +1,79 @@
+type Shape ='triangle'| 'circle' | 'rectangle';
+type Color = 'red'|'green'|'blue';
 export interface Figure {
-
+  shape:Shape;
+  color:Color;
+  getArea(): number;
```
Anna-Tkachenko, line 6:
```suggestion
  shape: Shape;
  color: Color;
  getArea(): number;
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,79 @@
...
+export class Triangle implements Figure {
+  shape:Shape;
 
-}
+  constructor(
+    public color:Color,
+    public a :number,
+    public b: number,
+    public c:number,
+
```
Anna-Tkachenko, line 17: Check indents

Anna-Tkachenko (approved):

PR #242, opened 2022-02-09
Solution

GitRoko:


pogrebnyakp (approved):

src/figuresClasses.ts
```diff
@@ -1,19 +1,82 @@
...
+    public c: number,
+  ) {
+    this.shape = 'triangle';
+
+    if ((a >= b + c)
+    || (b >= a + c)
+    || (c >= b + a)
+    || (a <= 0 || b <= 0 || c <= 0)) {
+      throw new Error('Error, a >= b + c, b >= a + c, c >= b + a,'
+        + ' a <= 0 || b <= 0 || c <= 0');
```
Anna-Tkachenko, line 65: It's better to specify text reason why this values are incorrect

src/figuresClasses.ts
```diff
@@ -1,19 +1,82 @@
...
+  }
+
+  getArea(): number {
+    const sp = (this.a + this.b + this.c) / 2;
+    const aDifference = sp - this.a;
+    const bDiffernece = sp - this.b;
+    const cDifference = sp - this.c;
+    const area = Math.sqrt(sp * aDifference * bDiffernece * cDifference);
+
+    return Math.floor(area * 100) / 100;
```
Anna-Tkachenko, line 76: Don't repeat yourself: you can create separate function for Math.floor(... * 100) / 100 logic

Anna-Tkachenko (approved):

PR #243, opened 2022-02-09
Dev

KostyaKovbel:


src/figuresClasses.ts
```diff
@@ -1,19 +1,111 @@
...
+    color: 'red' | 'green' | 'blue',
+    a: number,
+    b: number,
+    c: number,
+  ) {
+    this.color = color;
+    this.shape = 'triangle';
+    this.a = a;
+    this.b = b;
+    this.c = c;
```
mykhailoyanush, line 30: Тут можна передати в конструктор параметр доступу і непотрібно буде оголошувати і присвоювати дані:
`constructor(
    public color: Color,
    public x: number,
    public y: number,
    public c: number,
  )`

src/figuresClasses.ts
```diff
@@ -1,19 +1,111 @@
...
+
+    if (this.a >= this.b + this.c
+      || this.c >= this.a + this.b
+      || this.b >= this.c + this.a
+    ) {
+      throw new Error('please, put correct data');
+    }
+
+    if (this.a === 0 || this.b === 0 || this.c === 0) {
```
mykhailoyanush, line 39: В умові вказано :
any length is <= 0

mykhailoyanush (approved):
Хороша робота!
Пару рекомендацій по коду

src/figuresClasses.ts
```diff
@@ -1,19 +1,109 @@
-export interface Figure {
 
+export interface Figure {
+  color: 'red' | 'green' | 'blue';
+  shape: string;
```
Anna-Tkachenko, line 4: It's better to create separate types for color values and shape values and use it here instead of string type

src/figuresClasses.ts
```diff
@@ -1,19 +1,109 @@
...
+  radius?: number;
```
Anna-Tkachenko, line 5: In Figure interface you should add only those properties, that should be in each figure
Add radius only for Circle, not for each Figure

src/figuresClasses.ts
```diff
@@ -1,19 +1,109 @@
...
+  getArea: () => number;
 }
 
-export class Triangle {
+export class Triangle implements Figure {
+  color: 'red' | 'green' | 'blue';
```
Anna-Tkachenko, line 10: Create separate type for all available colors and don't repeat them in your code

src/figuresClasses.ts
Anna-Tkachenko: Agree

Anna-Tkachenko (changes requested):
Fix comments

src/figuresClasses.ts
```diff
@@ -1,19 +1,83 @@
...
-export class Circle {
+    if (this.a >= this.b + this.c
+      || this.c >= this.a + this.b
+      || this.b >= this.c + this.a
+    ) {
+      throw new Error('please, put correct data');
+    }
 
+    if (this.a <= 0 || this.b <= 0 || this.c <= 0) {
+      throw new Error('please, put correct data');
```
Anna-Tkachenko, line 29: It's better to define the reason why data incorrect

Anna-Tkachenko (approved):
Good job 👍

PR #246, opened 2022-02-09
Complite

NovikovM18:


src/figuresClasses.ts
```diff
@@ -1,19 +1,108 @@
 export interface Figure {
-
+  color: 'red' | 'green' | 'blue';
+  shape: string;
+  radius?: number;
+  getArea: () => number;
```
Anna-Tkachenko, line 5: This interface should store only those properties, that should be in each figure

Anna-Tkachenko: It's better to create separate types to define what values can be in color and shape properties

src/figuresClasses.ts
```diff
@@ -1,19 +1,108 @@
...
+
+    if (this.a >= this.b + this.c
+      || this.c >= this.a + this.b
+      || this.b >= this.c + this.a
+    ) {
+      throw new Error('please, put correct data');
+    }
+
+    if (this.a === 0 || this.b === 0 || this.c === 0) {
+      throw new Error('please, put correct data');
```
Anna-Tkachenko, line 39: It's better to specify why this data is wrong

Anna-Tkachenko (changes requested):

PR #251, opened 2022-02-09
Save

DanyloLipar:


src/figuresClasses.ts
```diff
@@ -1,19 +1,80 @@
...
+    this.shape = 'triangle';
 
-export class Circle {
+    if (this.a <= 0
+      || this.b <= 0
+      || this.c <= 0
+      || a + b <= c
+      || a + c <= b
+      || b + c <= a) {
+      throw new Error('Error!');
```
Anna-Tkachenko, line 27: Check point 3 in checklist

Anna-Tkachenko (changes requested):

src/figuresClasses.ts
```diff
@@ -1,19 +1,80 @@
...
+  ) {
+    this.shape = 'rectangle';
+
+    if (width <= 0 || height <= 0) {
+      throw new Error('Resulting value should be even number');
+    }
+  }
+
+  getArea(): number {
+    return Math.floor((this.width * this.height) * 100) / 100;
```
Anna-Tkachenko, line 74: Don't repeat yourself: you can create separate function for Math.floor( ... * 100) / 100 logic

Anna-Tkachenko (approved):

PR #252, opened 2022-02-09
Figures classes

polosanya:


src/figuresClasses.ts
```diff
@@ -1,19 +1,93 @@
...
+  color: Color
+  getArea(): number
 }
 
-export class Triangle {
+function checkParameters(...params: number[]): void {
+  if (params.some((param) => param <= 0)) {
+    throw new Error('All numbers should be possitive');
+  }
+
```
Anna-Tkachenko, line 19: 👍

src/figuresClasses.ts
```diff
@@ -1,19 +1,93 @@
...
+    checkParameters(a, b, c);
+  }
+
+  getArea(): number {
+    const p = (this.a + this.b + this.c) / 2;
+
+    const area = Math.sqrt(p * (p - this.a) * (p - this.b) * (p - this.c));
 
+    return Math.floor(area * 100) / 100;
```
Anna-Tkachenko, line 48: Don't repeat yourself: you can create separate function for this logic

Anna-Tkachenko (approved):
Good job 👍

PR #253, opened 2022-02-09
solution

vsuslov29:


src/figuresClasses.ts
```diff
@@ -1,19 +1,86 @@
 export interface Figure {
+  shape: string,
+  color: string,
```
Anna-Tkachenko, line 3: It's better to create types to define what values can be stored in shape and color properties

src/figuresClasses.ts
```diff
@@ -1,19 +1,86 @@
...
+  getArea: () => number,
+}
+
+type Color = 'red' | 'green' | 'blue';
 
+function roundToHundreds(num: number): number {
+  return Math.floor(num * 100) / 100;
 }
 
 export class Triangle {
```
Anna-Tkachenko, line 13: Each class should implement Figure interface

Anna-Tkachenko (changes requested):

PR #256, opened 2022-02-09
Solution

cxrmx:


Tsekhmister (approved):
👍

src/figuresClasses.ts
```diff
@@ -1,19 +1,97 @@
...
 
-export class Circle {
+    if (a <= 0
+      || b <= 0
+      || c <= 0
+      || a + b <= c
+      || a + c <= b
+      || b + c <= a
+    ) {
+      throw new Error('invalid sides');
```
Anna-Tkachenko, line 35: It's better to specify the reason why sides values are wrong

src/figuresClasses.ts
```diff
@@ -1,19 +1,97 @@
...
+  height;
+
+  constructor(
+    color: Color,
+    width: number,
+    height: number,
+  ) {
+    this.shape = 'rectangle';
+    this.color = color;
+    this.width = width;
```
Anna-Tkachenko, line 82: You can do it simpler
```suggestion
  constructor(
    public color: Color,
    public width: number,
    public height: number,
  ) {
    this.shape = 'rectangle';
```

Anna-Tkachenko (changes requested):

Anna-Tkachenko (approved):

PR #257, opened 2022-02-09
Solution

maxim2310:


src/figuresClasses.ts
```diff
@@ -1,19 +1,71 @@
 export interface Figure {
-
+  shape: string,
+  color: 'red' | 'green' | 'blue',
```
Anna-Tkachenko, line 3: It's better to create types for defining values that can be in shape and color properties

Anna-Tkachenko (approved):

PR #288, opened 2022-03-16
Solution

Ol-Lav:


src/figuresClasses.ts
```diff
@@ -1,19 +1,75 @@
...
+  constructor(
+    public color: Color,
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    if (a >= (b + c) || b >= (a + c) || c >= (b + a)
+    || (a <= 0 || b <= 0 || c <= 0)
+    ) {
+      throw new Error('Error');
```
Olexandr6, line 22: Check the 3-rd point in the [checklist](https://github.com/mate-academy/ts_figures_classes/blob/master/checklist.md). Change it in all places

src/figuresClasses.ts
```diff
@@ -1,19 +1,75 @@
-export interface Figure {
+type Shape = 'triangle' | 'circle' | 'rectangle';
+type Color = 'red' | 'green' | 'blue';
```
Olexandr6, line 2: Would be better to use `enum` instead

src/figuresClasses.ts
```diff
@@ -1,19 +1,75 @@
...
+    public color: Color,
+    public radius: number,
+  ) {
+    if (radius <= 0) {
+      throw new Error('Error');
+    }
+  }
 
+  getArea(): number {
+    return Math.floor(Math.PI * this.radius * this.radius * 100) / 100;
```
Olexandr6, line 51:
```suggestion
    return Math.floor((Math.PI * this.radius ** 2) * 100) / 100;
```

Olexandr6 (changes requested):
Your solution is good👍But needs some small improvements

src/figuresClasses.ts
Ol-Lav: Fixed

src/figuresClasses.ts
Ol-Lav: Fixed

PR #302, opened 2022-04-05
 ts_figures_classes

ievgen-volkov:


src/figuresClasses.ts
```diff
@@ -1,19 +1,87 @@
...
 
+  constructor(
+    public color: Color,
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    if (this.a >= this.b + this.c
+      || this.b >= this.a + this.c
+      || this.c >= this.a + this.b
```
ArtemTopolskyi, line 26: maybe create separate function and call it here three times

ArtemTopolskyi (approved):

PR #317, opened 2022-04-23
solution Figures

maxxnikiforov:


src/figuresClasses.ts
```diff
@@ -1,19 +1,79 @@
...
+
+  constructor(
+    public color: Color,
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    if (a <= 0 || b <= 0 || c <= 0) {
+      throw new Error('Error!');
+    }
```
YegorKochetkov, line 30: non-informative error message

src/figuresClasses.ts
```diff
@@ -1,19 +1,79 @@
...
+export class Rectangle implements Figure {
+  public shape = Shape.Rectangle;
+
+  constructor(
+    public color: Color,
+    public width: number,
+    public height: number,
+  ) {
+    if (width <= 0 || height <= 0) {
+      throw new Error('Error!');
```
YegorKochetkov, line 68: non-informative error message

YegorKochetkov (changes requested):
messages should be more informative

src/figuresClasses.ts
maxxnikiforov: Согласен, поправлю!

Andr1yK (approved):
LGTM 🔥🚀

PR #318, opened 2022-04-23
figuresClasses

StasivPavlo:


src/figuresClasses.ts
```diff
@@ -1,19 +1,86 @@
...
+    public a: number,
+  ) {
+    if (a <= 0) {
+      throw new Error('value can`t be 0 and <');
+    }
+  }
 
+  getArea(): number {
+    return Math.trunc((Math.PI * (this.a ** 2)) * 100) / 100;
+  }
```
YegorKochetkov, line 33: a method getArea that returns the area of the figure rounded down to hundredths.
Math.floor !== Math.trunc

src/figuresClasses.ts
```diff
@@ -1,19 +1,86 @@
...
+  getArea(): number,
 }
 
-export class Triangle {
+export class Circle implements Figure {
+  public shape = Shape.Circle;
+
+  constructor(
+    public color: Color,
+    public a: number,
```
YegorKochetkov, line 24: maybe 'radius' is more informative

src/figuresClasses.ts
```diff
@@ -1,19 +1,86 @@
...
+    public b: number,
+    public c: number,
+  ) {
+    if (a <= 0 || b <= 0 || c <= 0) {
+      throw new Error('value can`t be 0 and <');
+    }
+
+    if (a + b <= c || b + c <= a || c + a <= b) {
+      throw new Error('a + b can`t be c');
+    }
```
YegorKochetkov, line 51: Why? it's not very obvious what's wrong. it would be better
`sides 1, 2 and 3 can not form a triangle`
or
`sum of cathets can`t be less then hypotenuse`

src/figuresClasses.ts
```diff
@@ -1,19 +1,86 @@
...
 }
 
 export class Rectangle {
+  public shape = Shape.Rectangle;
 
-}
+  constructor(
+    public color: Color,
+    public a: number,
+    public b: number,
```
YegorKochetkov, line 72: it would be better use 'width' and 'height'

YegorKochetkov (changes requested):
Good) let fix minor flaws

src/figuresClasses.ts
```diff
@@ -1,19 +1,86 @@
-export interface Figure {
+enum Color {
+  Red = 'red',
+  Green = 'green',
+  Blue = 'Blue',
```
YegorKochetkov, line 4: why 'Blue' not 'blue')

src/figuresClasses.ts
```diff
@@ -1,19 +1,86 @@
...
+    public color: Color,
+    public radius: number,
+  ) {
+    if (radius <= 0) {
+      throw new Error('value can`t be 0 and <');
+    }
+  }
 
+  getArea(): number {
+    return Math.floor((Math.PI * (this.radius ** 2)) * 100) / 100;
```
YegorKochetkov, line 32: sorry( it's my mistake. you shouldn't use Math.floor

YegorKochetkov (approved):
sorry with Math.floor

PR #327, opened 2022-04-25
solution

kkiirill:


HelenPetrovska (approved):
Good job! You can use enums, but I like this option too.

Andr1yK (approved):
GJ! 🔥

PR #331, opened 2022-04-26
done

Yurii837:


src/figuresClasses.ts
```diff
@@ -1,19 +1,73 @@
...
+      || b >= a + c
+      || c >= a + b) {
+      throw new Error('Is not triangle');
+    }
+  }
 
+  getArea(): number {
+    return Math.floor((Math.sqrt(4 * this.a ** 2 * this.b ** 2
+    - (this.a ** 2 + this.b ** 2 - this.c ** 2) ** 2)) * 0.25 * 100) / 100;
+  }
```
RuslanaTomnyuk, line 33: There was another way of Heron's formula, that way was easier for me. It is only my opinion))

RuslanaTomnyuk (approved):
`use [Heron's formula](https://en.wikipedia.org/wiki/Heron%27s_formula) for triangles`

PR #350, opened 2022-05-19
Solution

pohrebniak23:


src/figuresClasses.ts
```diff
@@ -1,19 +1,75 @@
 export interface Figure {
-
+  color: string,
+  shape?: string,
```
aliceblunt8, line 3: create enum Shape and type Color and use it everywhere
```suggestion
  color: Color,
  shape: Shape,
```

aliceblunt8: what about getArea?

src/figuresClasses.ts
```diff
@@ -1,19 +1,75 @@
...
 
-export class Triangle {
+export class Triangle implements Figure {
+  constructor(
+    public color: string,
+    public a: number,
+    public b: number,
+    public c: number,
+    public shape: string = 'triangle',
+  ) {
```
aliceblunt8, line 13:
```suggestion
export class Triangle implements Figure {
public shape = Shape.Triangle;

  constructor(
    public color: Color,
    public a: number,
    public b: number,
    public c: number,
  ) {
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,75 @@
...
 
+    return Math.floor(square * 100) / 100;
+  }
 }
 
 export class Circle {
+  constructor(
+    public color: string,
+    public radius: number,
+    public shape: string = 'circle',
```
aliceblunt8, line 31: same here

src/figuresClasses.ts
```diff
@@ -1,19 +1,75 @@
...
+    return Math.floor(square * 100) / 100;
+  }
 }
 
 export class Rectangle {
+  constructor(
+    public color: string,
+    public width: number,
+    public height: number,
+    public shape: string = 'rectangle',
```
aliceblunt8, line 50: and here

aliceblunt8 (changes requested):

vkryvytskyy (approved):

PR #351, opened 2022-05-20
Implemented 'Figures classes' task

dellachaise:


src/figuresClasses.ts
```diff
@@ -1,19 +1,113 @@
+type Shapes = 'triangle' | 'circle' | 'rectangle';
+type Color = 'red' | 'green' | 'blue';
+
 export interface Figure {
+  shape: Shapes,
```
alexandra-protyanova, line 5: you can create enum for all the shapes)

src/figuresClasses.ts
```diff
@@ -1,19 +1,113 @@
...
+  }
 }
 
-export class Triangle {
+export class Triangle implements Figure {
+  shape: Shapes = 'triangle';
+
+  color: Color;
+
+  a: number;
```
alexandra-protyanova, line 21: In my opinnion it's better to write properties in constructor as `public a: number`, not to duplicate code

src/figuresClasses.ts
```diff
@@ -1,19 +1,113 @@
+type Shapes = 'triangle' | 'circle' | 'rectangle';
+type Color = 'red' | 'green' | 'blue';
+
 export interface Figure {
+  shape: Shapes,
+  color: Color,
+  getArea(): number
+}
 
+function validator(key: string, value: number):void {
```
alexandra-protyanova, line 10: I like this error function so much)

src/figuresClasses.ts
```diff
@@ -1,19 +1,113 @@
...
+    validator('b', b);
+    validator('c', c);
+
+    const arraySides: number[] = [a, b, c]
+      .sort((number1, number2) => number1 - number2);
+
+    if (arraySides[2] >= (arraySides[1] + arraySides[0])) {
+      throw new Error('Please enter correct side valies!');
+    }
+    this.color = color;
```
alexandra-protyanova, line 43: why do you need lines 43 - 46

src/figuresClasses.ts
```diff
@@ -1,19 +1,113 @@
...
+    this.c = c;
+  }
+
+  getArea(): number {
+    const semiPerimetr = (this.a + this.b + this.c) / 2;
+
+    const square = Math.sqrt(semiPerimetr * (semiPerimetr - this.a)
+      * (semiPerimetr - this.b) * (semiPerimetr - this.c));
 
+    return Math.round(square * 100) / 100;
```
alexandra-protyanova, line 55: you can creat a function for rounding your result, `Math.floor` for all works perfectly

src/figuresClasses.ts
```diff
@@ -1,19 +1,113 @@
...
+
+  radius: number;
+
+  constructor(
+    color: Color,
+    radius: number,
+  ) {
+    validator('radius', radius);
+
+    this.color = color;
```
alexandra-protyanova, line 72: no need 72- 73

src/figuresClasses.ts
```diff
@@ -1,19 +1,113 @@
...
+  height: number;
+
+  constructor(
+    color: Color,
+    width: number,
+    height: number,
+  ) {
+    validator('width', width);
+    validator('heigth', height);
+    this.color = color;
```
alexandra-protyanova, line 99: also here

alexandra-protyanova (changes requested):

src/figuresClasses.ts
dellachaise: When I wrote code first time I had errors because of it, thats why wrote the long way. But now I refactored code, and everything's ok!

Denchepornyuk (approved):

PR #352, opened 2022-05-20
Solution

59MrRobot:


meforoma (approved):
Good one 👍 
great that you've introduced error validation helper

PR #353, opened 2022-05-20
solution

SavMary:


src/figuresClasses.ts
```diff
@@ -1,19 +1,82 @@
...
+export class Triangle implements Figure {
+  public shape = Shape.Triangle;
+
+  constructor(
+    public color: Color,
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    if (this.a <= 0 || this.b <= 0 || this.c <= 0) {
```
alexandra-protyanova, line 26: as an option you can add all parametrs in array and use `some`)

src/figuresClasses.ts
```diff
@@ -1,19 +1,82 @@
...
+      throw new Error('Ooops');
+    }
+
+    if ((a + b) <= c
```
alexandra-protyanova, line 30: also as opyion `sort` here)

alexandra-protyanova (approved):
Good job)

src/figuresClasses.ts
```diff
@@ -1,19 +1,82 @@
...
+    || (a + c) <= b || (c + b) <= a) {
+      throw new Error('Ooops');
```
Anna-Tkachenko, line 32: I think 'oops' won't be understandable error message for users)

src/figuresClasses.ts
```diff
@@ -1,19 +1,82 @@
...
+enum Shape {
+  Triangle = 'triangle',
+  Circle = 'circle',
+  Rectangle = 'rectangle',
 }
+type Color = 'red' | 'green' | 'blue';
+type Callback = (arg: number) => number;
 
-export class Triangle {
+const callback: Callback = (n) => Math.floor(n * 100) / 100;
```
Anna-Tkachenko, line 9: Will be better create more meaningful variable name for this function to specify what it should do

src/figuresClasses.ts
```diff
@@ -1,19 +1,82 @@
...
+export class Circle implements Figure {
+  public shape = Shape.Circle;
+
+  constructor(
+    public color: Color,
+    public radius: number,
+  ) {
+    if (radius <= 0) {
+      throw new Error('Ooops');
+    }
```
Anna-Tkachenko, line 54: You do this check in every class
Will be better to write separate function that can accept any number of params and check if all of them > 0

Anna-Tkachenko (approved):

PR #354, opened 2022-05-20
solution

alexandra-protyanova:


ZUBOV-ILLIA (approved):
Good Job!

src/figuresClasses.ts
```diff
@@ -1,19 +1,83 @@
...
-}
+  constructor(
+    public color: Color,
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    if ([a, b, c].some((number) => number <= 0)) {
+      throw new Error('One of the sides less than 1');
+    }
```
Anna-Tkachenko, line 30: You did this check for every initial params in every class
You can create separate function that can accept any numbers of params and check if every of this param > 0

Anna-Tkachenko (approved):
⭐️

PR #356, opened 2022-05-21
solution

ZUBOV-ILLIA:


alexandra-protyanova (approved):
Good job)

src/figuresClasses.ts
```diff
@@ -1,19 +1,82 @@
...
+    public color: Color,
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    this.a = a;
+
+    if (a <= 0 || b <= 0 || c <= 0) {
+      throw new Error('your error message');
+    }
```
Anna-Tkachenko, line 28: Let's specify understandable error message for users)

src/figuresClasses.ts
```diff
@@ -1,19 +1,82 @@
...
+export class Circle implements Figure {
+  public shape: Shape = 'circle';
+
+  constructor(
+    public color: Color,
+    public radius: number,
+  ) {
+    if (radius < 1) {
+      throw new Error('your error message');
+    }
```
Anna-Tkachenko, line 54: You do this check in every class for initial params
Will be better to create separate function that can accept any number of values and check that every value > 0

Anna-Tkachenko (approved):

PR #358, opened 2022-05-21
areas of figures can be calculated now

enrikadej:


src/figuresClasses.ts
```diff
@@ -1,19 +1,92 @@
-export interface Figure {
+function paramsOverZero(...args: number[]): void {
+  const params = args;
```
alexandra-protyanova, line 2: you don't need `params`, you may use `args` instead

src/figuresClasses.ts
```diff
@@ -1,19 +1,92 @@
...
+  Triangle = 'triangle',
+  Circle = 'circle',
+  Rectangle = 'rectangle',
+}
+
+export interface Figure {
+  shape: Shape
+  color: Color
 
+  getArea: () => {}
```
alexandra-protyanova, line 32: in my opinion it's better to use number as a type of function return

src/figuresClasses.ts
```diff
@@ -1,19 +1,92 @@
...
+  }
+
+  getArea(): number {
+    const semiP = (this.a + this.b + this.c) / 2;
+
+    const bigRes = Math.sqrt(
+      semiP * (semiP - this.a) * (semiP - this.b) * (semiP - this.c),
+    );
+
+    return Math.floor(bigRes * 100) / 100;
```
alexandra-protyanova, line 55: you also can create function for rounding

alexandra-protyanova (approved):
Well done!)

PR #359, opened 2022-05-22
Figures was done

vlasiuk-anatolii:


src/figuresClasses.ts
```diff
@@ -1,19 +1,105 @@
...
+
+  public color: Color;
+
+  constructor(
+    color: Color,
+    aSide: number,
+    bSide: number,
+    cSide: number,
+  ) {
+    this.aSide = aSide;
```
v-vytrykush, line 36: you can use public instead of using .this

v-vytrykush (commented):
Good decision with Enums!

src/figuresClasses.ts
```diff
@@ -1,19 +1,105 @@
...
+    this.bSide = bSide;
+    this.cSide = cSide;
+    this.color = color;
```
Denchepornyuk, line 39: you don't need this here, use public

src/figuresClasses.ts
```diff
@@ -1,19 +1,105 @@
...
+    if (aSide + bSide <= cSide
+    || bSide + cSide <= aSide
+    || cSide + aSide <= bSide) {
+      throw new Error('Wrong length of sides');
+    }
+  }
 
-export class Circle {
+  getArea(): number {
+    const halfPerimeter: number = (this.aSide + this.bSide + this.cSide) / 2;
```
Denchepornyuk, line 49: if your variables have a primitive type such as number, boolean, string, then you can avoid write type for these variables

src/figuresClasses.ts
```diff
@@ -1,19 +1,105 @@
...
+export class Circle implements Figure {
+  shape = Shape.CircleShape;
 
+  color: Color;
+
+  radius: number;
+
+  constructor(color: Color, radius: number) {
+    this.radius = radius;
+    this.color = color;
```
Denchepornyuk, line 67: use public

src/figuresClasses.ts
```diff
@@ -1,19 +1,105 @@
...
+
+    if (radius <= 0) {
+      throw new Error('fail radius');
+    }
+  }
+
+  getArea(): number {
+    return Math.floor(Math.PI * this.radius ** 2 * 100) / 100;
```
Denchepornyuk, line 75: create a separate function for `* 100) / 100`

Denchepornyuk (changes requested):

src/figuresClasses.ts
```diff
@@ -1,19 +1,83 @@
...
+    * (halfPerimeter - this.bSide)
+    * (halfPerimeter - this.cSide)));
+  }
 }
 
-export class Rectangle {
+export class Circle implements Figure {
+  shape = Shape.CircleShape;
+
+  constructor(public color: Color, public radius: number) {
```
Denchepornyuk, line 52:
```suggestion
  constructor(
    public color: Color,
    public radius: number,
  ) {
```

Denchepornyuk (approved):

PR #361, opened 2022-05-22
figures type script solution

valerii-derkach:


src/figuresClasses.ts
```diff
@@ -1,19 +1,93 @@
...
+  getArea(): number {
+    const semiperimeter: number = (this.a + this.b + this.c) / 2;
+
+    const area
+      = Math.sqrt(semiperimeter
+      * (semiperimeter - this.a)
+      * (semiperimeter - this.b)
+      * (semiperimeter - this.c));
 
+    return Math.floor(area * 100) / 100;
```
alexandra-protyanova, line 27: you use rounding twice at your code, it would be better to create separate function

src/figuresClasses.ts
```diff
@@ -1,19 +1,93 @@
...
+  }
+
+  constructor(
+    public color: Color,
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    if (a <= 0 || b <= 0 || c <= 0) {
+      throw new Error('No, no! This is not rigth! '
```
alexandra-protyanova, line 37: same thing here you will check arguments three times

alexandra-protyanova (changes requested):
Fix comments, please)

src/figuresClasses.ts
```diff
@@ -1,19 +1,104 @@
...
+
+export type Color = 'red' | 'green' | 'blue';
+
 export interface Figure {
+  shape: Shape,
+  color: Color,
+  getArea(): number
+}
 
+export function roundToHundredths(uglyNumber: number): number {
```
Denchepornyuk, line 15: if your function returns a primitive type you can remove function return type description
```suggestion
export function roundToHundredths(uglyNumber: number) {
```

Denchepornyuk (approved):

src/figuresClasses.ts
valerii-derkach: yes i agree, but linter insisted that I leave the function return type

valerii-derkach: Done

PR #362, opened 2022-05-22
commit message

AndriySimyachenko:


src/figuresClasses.ts
```diff
@@ -1,19 +1,90 @@
...
+
+  constructor(
+    public color: Color,
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    const sides: number[] = [a, b, c].sort((sideA, sideB) => sideB - sideA);
+
+    if (a < 0 || b < 0 || c < 0) {
```
alexandra-protyanova, line 34: you use check for valid arguments three times in your code, create separate function

alexandra-protyanova (approved):
Well done)

PR #364, opened 2022-05-22
solution

AnastasiyaSopotnytska:


src/figuresClasses.ts
```diff
@@ -1,19 +1,80 @@
...
+      throw new Error('IT IS NOT A TRIANGLE');
+    }
+  }
+
+  getArea(): number {
+    const p: number = (this.a + this.b + this.c) / 2;
+    const area: number = Math.sqrt(p
+      * (p - this.a) * (p - this.b) * (p - this.c));
+
+    return Math.floor(100 * area) / 100;
```
vlad-demchuk, line 35: you could create a function for rounding, because you do the same several times

src/figuresClasses.ts
```diff
@@ -1,19 +1,80 @@
...
+  ) {
+    if (this.radius <= 0) {
+      throw new Error('Incorrect RADIUS');
+    }
+  }
 
+  getArea(): number {
+    const area: number = Math.PI * this.radius * this.radius;
+
+    return Math.floor(100 * area) / 100;
```
vlad-demchuk, line 54: here

src/figuresClasses.ts
```diff
@@ -1,19 +1,80 @@
...
+  ) {
+    if (this.width <= 0 || this.height <= 0) {
+      throw new Error('Incorrect SIDES');
+    }
+  }
+
+  getArea(): number {
+    const area: number = this.height * this.width;
+
+    return Math.floor(100 * area) / 100;
```
vlad-demchuk, line 74: and here

vlad-demchuk (approved):
Good job)

Denchepornyuk (approved):

PR #372, opened 2022-05-23
Figures classes

dasha-chekmariova:


iwanttosmile4u (approved):
Looks good!)

src/figuresClasses.ts
```diff
@@ -1,19 +1,81 @@
...
+    public width: number,
+    public height: number,
+  ) {
+    if (width <= 0 || height <= 0) {
+      throw new Error('You are trying to draw the non-existent');
+    }
+  }
+
+  getArea(): number {
+    return this.width * this.height;
```
Denchepornyuk, line 75: use should use round function here

Denchepornyuk (approved):

PR #374, opened 2022-05-23
solution

efdalyalcin:


MabuzaM (approved):
Nice one!

src/figuresClasses.ts
```diff
@@ -1,19 +1,84 @@
...
 
-export function getInfo(figure) {
+export class Triangle extends Rectangle {
+  public shape = Shapes.Triangle;
+
+  constructor(
+    color: Color,
+    a: number,
+    b: number,
+    public c: number,
```
meforoma, line 63: better to keep figure-specific measures private (radius is private for circle, 3 sides - for triangle, etc.; while color is public - shared across all figures)
```suggestion
    private a: number,
    private b: number,
    private c: number,
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,84 @@
...
+  ) {
+    super(color, a, b);
```
meforoma, line 65: you could skip using `super()` if properly define constructor props

src/figuresClasses.ts
```diff
@@ -1,19 +1,84 @@
...
+    return roundNumbers(this.a * this.b);
+  }
 }
 
-export function getInfo(figure) {
+export class Triangle extends Rectangle {
+  public shape = Shapes.Triangle;
+
+  constructor(
+    color: Color,
```
meforoma, line 60:
```suggestion
    public color: Color,
```

meforoma (changes requested):

src/figuresClasses.ts
efdalyalcin: the linter gets very angry if I try to skip the super()

efdalyalcin: the linter gets very angry if I try to skip the super()

src/figuresClasses.ts
efdalyalcin: I think private doesn't give access to the extended classes so I had to use protected.

meforoma (approved):

PR #376, opened 2022-05-23
ts_figures_classes

anastasiia-nikita:


src/figuresClasses.ts
```diff
@@ -1,19 +1,88 @@
...
+    return roundTo2(area);
+  }
 
+  constructor(
+    public color: Color,
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    if (a <= 0 || b <= 0 || c <= 0
```
iwanttosmile4u, line 40: If you want, you can create separate function for checking figure sides (if they <= 0) and call this function with the appropriate parameters. Because this checking repeat in all classes.

iwanttosmile4u (approved):
Good job!) Just have a little recomendation in comments)

Denchepornyuk (approved):

PR #378, opened 2022-05-23
commit figures

volodya-leskiv:


src/figuresClasses.ts
```diff
@@ -1,19 +1,95 @@
...
+
+  public b;
+
+  public c;
+
+  constructor(color: Color, a: number, b: number, c: number) {
+    this.color = color;
+    this.a = a;
+    this.b = b;
+    this.c = c;
```
vlad-demchuk, line 29: don't repeat yourself, you can add variables to the constructor once (constructor(public color...){} )

src/figuresClasses.ts
```diff
@@ -1,19 +1,95 @@
...
+  public width;
+
+  public color;
+
+  public height;
+
+  constructor(color: Color, width: number, height: number) {
+    this.width = width;
+    this.height = height;
+    this.color = color;
```
vlad-demchuk, line 81: the same

vlad-demchuk (changes requested):

src/figuresClasses.ts
volodya-leskiv: thanks for review)

Denchepornyuk (approved):

PR #379, opened 2022-05-23
Solution

amitpatiljc:


src/figuresClasses.ts
```diff
@@ -1,19 +1,86 @@
...
+export class Triangle implements Figure {
+  shape: Shape.Triangle;
+
+  color: Color;
+
+  getArea = (): number => {
+    const s = (this.a + this.b + this.c) / 2;
+    const area = Math.sqrt(s * (s - this.a) * (s - this.b) * (s - this.c));
+
+    return Math.floor(area * 100) / 100;
```
meforoma, line 28: seems legit to introduce a rounding helper function to use it 3 times in your code

src/figuresClasses.ts
```diff
@@ -1,19 +1,86 @@
...
 
-export class Circle {
+export class Triangle implements Figure {
+  shape: Shape.Triangle;
+
+  color: Color;
+
+  getArea = (): number => {
+    const s = (this.a + this.b + this.c) / 2;
+    const area = Math.sqrt(s * (s - this.a) * (s - this.b) * (s - this.c));
```
meforoma, line 26: wondering, would de-structuring reduce the character-count:
`const { a, b, c } = this;`

meforoma (changes requested):

src/figuresClasses.ts
amitpatiljc: Done

meforoma (approved):
🔥

PR #384, opened 2022-05-24
task completed

andreisucman:


Uktam19980416 (approved):
Good job

src/figuresClasses.ts
```diff
@@ -1,19 +1,90 @@
...
 
-export function getInfo(figure) {
+export class Triangle extends Rectangle {
+  public shape = Shapes.Triangle;
+
+  constructor(
+    color: Color,
+    a: number,
+    b: number,
+    public c: number,
```
meforoma, line 63: generally `color` is a shared prop - so we could say `public`
while all the measures (sides) are `private` for each class
```suggestion
    public color: Color,
    private a: number,
    private b: number,
    private c: number,
```

meforoma (changes requested):
pleas apply comments from Triangle for all figure classes

src/figuresClasses.ts
```diff
@@ -1,19 +1,90 @@
...
+export class Triangle implements Figure {
+  public shape = Shapes.Triangle;
+
+  constructor(
+    public color: Color,
+    private a: number,
+    private b: number,
+    private c: number,
+  ) {
+    super(color, a, b);
```
meforoma, line 65: seems you can rid of super()
super()

src/figuresClasses.ts
```diff
@@ -1,19 +1,90 @@
...
+    }
+  }
 
+  getArea(): number {
+    return roundNumbers((this.a ** 2) * Math.PI);
+  }
 }
 
-export class Rectangle {
+export class Rectangle extends Circle {
```
meforoma, line 36: export class Circle `implements Figure {`
export class Triangle `implements Figure {`
BUT
export class Rectangle `extends Circle {`

meforoma (changes requested):
few more

src/figuresClasses.ts
andreisucman: Fixed

src/figuresClasses.ts
andreisucman: Okay, fixed.

VinnieJ-2k20 (approved):

PR #385, opened 2022-05-24
First_commit

igoree88:


src/figuresClasses.ts
```diff
@@ -1,19 +1,90 @@
...
+
+function squareRound(square: number):number {
+  return Math.floor(square * 100) / 100;
+}
+
 export interface Figure {
+  shape: Shape;
+  color: Color;
 
+  getArea: () => {};
```
AndriiRut, line 12: what dose this method chould return? I think you need to type it.

AndriiRut (changes requested):
GOOD Job!

Anna-Tkachenko (approved):

PR #386, opened 2022-05-24
figure_classes

elkhayate:


src/figuresClasses.ts
```diff
@@ -1,19 +1,79 @@
...
+    const { a, b, c } = this;
+    const semiPerimeter: number = (a + b + c) / 2;
+
+    return Number(
+      (Math.sqrt(semiPerimeter
+        * (semiPerimeter - a)
+        * (semiPerimeter - b)
+        * (semiPerimeter - c))
+      ).toFixed(2),
+    );
```
amitpatiljc, line 38: i think rounding helper function will be helpful as we have to round up all three  'getArea' values.

src/figuresClasses.ts
```diff
@@ -1,19 +1,79 @@
...
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    const sides = [a, b, c].sort((x, y) => y - x);
 
-export class Circle {
+    if ((sides[0] >= sides[1] + sides[2])
+        || a < 1 || b < 1 || c < 1) {
+      throw new Error('All sides must be greater than 0');
```
amitpatiljc, line 24: we have a condition that all sides must be greater than 0 , and we can have side = 1
so we should compare like a < 0 or a <= 1

amitpatiljc (changes requested):
Good work but i have two suggestions

src/figuresClasses.ts
```diff
@@ -1,19 +1,79 @@
...
+    public color: Color,
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    const sides = [a, b, c].sort((x, y) => y - x);
 
-export class Circle {
+    if ((sides[0] >= sides[1] + sides[2])
+        || a < 1 || a <= 1) {
```
meforoma, line 23: these two evaluations seem contradictory
if a = 1: 
```
  || a < 1 // false
  || a <= 1 // true
```
based on your evaluation , an Error will be thrown, yet it shouldn't

meforoma (changes requested):

meforoma (changes requested):

PR #388, opened 2022-05-25
solution

Sardorbek1122:


src/figuresClasses.ts
```diff
@@ -1,19 +1,89 @@
...
+  }
 }
 
-export function getInfo(figure) {
+export class Rectangle implements Figure {
+  shape: Shape.Rectangle;
+
+  color: Color;
+
+  getArea = (): number => Math.floor(this.height * this.width * 100) / 100;
```
efdalyalcin, line 75: You already have the rounding(), why not use it?

src/figuresClasses.ts
```diff
@@ -1,19 +1,89 @@
...
 
-export class Circle {
+    return rounding(Math.sqrt(s * (s - a) * (s - b) * (s - c)));
+  };
 
+  constructor(
+    color: Color,
+    public a: number,
+    public b: number,
+    public c: number,
```
efdalyalcin, line 38: `These values shouldn't be reachable from outside` is what i was told..

src/figuresClasses.ts
```diff
@@ -1,19 +1,89 @@
...
+export class Circle implements Figure {
+  shape: Shape.Circle;
+
+  color: Color;
 
+  getArea = (): number => rounding(Math.PI * (this.radius ** 2));
+
+  constructor(
+    color: Color,
+    public radius: number,
```
efdalyalcin, line 59: same here

src/figuresClasses.ts
```diff
@@ -1,19 +1,89 @@
...
 
-export function getInfo(figure) {
+export class Rectangle implements Figure {
+  shape: Shape.Rectangle;
+
+  color: Color;
+
+  getArea = (): number => Math.floor(this.height * this.width * 100) / 100;
+
+  constructor(color: Color, public width: number, public height: number) {
```
efdalyalcin, line 77: I think you should put them in separate lines and make them private

efdalyalcin (changes requested):
Well done! I left just small notes.

src/figuresClasses.ts
VinnieJ-2k20: Agree

VinnieJ-2k20 (changes requested):
Some changes needed.

VinnieJ-2k20 (changes requested):

VinnieJ-2k20 (approved):

PR #389, opened 2022-05-25
Solution

huberciak:


src/figuresClasses.ts
```diff
@@ -1,19 +1,100 @@
...
+      throw new Error('Side/Sides lesser than 0');
+    }
+
+    if (!heronsFormula([a, b, c])) {
+      throw new Error('Triangle sides doesnt match Herons Formula');
+    }
+  }
+
+  getArea(): number {
+    const s: number = (this.a + this.b + this.c) / 2;
```
efdalyalcin, line 56: It is a good place to use object destructing

meforoma (approved):
great implementation @huberciak 👍

PR #392, opened 2022-05-25
Figures

proznerka:


andreisucman (approved):
Great job.

meforoma (approved):
nice job 👍 
just one comment :
all measures for Figures are not `public` - but `private`, 
while `color:` is public indeed

PR #394, opened 2022-05-26
solution

ShapovalDenys:


src/figuresClasses.ts
```diff
@@ -1,19 +1,98 @@
-export interface Figure {
+type Color = 'red' | 'green' | 'blue';
 
+export interface Figure {
+  shape: 'triangle' | 'circle' | 'rectangle',
```
vkryvytskyy, line 4: create enum Figures

src/figuresClasses.ts
```diff
@@ -1,19 +1,98 @@
...
-export class Rectangle {
+export class Circle implements Figure {
+  public shape: 'circle';
 
+  public color: Color;
+
+  public radius: number;
+
+  constructor(color: Color, radius: number) {
+    this.shape = 'circle';
```
vkryvytskyy, line 57: use values from Figures enum here

src/figuresClasses.ts
```diff
@@ -1,19 +1,98 @@
-export interface Figure {
+type Color = 'red' | 'green' | 'blue';
```
vkryvytskyy, line 1: also would be better to convert it to enum and use in classes

vkryvytskyy (changes requested):

PR #395, opened 2022-05-26
this one took a while

WojciechCzarnowski:


rHashTag (approved):
LGTM!

meforoma (approved):
nice job 👍
just one comment :
all measures for Figures are not `public` - but `private`,
while `color:` is public indeed

PR #399, opened 2022-06-02
Solution

naziamoff:


src/figuresClasses.ts
```diff
@@ -1,19 +1,74 @@
-export interface Figure {
+type Color = 'red'| 'green'| 'blue';
+type Shape = 'triangle' | 'circle' | 'rectangle';
```
aliceblunt8, line 2: better create enum for Shape, then you can use it like Shape.Triangle or other in classes

aliceblunt8 (approved):

PR #400, opened 2022-06-02
figure classes

roman-lipatov:


src/figuresClasses.ts
```diff
@@ -1,19 +1,78 @@
-export interface Figure {
+type Shape = 'triangle' | 'circle' | 'rectangle';
+type Color = 'red' | 'green' | 'blue';
```
Bohdan-Maksymiuk, line 2: use enum

src/figuresClasses.ts
```diff
@@ -1,19 +1,78 @@
...
+      .sort((sideA, sideB) => sideB - sideA);
+
+    if (sides[0] >= sides[1] + sides[2]) {
+      throw new Error('Sides of triangle are not correct');
+    }
+  }
 
-export class Circle {
+  getArea(): number {
+    const p: number = (this.a + this.b + this.c) / 2;
```
Bohdan-Maksymiuk, line 32: avoid types here, because the result of your calc would be a number

src/figuresClasses.ts
```diff
@@ -1,19 +1,78 @@
...
+    public width: number,
+    public height: number,
+  ) {
+    if (this.width <= 0 || this.height <= 0) {
+      throw new Error('Invalid data');
+    }
+  }
+
+  getArea(): number {
+    const square: number = this.width * this.height;
```
Bohdan-Maksymiuk, line 70: avoid types here, because the result of your calc would be a number

Bohdan-Maksymiuk (changes requested):

aliceblunt8 (approved):

PR #404, opened 2022-06-19
ts_figures_classes

vitaliy-kovbasinskyi:


src/figuresClasses.ts
```diff
@@ -1,19 +1,89 @@
 export interface Figure {
-
+  shape: 'triangle'| 'circle' | 'rectangle',
+  color: 'red'| 'green' | 'blue'
```
vkryvytskyy, line 3: create enums with this values and use them in all of the classes

src/figuresClasses.ts
```diff
@@ -1,19 +1,89 @@
...
+    public b: number,
+    public c: number,
+  ) {
+    this.color = color;
+
+    const sides: number[] = [a, b, c].sort((sideA, sideB) => sideB - sideA);
+
+    if (a <= 0 || b <= 0 || c <= 0) {
+      throw new Error('Not a triangle!');
+    }
```
vkryvytskyy, line 23: check this **before** sorting

vkryvytskyy (changes requested):

PR #406, opened 2022-06-23
Solution

mira1zu:


AndriiNychaiuk (approved):
Great job 🔥 . The only thing I'm not sure your interface has to contain all the variables. The meaning of the interface is to pass to the class info about what exactly it has to contain

src/figuresClasses.ts
```diff
@@ -1,19 +1,82 @@
-export interface Figure {
+type Shape = 'triangle' | 'circle' | 'rectangle';
+type Color = 'red' | 'green' | 'blue';
 
+export interface Figure {
+  a?: number,
+  b?: number,
+  c?: number,
+  r?: number
```
mykola-koval, line 8: do we need this?

mykola-koval (approved):

PR #409, opened 2022-06-25
Solution

liliia-livandovskaya:


src/figuresClasses.ts
```diff
@@ -1,19 +1,83 @@
...
+    if (this.a >= this.b + this.c
+      || this.b >= this.a + this.c
+      || this.c >= this.a + this.b) {
+      throw new Error('The side value cant be more then sum of two others');
+    }
+  }
 
+  getArea(): number {
+    const { a, b, c } = this;
+    const semiPerimeter = (this.a + this.b + this.c) / 2;
```
Angelion541, line 32: You can use short param there.

src/figuresClasses.ts
```diff
@@ -1,19 +1,83 @@
...
 
-}
+  constructor(
+    public color: Color,
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    if (this.a <= 0 || this.b <= 0 || this.c <= 0) {
+      throw new Error('Please enter correct side values');
```
Angelion541, line 20: If you see this error you can understood what need input? Write such us in 50 line.

Angelion541 (approved):

src/figuresClasses.ts
Olexandr6: Agree, what is "correct side"? Let's rename this error title

src/figuresClasses.ts
Olexandr6, line 32:
```suggestion
    const semiPerimeter = (a + b + c) / 2;
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,83 @@
...
-export class Rectangle {
+export class Circle implements Figure {
+  public shape: Shape = 'circle';
+
+  constructor(
+    public color: Color,
+    public radius: number,
+  ) {
+    if (this.radius <= 0) {
+      throw new Error('Radius value must be more then 0');
```
Olexandr6, line 50:
```suggestion
      throw new Error('Radius value must be more than 0');
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,83 @@
...
+export class Rectangle implements Figure {
+  public shape: Shape = 'rectangle';
+
+  constructor(
+    public color: Color,
+    public width: number,
+    public height: number,
+  ) {
+    if (this.width <= 0 || this.height <= 0) {
+      throw new Error('Sides of rectangle  must be more then 0');
```
Olexandr6, line 70:
```suggestion
      throw new Error('Sides of rectangle must be more than 0');
```

Olexandr6 (changes requested):
Let's fix small things

Olexandr6 (approved):

PR #411, opened 2022-06-26
solution

vinnibea:


src/figuresClasses.ts
```diff
@@ -1,19 +1,80 @@
...
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    if (a === 0 || b === 0 || c === 0) {
+      throw new Error('One of the sides is Zero(0)');
+    }
+
+    if (a + b <= c || c + b <= a || a + c <= b) {
+      throw new Error('Use proper values to calculate area');
```
Olexandr6, line 26: What is "proper values"? It's not obvious to a user. Fix error msg
```suggestion
      throw new Error('Use proper values to calculate area');
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,80 @@
...
+    }
+  }
+
+  getArea(): number {
+    const p: number = 0.5 * (this.a + this.b + this.c);
+    const S: number = Math.sqrt(p * ((p - this.a)
```
Olexandr6, line 32: What is `p`, `S` rename these variables
```suggestion
    const p = 0.5 * (this.a + this.b + this.c);
    const S = Math.sqrt(p * ((p - this.a)
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,80 @@
-export interface Figure {
 
-}
+type Shape = 'triangle' | 'circle' | 'rectangle';
```
Olexandr6, line 2: Better to create Enum for these values, cause you use them in your code

src/figuresClasses.ts
```diff
@@ -1,19 +1,80 @@
...
 
+export interface Figure {
+  shape: Shape,
+  color: Color,
+  getArea(): number,
 }
 
-export class Circle {
+export class Triangle implements Figure {
+  shape: Shape = 'triangle';
```
Olexandr6, line 13: Enum

src/figuresClasses.ts
```diff
@@ -1,19 +1,80 @@
...
+export class Rectangle implements Figure {
+  shape: Shape = 'rectangle';
+
+  constructor(
+    public color: Color,
+    public a: number,
+    public b: number,
+  ) {
+    if (this.a <= 0 || this.b <= 0) {
+      throw new Error('Use proper values to calculate area');
```
Olexandr6, line 67: The same here

src/figuresClasses.ts
```diff
@@ -1,19 +1,80 @@
...
+    }
+  }
+
+  getArea(): number {
+    const S: number = this.a * this.b;
```
Olexandr6, line 72: Rename
```suggestion
    const S = this.a * this.b;
```

Olexandr6 (changes requested):

Olexandr6 (approved):

PR #414, opened 2022-06-27
solution

Angelion541:


src/figuresClasses.ts
```diff
@@ -1,19 +1,73 @@
-export interface Figure {
+type Shape = 'triangle' | 'circle' | 'rectangle';
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,73 @@
...
+      throw new Error('Sides can`t be equal to or less than 0');
+    }
+
+    if (a >= b + c || b >= a + c || c >= a + b) {
+      throw new Error('One of the sides can`t be bigger than two others');
+    }
+  }
 
+  getArea(): number {
+    const s = (this.a + this.b + this.c) / 2;
```
Olexandr6, line 29: What is `s`?
```suggestion
    const s = (this.a + this.b + this.c) / 2;
```

Olexandr6 (changes requested):
Good job👍 Let's fix some small things

Olexandr6 (approved):

PR #419, opened 2022-06-28
Solution

GavrilyukYura:


src/figuresClasses.ts
```diff
@@ -1,19 +1,76 @@
...
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    if (a <= 0
+      || b <= 0
+      || c <= 0
+      || Math.max(a, b, c) >= (a + b + c) - Math.max(a, b, c)
+    ) {
+      throw new Error('it is not a triangle');
```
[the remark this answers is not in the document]
Olexandr6: Fix error msg in all places in your code

src/figuresClasses.ts
```diff
@@ -1,19 +1,76 @@
-export interface Figure {
+type Shape = 'triangle'| 'circle' | 'rectangle';
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,76 @@
...
+    ) {
+      throw new Error('it is not a triangle');
+    }
+  }
 
-export class Circle {
+  getArea(): number {
+    const semiPerimeter: number = (this.a + this.b + this.c) / 2;
 
+    return +(Math.sqrt(semiPerimeter
```
Olexandr6, line 31: Using `Number()` constuctor is more obvious
```suggestion
    return Number(Math.sqrt(semiPerimeter
```

Olexandr6 (approved):

PR #423, opened 2022-06-28
Initial commit

Kaunin1:


NikitaKarbazin (approved):

src/figuresClasses.js
```diff
@@ -0,0 +1,60 @@
+"use strict";
+exports.__esModule = true;
+exports.getInfo = exports.Rectangle = exports.Circle = exports.Triangle = void 0;
+function roundNumber(num) {
+    return Math.floor(num * 100) / 100;
+}
+var Triangle = /** @class */ (function () {
```
Olexandr6, line 7: You don't need to push this changes

src/figuresClasses.ts
```diff
@@ -1,19 +1,88 @@
-export interface Figure {
+type Shape = 'triangle'| 'circle'| 'rectangle';
```
Olexandr6, line 1: Better to use `enum` for these values, cause you use them in your code as values

src/figuresClasses.ts
```diff
@@ -1,19 +1,88 @@
...
+    color: Color,
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    this.shape = 'triangle';
+    this.color = color;
+
+    if (a >= b + c || b >= a + c || c >= a + b) {
+      throw new Error('Please provide valid dimension of Triangle side');
```
Olexandr6, line 29: What is "valid dimension"? Let's make this msg more informative for a user.

src/figuresClasses.ts
```diff
@@ -1,19 +1,88 @@
...
+    }
+  }
+
+  getArea(): number {
+    const area = (this.a + this.b + this.c) / 2;
```
Olexandr6, line 34: It's not area, rather `semiPerimetr`

Olexandr6 (changes requested):
Good job🔥 Small things should be fixed

Olexandr6 (approved):

PR #433, opened 2022-06-29
Solution

viktorlyt:


AndrewKhorsun (approved):
Все ок !

src/figuresClasses.ts
```diff
@@ -1,19 +1,91 @@
...
+  public shape: Shapes = Shapes.triangle;
 
+  constructor(
+    public color: Colors,
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    if (a <= 0 || b <= 0 || c <= 0 || a >= b + c || b >= a + c || c >= a + b) {
+      throw new Error('Invalid data');
```
Olexandr6, line 24: Let's make this error msg more informative. Fix this in all places

src/figuresClasses.ts
```diff
@@ -1,19 +1,91 @@
...
+type Colors = 'red' | 'green' | 'blue';
+
+enum Shapes {
+  triangle,
+  circle,
+  rectangle,
+}
 
+export interface Figure {
+  shape?: Shapes;
```
Olexandr6, line 10: Why is optional?
```suggestion
  shape: Shapes;
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,91 @@
-export interface Figure {
+type Colors = 'red' | 'green' | 'blue';
+
+enum Shapes {
```
Olexandr6, line 3: Don't name type in the plural. Type can be only one triangle or circle or rectangle, no several
```suggestion
enum Shape {
```

Olexandr6: Fix this in all places

src/figuresClasses.ts
```diff
@@ -1,19 +1,91 @@
...
+    public b: number,
+    public c: number,
+  ) {
+    if (a <= 0 || b <= 0 || c <= 0 || a >= b + c || b >= a + c || c >= a + b) {
+      throw new Error('Invalid data');
+    }
+  }
+
+  getArea(): number {
+    const s: number = (this.a + this.b + this.c) / 2;
```
Olexandr6, line 29: What is `s`? Rename this var

src/figuresClasses.ts
```diff
@@ -1,19 +1,91 @@
...
+    const area: number = figure.getArea();
+
+    return `A ${figure.color} circle - ${area}`;
+  }
+
+  if (figure instanceof Triangle) {
+    const area: number = figure.getArea();
+
+    return `A ${figure.color} triangle - ${area}`;
+  }
```
Olexandr6, line 88: You don't need to check `instanceof` for each figure. All figures have` figure.color`, `figure.shape` and `figure.getArea()` So let's write this function in one line

Olexandr6 (changes requested):
Almost done

Olexandr6 (approved):

PR #434, opened 2022-06-29
solution

IvanGrabovsky:


src/figuresClasses.ts
```diff
@@ -1,19 +1,82 @@
...
+export class Rectangle implements Figure {
+  shape: Shape = 'rectangle';
+
+  constructor(
+    public color: Color,
+    public a: number,
+    public b: number,
+  ) {
+    if (this.a <= 0 || this.b <= 0) {
+      throw new Error('all sides of rectangle can be positive');
```
Olexandr6, line 71:
```suggestion
      throw new Error('all sides of a rectangle must be positive numbers');
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,82 @@
-export interface Figure {
+type Shape = 'triangle' | 'circle' | 'rectangle';
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,82 @@
...
+  ) {
+    if (this.a <= 0 || this.b <= 0 || this.c <= 0) {
+      throw new Error('all sides can be positive numbers');
+    }
+
+    if (this.a + this.b <= this.c
+      || this.b + this.c <= this.a
+      || this.a + this.c <= this.b
+    ) {
+      throw new Error('all sides can be positive numbers');
```
Olexandr6, line 27: It's not a correct error. Let's make the error msg more informative. [Checklist](https://github.com/mate-academy/ts_figures_classes/blob/master/checklist.md) the 3rd point

src/figuresClasses.ts
```diff
@@ -1,19 +1,82 @@
...
-export class Rectangle {
+export class Circle implements Figure {
+  shape: Shape = 'circle';
 
+  constructor(
+    public color: Color,
+    public radius: number,
+  ) {
+    if (this.radius <= 0) {
+      throw new Error('radius can be a positive number');
```
Olexandr6, line 53:
```suggestion
      throw new Error('radius must be a positive number');
```

Olexandr6 (changes requested):
Almost done🔥

PR #437, opened 2022-06-29
solution was added

Bielichenko:


Olexandr6 (approved):

src/figuresClasses.ts
```diff
@@ -1,19 +1,73 @@
-export interface Figure {
+type Shape = 'triangle' | 'circle' | 'rectangle';
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,73 @@
...
+      throw new Error('all sides must have a positive length');
+    }
+
+    if (a + b <= c || a + c <= b || b + c <= a) {
+      throw new Error('one side must be shorter than sum of the others sides');
+    }
+  }
 
+  getArea(): number {
+    const semiP = (this.a + this.b + this.c) / 2;
```
Olexandr6, line 29:
```suggestion
    const semiPerimetr = (this.a + this.b + this.c) / 2;
```

PR #444, opened 2022-06-30
solution

LAndrikevych:


src/figuresClasses.ts
```diff
@@ -1,19 +1,75 @@
-export interface Figure {
+type Shape = 'triangle' | 'circle' | 'rectangle';
```

Olexandr6 (changes requested):
Good job, small improvements left

Olexandr6 (approved):

PR #459, opened 2022-06-30
Solution

ik-web:


src/figuresClasses.ts
```diff
@@ -1,19 +1,83 @@
...
 
 export class Circle {
+  shape: Shape = 'circle';
+
+  constructor(
+    public color: Color,
+    public radius: number,
+  ) {
+    if (radius <= 0) {
+      throw new Error(errorMessage);
```
Olexandr6, line 54: This error msg is not suitable for circle, circle doesn't have sides

Olexandr6 (approved):

PR #465, opened 2022-07-11
solution

FANTAZER-NURE:


src/figuresClasses.ts
```diff
@@ -1,19 +1,102 @@
...
+
+    this.shape = 'triangle';
+    this.color = color;
+    this.a = a;
+    this.b = b;
+    this.c = c;
+  }
+
+  getArea():number {
+    const p:number = (this.a + this.b + this.c) / 2;
```
Bohdan-Maksymiuk, line 40:
```suggestion
    const p = (this.a + this.b + this.c) / 2;
```
the result of calculation will be number so typescript will know about that

Bohdan-Maksymiuk (changes requested):
Add types for
* a shape (triangle, circle or rectangle);
* a color (red, green or blue);
try to use enums

Bohdan-Maksymiuk (approved):

PR #470, opened 2022-07-14
Solution figures classes

DimaShm:


src/figuresClasses.ts
```diff
@@ -1,19 +1,85 @@
...
 
-}
+  constructor(
+    public color: Color,
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    if (this.a <= 0 || this.b <= 0 || this.c <= 0) {
+      throw new Error('Incorrect value');
```
valerii-derkach, line 28: If you throw an error try to provide its detailed description in error message.

src/figuresClasses.ts
```diff
@@ -1,19 +1,85 @@
...
-export class Rectangle {
+export class Circle implements Figure {
+  shape = Shape.circle;
+
+  constructor(
+    public color: Color,
+    public radius: number,
+  ) {
+    if (this.radius <= 0) {
+      throw new Error('Incorrect value');
```
valerii-derkach, line 54: If you throw an error try to provide its detailed description in error message.

src/figuresClasses.ts
```diff
@@ -1,19 +1,85 @@
...
+export class Rectangle implements Figure {
+  shape = Shape.rectangle;
+
+  constructor(
+    public color: Color,
+    public width: number,
+    public height: number,
+  ) {
+    if (this.width <= 0 || this.height <= 0) {
+      throw new Error('Incorrect value');
```
valerii-derkach, line 72: If you throw an error try to provide its detailed description in error message.

src/figuresClasses.ts
```diff
@@ -1,19 +1,85 @@
...
 
-export class Circle {
+    if (max >= (this.a + this.b + this.c - max)) {
+      throw new Error('Sides of triangle not correct');
+    }
+  }
 
+  getArea(): number {
+    const p: number = (this.a + this.b + this.c) / 2;
+    const s: number = Math.sqrt(p * (p - this.a) * (p - this.b) * (p - this.c));
```
valerii-derkach, line 40: use more readable names for variables, e.g. `area`
you can use a, b, c for sides it's ok, but for area and perimeter better use full names
and check it in whole code

src/figuresClasses.ts
```diff
@@ -1,19 +1,85 @@
...
+  ) {
+    if (this.width <= 0 || this.height <= 0) {
+      throw new Error('Incorrect value');
+    }
+  }
+
+  getArea(): number {
+    const s: number = this.width * this.height;
+
+    return +s.toFixed(2);
```
valerii-derkach, line 79: create a function for rounding values and call it where it is necessary
check if it works properly, when you have 1.5 it shouldn't be 1.50 after rounding

valerii-derkach (changes requested):

src/figuresClasses.ts
amitpatiljc, line 40: Let's de-structuring and reduce the character-count:
const { a, b, c } = this;

amitpatiljc (approved):
Good Job!

ArturErmolenko (approved):

PR #477, opened 2022-07-22
Solution

fledith:


src/figuresClasses.ts
```diff
@@ -1,19 +1,70 @@
...
+    }
+
+    if ((a + b) <= c || (a + c) <= b || (b + c) <= a) {
+      throw new Error(`sides ${a}, ${b} and ${c} can't form a ${this.shape}`);
+    }
+  }
 
-export class Circle {
+  getArea(): number {
+    const s = (this.a + this.b + this.c) / 2;
```
BudnikOleksii, line 29: Avoid using one-letter variables

BudnikOleksii (approved):
Good job! Just rename one variable

src/figuresClasses.ts
Ihor-Karpyn, line 29: What does `s` mean? Is it cartoon south park or substring?

Ihor-Karpyn (changes requested):
Variables naming

Ihor-Karpyn (approved):

PR #478, opened 2022-07-22
figures

ghost:


src/figuresClasses.ts
```diff
@@ -1,19 +1,91 @@
...
+    }
+  }
 
+  getArea(): number {
+    const perimeter = (this.a + this.b + this.c) / 2;
+    const area: number = Math.sqrt(perimeter * (perimeter - this.a)
+      * (perimeter - this.b)
+      * (perimeter - this.c));
+
+    return Math.floor(area * 100) / 100;
```
alexandra-protyanova, line 38: u can move logic of rounding to separate function

alexandra-protyanova (approved):
Good job

PR #486, opened 2022-07-26
Solution

LordAsmodey:


OlgaStakhova (approved):
Як на мене гарна робота

src/figuresClasses.ts
```diff
@@ -1,19 +1,76 @@
...
+    public width: number,
+    public height: number,
+  ) {
+    if (width <= 0 || height <= 0) {
+      throw new Error('The sides of a rectangle cannot be zero or negative!');
+    }
+  }
+
+  getArea(): number {
+    return this.width * this.height;
```
ivan-kadykalo, line 70: Probably we should also round this value

ivan-kadykalo (approved):
🔥

PR #498, opened 2022-07-26
solution

Vova-Balyshev:


rodiontseva (approved):
nice!
I'm not sure, maybe better to use the same type of calculation to take rounded down to hundredths for all figures.

ivan-kadykalo (approved):
GJ🔥

PR #512, opened 2022-07-27
Solution

LovelyDrug:


src/figuresClasses.ts
```diff
@@ -1,19 +1,77 @@
 export interface Figure {
-
+  shape: string;
+  color: string;
```
HennadiiKupr, line 3: I would write types for shapes and colors. So that no additional data can be transmitted.

HennadiiKupr:
GJ

src/figuresClasses.ts
VETALEK: even not types, but enums 🤔

VETALEK (changes requested):
GJ 👍 
describe your types not as `string`s but as enums

VETALEK (approved):
🔥

PR #516, opened 2022-07-27
Solution

blcts:


Lentyaeev (approved):
Good use of static methods and overall good work🤩

src/figuresClasses.ts
```diff
@@ -1,19 +1,105 @@
...
 
-export class Rectangle {
+export class Circle implements Figure {
+  public shape: Shape;
 
+  constructor(
+    public color: Color,
+    public radius: number,
+  ) {
+    this.shape = Shape.Circle;
```
ivan-kadykalo, line 61: You can assign this value to your shape immediately
Here and in other figures.
```suggestion
  public shape: Shape = Shape.Circle;

  constructor(
    public color: Color,
    public radius: number,
  ) {
```

ivan-kadykalo (approved):
🚀

PR #519, opened 2022-07-28
solution

roman-romanenko:


src/figuresClasses.ts
```diff
@@ -1,19 +1,70 @@
...
+  ) {
+    if (radius <= 0) {
+      throw new Error(`radius ${radius}can not form a circle`);
+    }
+  }
 
+  getArea(): number {
+    return +(Math.floor(Math.PI * (this.radius ** 2) * 100) / 100).toFixed(2);
+  }
 }
```
AleksandrYuschenko, line 48:
```suggestion
  getArea(): number {
    return +(Math.floor(Math.PI * (this.radius ** 2) * 100) / 100);
  }
}
```
I think you don't need it here

src/figuresClasses.ts
```diff
@@ -1,19 +1,70 @@
+type Shape = 'triangle'| 'circle' | 'rectangle';
```
ivan-kadykalo, line 1: Will be better to create enum

src/figuresClasses.ts
```diff
@@ -1,19 +1,70 @@
...
 export interface Figure {
+  shape: Shape;
+  color: Color;
 
+  getArea(): number;
 }
 
-export class Triangle {
+export class Triangle implements Figure {
+  shape: Shape = 'triangle';
```
ivan-kadykalo, line 12: and use this enum here
```suggestion
  shape: Shape = Shape.Triangle;
```

ivan-kadykalo (changes requested):

VETALEK (approved):
🔥

PR #521, opened 2022-07-29
solution

DarkMistyRoom:


src/figuresClasses.ts
```diff
@@ -1,19 +1,80 @@
...
+export class Rectangle implements Figure {
+  shape: Shape = 'rectangle';
+
+  constructor(
+    public color: Color,
+    public width: number,
+    public height: number,
+  ) {
+    if (width <= 0 || height <= 0) {
+      throw new Error('error');
```
valerii-derkach, line 69: If you throw an error try to provide its detailed description in error message

valerii-derkach (changes requested):

src/figuresClasses.ts
```diff
@@ -1,19 +1,84 @@
...
+    const sumOfLowSides = [a, b, c]
+      .filter((side) => side !== maxSide)
+      .reduce((sum, side) => sum + side);
 
-export class Circle {
+    if (a <= 0 || b <= 0 || c <= 0 || maxSide >= sumOfLowSides) {
+      throw new Error(
+        'Incorrect data: impossible to build a triangle with this values',
+      );
+    }
```
Anna-Tkachenko, line 32: You check that input values are bigger than 0 for each class
Will be better to create separate function, that will accept initial parameters, check it and return the corresponding error if not

Anna-Tkachenko (changes requested):

alexandra-protyanova (approved):
Well done 👍

PR #527, opened 2022-08-08
Solution

vyurkov:


src/figuresClasses.ts
```diff
@@ -1,19 +1,90 @@
...
+      throw new Error();
+    }
+    this.shape = 'rectangle';
+    this.color = color;
+    this.width = width;
+    this.height = height;
+  }
+
+  getArea(): number {
+    return Math.floor((this.width * this.height) * 100) / 100;
```
alexandra-protyanova, line 84: it is better to remove logic of rounding to separate function & to use it everywhere

alexandra-protyanova (changes requested):

Bohdan-Maksymiuk (approved):

PR #533, opened 2022-08-19
Solution

aleksandra-prognimak:


src/figuresClasses.ts
```diff
@@ -1,19 +1,77 @@
...
+    public color: Color,
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    if (a <= 0 || b <= 0 || c <= 0) {
+      throw new Error('Invalid side length');
+    }
+
+    if ((a + b) <= c || (a + c) <= b || (b + c) <= a) {
```
Dmytro-Matsehora, line 23: I think that here it is better to make line breaks before each operator | | for better readability, but it's up to you

Dmytro-Matsehora (approved):
Well done 🔥

AlinaMatuschak (approved):
GJ!

PR #535, opened 2022-08-22
solution

Kulchevych:


src/figuresClasses.ts
```diff
@@ -1,19 +1,86 @@
...
-
+export interface Figure {
+  shape: Shape;
+  color: Color;
+  getArea(): number;
 }
 
-export class Circle {
+export class Triangle implements Figure {
+  public shape;
```
tiserett, line 16: I think that would be better to initialize shape with value
```suggestion
  public shape = Shape.type;
```

tiserett (approved):
well done

src/figuresClasses.ts
dellachaise, line 16: Why do you need this and 32 line?

You could simply write here 👇 
```suggestion
  shape = Shape.Triangle;
```

dellachaise (changes requested):
Great start 👍 

src/figuresClasses.ts
```diff
@@ -1,19 +1,85 @@
...
+  }
 
+  getArea(): number {
+    const semiPerimetr = (this.a + this.b + this.c) / 2;
+    const area = Math.sqrt(
+      semiPerimetr * (semiPerimetr - this.a)
+        * (semiPerimetr - this.b) * (semiPerimetr - this.c),
+    );
+
+    return Math.floor(area * 100) / 100;
```
dellachaise, line 44: Optional: you round with the same logic twice over the code. You also could create a separate helper function - `rounder` for example for this purpose.

dellachaise (approved):
Well done 👍

PR #540, opened 2022-08-25
Solution

VolodymyrChuyko:


AndreyJigijalkin (approved):
Looking good:) You can use explicit form indication in classes like `public shape: Shape = 'rectangle` :)

VolodymyrChuyko:
> 

As I understand it, using enum members is a better solution than using literal types. When you need to change its value in your code you have to change it only in one place.

src/figuresClasses.ts
```diff
@@ -1,19 +1,81 @@
...
+  ) {
+    const maxSide: number = Math.max(a, b, c);
+
+    if (maxSide >= a + b + c - maxSide) {
+      throw new Error(`sides ${a}, ${b} and ${c} can't form a triangle`);
+    }
+  }
 
+  getArea(): number {
+    const s: number = (this.a + this.b + this.c) / 2;
```
dellachaise, line 36: in this case better rename `s` to `semiPerimeter`.

src/figuresClasses.ts
```diff
@@ -1,19 +1,81 @@
...
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    const maxSide: number = Math.max(a, b, c);
+
+    if (maxSide >= a + b + c - maxSide) {
+      throw new Error(`sides ${a}, ${b} and ${c} can't form a triangle`);
+    }
+  }
```
dellachaise, line 34: You also should check in Triangle is all sides greater then 0.

Better create a helper function that will accept arguments (any length) and determine if some have `value <= 0`.
```
(...args) => args.some() ....
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,81 @@
...
+    public color: Color,
+    public radius: number,
+  ) {
+    if (radius <= 0) {
+      throw new Error('radius should be greater than 0');
+    }
+  }
 
+  getArea(): number {
+    return Math.floor(100 * Math.PI * this.radius ** 2) / 100;
```
dellachaise, line 57: Optional: you round with the same logic twice over the code. You also could create a separate helper function - `rounder` for example for this purpose.

dellachaise (changes requested):
Great job 👍 

src/figuresClasses.ts
```diff
@@ -1,19 +1,94 @@
...
 }
 
-export class Triangle {
+export interface Figure {
+  shape: Shape;
+  color: Color;
+  getArea(): number;
+}
 
+function haveIvalidValue(...args: number[]): boolean {
```
dellachaise, line 19:
```suggestion
function haveInvalidValue(...args: number[]): boolean {
```

dellachaise (approved):
Great job 🚀

PR #541, opened 2022-08-25
Solution

Githi54:


src/figuresClasses.ts
```diff
@@ -1,19 +1,84 @@
...
-}
+  constructor(
+    public color: Color,
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    if (a <= 0 || b <= 0 || c <= 0) {
+      throw new Error('Sides must be greater than zero');
+    }
```
sergii-nosachenko, line 31: You can create a helper function that will accepts arguments (any length) and determive if some has value <= 0:
```
(...args) => args.some() ....
```
And you can use it for any figure you have to check :)

src/figuresClasses.ts
```diff
@@ -1,19 +1,84 @@
...
+}
 
+function isBadTriangle(
+  side1: number,
+  side2: number,
+  side3: number,
+): boolean {
+  return (side1 + side2) <= side3
+    || (side1 + side3) <= side2
+    || (side2 + side3) <= side1;
```
sergii-nosachenko, line 17: You can check it by getting largest side with `Math.max(a, b, c)` and calculating only once:
`(a + b + c - largest) <= largest`

src/figuresClasses.ts
```diff
@@ -1,19 +1,84 @@
+type Color = 'red' | 'green' | 'blue';
+type Shapes = 'triangle' | 'circle' | 'rectangle';
+
 export interface Figure {
+  shape: Shapes;
+  color: Color;
+  getArea(): number;
+}
 
+function isBadTriangle(
```
sergii-nosachenko, line 10: Maybe you can make this function as a method of Circle. It looks like it is a part of class. But it's on you own decision.

src/figuresClasses.ts
```diff
@@ -1,19 +1,84 @@
+type Color = 'red' | 'green' | 'blue';
+type Shapes = 'triangle' | 'circle' | 'rectangle';
```
sergii-nosachenko, line 2: Better use enum here - it will give you more control and flexibility for Shapes type.

sergii-nosachenko (changes requested):
Have found some issues, need to be fixed.

sergii-nosachenko (approved):
Good job ✌

dellachaise (approved):
Great job 👍

PR #543, opened 2022-08-25
solution

AlexanderKolomiiets:


Kulchevych (approved):
Good job

src/figuresClasses.ts
```diff
@@ -1,19 +1,77 @@
+type Shape = 'triangle' | 'circle' | 'rectangle';
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,77 @@
...
+  constructor(
+    public color: Color,
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    if (this.a <= 0
+      || this.b <= 0
+      || this.c <= 0) {
+      throw new Error('Invalid value');
```
dellachaise, line 21: You can create a helper function that will accept arguments (any length) and determine if some have `value <= 0`.

src/figuresClasses.ts
```diff
@@ -1,19 +1,77 @@
...
+      throw new Error('Sum of two sides is shorter than hypotenuse');
+    }
+  }
 
+  getArea(): number {
+    const s: number = (this.a + this.b + this.c) / 2;
+
+    return +Math.sqrt(
+      s * (s - this.a) * (s - this.b) * (s - this.c),
+    ).toFixed(2);
```
dellachaise, line 36: Use meaningful variable name instead of `s`.

dellachaise (changes requested):
Great job 👍 

dellachaise (approved):
Well done 👍

PR #546, opened 2022-08-26
Task solution

sergii-nosachenko:


AndreyJigijalkin (approved):
🔥

src/figuresClasses.ts
```diff
@@ -1,19 +1,112 @@
...
+export interface Figure {
+  shape: Shapes;
+  color: Colors;
+  getArea(): number;
 }
 
-export class Triangle {
+abstract class BaseFigure implements Figure {
+  protected isOk: boolean = true;
+
```
AlinaMatuschak, line 17: Is it necessary? You can just throw an error if needed it. Also, I believe that you can find better name )

AlinaMatuschak (approved):
🚀🚀🚀

src/figuresClasses.ts
sergii-nosachenko: @AlinaMatuschak Good advice) Now it's perfect)

PR #549, opened 2022-08-29
solution

AndreyJigijalkin:


src/figuresClasses.ts
```diff
@@ -1,19 +1,79 @@
...
 
-export class Rectangle {
+export class Circle implements Figure {
+  shape: Shape = 'circle';
+
+  constructor(
+    public color: Color,
+    public radius: number,
+  ) {
+    if (this.radius < 0) {
```
VolodymyrChuyko, line 49:
```suggestion
    if (this.radius <= 0) {
```
The constructors should `throw new Error('your error message')` if any length is <= 0

src/figuresClasses.ts
```diff
@@ -1,19 +1,79 @@
...
+      throw new Error('Invalid radius length.');
+    }
+  }
 
+  getArea(): number {
+    return +Math.floor(Math.PI * this.radius ** 2 * 100) / 100;
```
VolodymyrChuyko, line 55:
```suggestion
    return Math.floor(Math.PI * this.radius ** 2 * 100) / 100;
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,79 @@
...
+      || this.c <= 0) {
+      throw new Error('Invalid value.');
+    }
 
-export class Circle {
+    if (this.a + this.b <= this.c
+      || this.a + this.c <= this.b
+      || this.b + this.c <= this.a) {
+      throw new Error('Invalid sizes length.');
+    }
```
VolodymyrChuyko, line 30: There is a simple way to check if the triangle sides are valid: just compare the maxSide with (a + b + c - maxSide), but it's up to you

VolodymyrChuyko (approved):
Good job! Just small issues.

src/figuresClasses.ts
```diff
@@ -1,19 +1,79 @@
+type Shape = 'triangle' | 'circle' | 'rectangle';
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,79 @@
...
+    public color: Color,
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    if (this.a <= 0
+      || this.b <= 0
+      || this.c <= 0) {
+      throw new Error('Invalid value.');
+    }
```
dellachaise, line 24: Better create a helper function that will accept arguments (any length) and determine if some have `value <= 0`.
```
(...args) => args.some() ....
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,79 @@
...
+    const largestSide: number = Math.max(this.a, this.b, this.c);
+
+    if ((this.a + this.b + this.c - largestSide) <= largestSide) {
+      throw new Error('Invalid sizes length.');
+    }
+  }
 
-export class Circle {
+  getArea(): number {
+    const p: number = (this.a + this.b + this.c) / 2;
```
dellachaise, line 34: Better use a more meaningful variable name instead of `p` - `semiPerimetr` for example.

src/figuresClasses.ts
```diff
@@ -1,19 +1,79 @@
...
+    public color: Color,
+    public radius: number,
+  ) {
+    if (this.radius <= 0) {
+      throw new Error('Invalid radius length.');
+    }
+  }
+
+  getArea(): number {
+    return Math.floor(Math.PI * this.radius ** 2 * 100) / 100;
```
dellachaise, line 55: Optional: you round with the same logic twice over the code. You also could create a separate helper function - `rounder` for example for this purpose.

dellachaise (changes requested):
Good job 👍 

Just fix a few issues 👇

dellachaise (approved):
Great job 🚀

PR #550, opened 2022-08-30
Solution

oksana-logos-frontend:


src/figuresClasses.ts
```diff
@@ -1,19 +1,75 @@
...
+  shape: Shape = 'triangle';
 
-}
+  constructor(
+    public color: Color,
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    if (this.a <= 0 || this.b <= 0 || this.c <= 0) {
```
Githi54, line 19: You can create a helper function that will accepts arguments (any length) and determive if some has value <= 0:

(...args) => args.some() ....
And you can use it for any figure you have to check :)

src/figuresClasses.ts
```diff
@@ -1,19 +1,75 @@
...
+      throw new Error('Invalid value');
+    }
+
+    if (this.a + this.b <= this.c
```
Githi54, line 23: You can check it by getting largest side with Math.max(a, b, c) and calculating only once:
(a + b + c - largest) <= largest

Githi54 (changes requested):
Good start!🚀

Githi54 (approved):
Great!🚀

src/figuresClasses.ts
```diff
@@ -1,19 +1,83 @@
...
+    if ((this.a + this.b + this.c - largest) <= largest) {
+      throw new Error('Sides 1, 2 and 3 can not form a triangle');
+    }
+  }
 
+  getArea(): number {
+    const sPerimeter : number = (this.a + this.b + this.c) / 2;
+
+    return Math.floor(Math.sqrt(sPerimeter * (sPerimeter - this.a)
+    * (sPerimeter - this.b) * (sPerimeter - this.c)) * 100) / 100;
```
AlinaMatuschak, line 42: Hard to read 🥲

AlinaMatuschak (approved):
🔥

src/figuresClasses.ts
oksana-logos-frontend: renamed)

PR #552, opened 2022-08-30
Solution

Jamal-Amoian:


src/figuresClasses.ts
```diff
@@ -1,19 +1,84 @@
...
+    public width: number,
+    public height: number,
+  ) {
+    if (width <= 0 || height <= 0) {
+      throw new Error('Side less or equals zero');
+    }
+  }
+
+  getArea(): number {
+    return Math.floor((this.width * this.height) * 100) / 100;
```
aleksandra-prognimak, line 78:
```suggestion
    return this.width * this.height;
```
Тут можно не округлять)

aleksandra-prognimak (approved):
Wonderful🌺

src/figuresClasses.ts
```diff
@@ -1,19 +1,84 @@
...
+
+  constructor(
+    public color: Color,
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    if (a <= 0 || b <= 0 || c <= 0) {
+      throw new Error('One of sides less or equals zero');
+    }
```
dellachaise, line 26: Better create a helper function that will accept arguments (any length) and determine if some have `value <= 0`.
```
(...args) => args.some() ....
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,84 @@
...
+    public radius: number,
+  ) {
+    if (radius <= 0) {
+      throw new Error('Radius less or equals zero');
+    }
+  }
 
+  getArea(): number {
+    return Math.floor(Math.PI * (this.radius ** 2) * 100) / 100;
+  }
```
dellachaise, line 61: Optional: you round with the same logic twice over the code. You also could create a separate helper function - `rounder` for example for this purpose.

dellachaise (changes requested):
Great job 👍 

Just create a minimum 1 helper function, it will help you not repeat almost the same code.

dellachaise (approved):
Well done 👍

PR #554, opened 2022-08-30
Classes Triange, Circle and Rectangle implemented

andrii-dmytruk:


src/figuresClasses.ts
```diff
@@ -1,19 +1,77 @@
...
+    public color: Color,
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    if (
+      (a + b <= c || a + c <= b || b + c <= a)
+      || (a <= 0 || b <= 0 || c <= 0)
+    ) {
+      throw new Error('Wrong figure parameters');
```
zinchenko-anastasiia, line 23: i suppose would be better if you create 2 different errors. 'cause user does not understand what mistake he made

zinchenko-anastasiia (changes requested):

zinchenko-anastasiia (approved):
good job!)

src/figuresClasses.ts
```diff
@@ -1,19 +1,78 @@
-export interface Figure {
+type Shape = 'triangle' | 'circle' | 'rectangle';
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,78 @@
...
-}
+  constructor(
+    public color: Color,
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    if (a <= 0 || b <= 0 || c <= 0) {
+      throw new Error('a, b and c must be > 0');
+    }
```
dellachaise, line 21: Better create a helper function that will accept arguments (any length) and determine if some have `value <= 0`.
```
(...args) => args.some() ....
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,78 @@
...
+      throw new Error('The longest side must be >= than a sum of two others');
+    }
+  }
 
+  getArea(): number {
+    const s = (this.a + this.b + this.c) / 2;
+
+    const area = Math.sqrt(s * (s - this.a) * (s - this.b) * (s - this.c));
+
+    return Math.floor(100 * area) / 100;
```
dellachaise, line 33: You round with the same logic 3 times over the code. You also could create a separate helper function - `rounder` for example for this purpose.

src/figuresClasses.ts
```diff
@@ -1,19 +1,78 @@
...
+    }
 
-export class Circle {
+    if (a + b <= c || a + c <= b || b + c <= a) {
+      throw new Error('The longest side must be >= than a sum of two others');
+    }
+  }
 
+  getArea(): number {
+    const s = (this.a + this.b + this.c) / 2;
```
dellachaise, line 29: `s` variable name is not obvious, better rename to `semiPerimeter` for example.

dellachaise (changes requested):
Great job 👍 

src/figuresClasses.ts
```diff
@@ -1,19 +1,96 @@
...
 
+export interface Figure {
+  shape: Shape;
+  color: Color;
+  getArea(): number;
 }
 
-export class Circle {
+const checkIfValid = (...args: number[]): boolean => (
+  args.every((el: number) => el > 0));
```
dellachaise, line 20: `value or side` - better name than `el`.

dellachaise (approved):
Well done 👍

PR #556, opened 2022-08-31
figures classes solution

artem-kodesnikov:


src/figuresClasses.ts
```diff
@@ -1,19 +1,82 @@
-export interface Figure {
+type Color = 'red' | 'green' | 'Bundle';
```
plemiannik, line 1: blue

plemiannik (approved):
excellent ✈️

AndriiRut (changes requested):
have to fix according to the coment

AndriiRut (approved):
:rocket

PR #560, opened 2022-08-31
Solution

Mejeezz:


ibondrk (approved):
GJ🔥

src/figuresClasses.ts
```diff
@@ -1,19 +1,82 @@
+type Shape = 'triangle' | 'circle' | 'rectangle';
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,82 @@
...
+type Color = 'red' | 'green' | 'blue';
+
 export interface Figure {
+  shape: Shape,
+  color: Color,
 
+  getArea(): number;
```
dellachaise, line 8:
```suggestion
  color: Color,
  getArea(): number;
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,82 @@
...
+    public color: Color,
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    if (this.a <= 0
+     || this.b <= 0
+     || this.c <= 0) {
+      throw new Error('incorrect figure parameters');
+    }
```
dellachaise, line 24: Better create a helper function that will accept arguments (any length) and determine if some have `value <= 0`.
```
(...args) => args.some() ....
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,82 @@
...
 
-export class Circle {
+  getArea(): number {
+    const perimeter = (this.a + this.b + this.c) / 2;
+    const area = Math.sqrt(perimeter
+      * (perimeter - this.a)
+      * (perimeter - this.b)
+      * (perimeter - this.c));
 
+    return Math.round(area * 100) / 100;
```
dellachaise, line 40: Optional: you round with the same logic twice over the code. You also could create a separate helper function - `rounder` for example for this purpose.

dellachaise (changes requested):
Great job 👍 

dellachaise (approved):
Well done 👍

PR #562, opened 2022-08-31
Solution

OlehMy:


kir8mir (approved):
GJ!!!

src/figuresClasses.ts
```diff
@@ -1,19 +1,77 @@
...
+    public color: string,
+    public radius: number,
+  ) {
+    if (radius <= 0) {
+      throw new Error('Radius length is not correct');
+    }
+  }
 
+  getArea(): number {
+    return Math.floor(Math.PI * this.radius ** 2 * 100) / 100;
```
AlinaMatuschak, line 53: you round down to hundredths on each getArea, maybe better create a separate method for it?

AlinaMatuschak (approved):

src/figuresClasses.ts
OlehMy: Maybe you're right)))

PR #563, opened 2022-08-31
Solution for Figura Class

OlgaOchichenko:


src/figuresClasses.ts
```diff
@@ -1,19 +1,98 @@
...
+      throw new Error(
+        'side "c" must be less than the sum of the other two sides',
+      );
+    }
+
+    if (this.b >= this.a + this.c) {
+      throw new Error(
+        'side "b" must be less than the sum of the other two sides',
+      );
+    }
```
AlinaMatuschak, line 41: can you combine it? You can throw `The sum of sides less than the third side` or something like that

src/figuresClasses.ts
```diff
@@ -1,19 +1,98 @@
...
 
+    const area: number = (Math.sqrt(
+      semiPerimeter
+      * (semiPerimeter - this.a)
+      * (semiPerimeter - this.b)
+      * (semiPerimeter - this.c),
+    )
+    );
+
+    return Math.floor(area * 100) / 100;
```
AlinaMatuschak, line 55: you round down to hundredths on each getArea, maybe better create a separate method for it?

AlinaMatuschak (changes requested):

src/figuresClasses.ts
```diff
@@ -1,19 +1,90 @@
-export interface Figure {
+// import { type } from "os";
```
AlinaMatuschak, line 1: delete comments

src/figuresClasses.ts
AlinaMatuschak: Why didn’t you create a separate function for rounds down to hundredths? You duplicate code

AlinaMatuschak (approved):

PR #564, opened 2022-08-31
Implement Figure Classes and Funcs in TS

max-kravchenko:
Implement Figure Classes and Functions in TS


src/figuresClasses.ts
```diff
@@ -1,19 +1,81 @@
...
+      throw new Error('Triangle sides should be greater than 0');
+    }
+
+    if (a >= b + c || b >= a + c || c >= a + b) {
+      throw new Error('Cannot form a triangle. One side is too long bro/sis');
+    }
+  }
 
+  getArea(): number {
+    const s = (this.a + this.b + this.c) / 2;
```
dellachaise, line 36: better rename to more obvious `semiPerimetr`

src/figuresClasses.ts
```diff
@@ -1,19 +1,81 @@
...
+
+    const area = Math.sqrt(s * ((s - this.a) * (s - this.b) * (s - this.c)));
+
+    return Math.floor(area * 100) / 100;
```
dellachaise, line 40: you round with the same logic twice over the code. You also could create a separate helper function - `rounder` for example for this purpose.

dellachaise (changes requested):
Great job 👍 

Just add a few little improvements :po

src/figuresClasses.ts
max-kravchenko: Thank you for comments!

dellachaise (approved):
Well done 🚀

PR #567, opened 2022-08-31
Solution

anna-kolmakova:


max-kravchenko (approved):
Great solution!
p.s. In Circle error I would add smth like 'Value should be greater than 0'

src/figuresClasses.ts
```diff
@@ -1,19 +1,84 @@
...
+
+  constructor(
+    public color: Color,
+    public sideA: number,
+    public sideB: number,
+    public sideC: number,
+  ) {
+    if ([sideA, sideB, sideC].some((side) => side === 0)) {
+      throw new Error('Side is less than 0');
+    }
```
dellachaise, line 30: Better create a helper function that will accept arguments (any length) and determine if some have `value <= 0`.
```
(...args) => args.some() ....
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,84 @@
...
+      || sideC >= sideA + sideB) {
+      throw new Error('One side is longer than sum of two others');
+    }
+  }
 
+  getArea(): number {
+    const semiPerimeter = (this.sideA + this.sideB + this.sideC) / 2;
+
+    return +Math.sqrt(semiPerimeter * (semiPerimeter - this.sideA)
+    * (semiPerimeter - this.sideB) * (semiPerimeter - this.sideC)).toFixed(2);
```
dellachaise, line 43: Optional: you could round with the same logic in 3 functions over the code. You also could create a separate helper function - `rounder` for example for this purpose. (use `toFixed(2) or Math.floor like in line 60`)

dellachaise (changes requested):
Great job 👍 

Just add a minimum 1 helper function.

dellachaise (approved):
Well done 👍

PR #568, opened 2022-08-31
FiguresClasses solution

NataliyaAn:


src/figuresClasses.ts
```diff
@@ -1,19 +1,86 @@
...
+  color: Color;
+  getArea(): number;
 }
 
-export class Triangle {
+function roundedArea(area:number):number {
+  return Math.floor(area * 100) / 100;
+}
 
+function chackArg(arg: number):void {
```
Jamal-Amoian, line 19: correctly `checkArg`

src/figuresClasses.ts
```diff
@@ -1,19 +1,86 @@
...
+  ) {
+    [width, height].forEach((side):void => chackArg(side));
+  }
+
+  public getArea():number {
+    return roundedArea(this.width * this.height);
+  }
+}
 
+export function getInfo(figure:Rectangle|Circle|Triangle):string {
```
Jamal-Amoian, line 84: You should use here `figure: Figure`, also, please, add everywhere spaces, like that `export function getInfo(figure: Rectangle | Circle | Triangle): string`

Jamal-Amoian (approved):
Good Job!

src/figuresClasses.ts
```diff
@@ -1,19 +1,87 @@
-export interface Figure {
 
+function roundedArea(area:number):number {
```
dellachaise, line 2: Great idea about name 🔥 
Just function names should start from the **verb**.
```suggestion
function roundArea(area:number):number {
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,87 @@
...
+  return Math.floor(area * 100) / 100;
+}
+
+function checkArg(arg: number):void {
+  if (arg <= 0) {
+    throw new Error('One of side or radius <= 0');
+  }
 }
```
dellachaise, line 10: Optional: you could rewrite `checkArg` function so that it will accept arguments (any length) and determine if some have `value <= 0`.
```
(...args) => args.some() ....
```

dellachaise (approved):
Great job 👍

PR #569, opened 2022-08-31
figures solution was created

ibondrk:


src/figuresClasses.ts
```diff
@@ -1,19 +1,80 @@
...
+export class Triangle implements Figure {
+  shape = Shape.Triangle;
+
+  constructor(
+    public color:Color,
+    public a:number,
+    public b:number,
+    public c:number,
+  ) {
+    if (a <= 0 || b <= 0 || c <= 0) {
```
Mejeezz, line 25: You can create a function to check if some of params will be <= 0; because you use this tree times in your code.
`(...args) => args.some() ....`

src/figuresClasses.ts
```diff
@@ -1,19 +1,80 @@
...
+enum Shape {
+  Triangle = 'triangle',
+  Circle = 'circle',
+  Rectangle = 'rectangle',
 }
 
-export class Triangle {
+export interface Figure {
+  shape: Shape;
+  color: Color;
```
Mejeezz, line 12: delete this space ^_^

Mejeezz (changes requested):

src/figuresClasses.ts
dellachaise: Agreed)

src/figuresClasses.ts
```diff
@@ -1,19 +1,86 @@
...
+export class Circle implements Figure {
+  shape = Shape.Circle;
+
+  constructor(
+    public color: Color,
+    public radius: number,
+  ) {
+    if (radius <= 0) {
+      throw new Error('Incorrect radius. Radius must be greater then zero');
+    }
```
dellachaise, line 58: You could use `isSideCorrect` here too, because you did almost the same check.

src/figuresClasses.ts
```diff
@@ -1,19 +1,86 @@
...
+  shape = Shape.Rectangle;
+
+  constructor(
+    public color: Color,
+    public width: number,
+    public height: number,
+  ) {
+    if (height <= 0 || width <= 0) {
+      throw new Error('width and height must be greater then zero');
+    }
```
dellachaise, line 76: You could use `isSideCorrect` here too, because you did almost the same check.

src/figuresClasses.ts
```diff
@@ -1,19 +1,86 @@
...
+    public color: Color,
+    public radius: number,
+  ) {
+    if (radius <= 0) {
+      throw new Error('Incorrect radius. Radius must be greater then zero');
+    }
+  }
 
+  getArea(): number {
+    return Math.floor(Math.PI * (this.radius ** 2) * 100) / 100;
```
dellachaise, line 62: Optional: you round twice over the code(43-46 lines too). You also could create a separate helper function - `rounder` for example for this purpose and round it using one way using - `Math.floor or toFixed()`.

src/figuresClasses.ts
```diff
@@ -1,19 +1,86 @@
...
+      throw new Error('Incorrect side length parameter');
+    }
+
+    if ((a >= b + c) || (b >= a + c) || (c >= a + b)) {
+      throw new Error(`${a}, ${b} and ${c} can't be a triangle`);
+    }
+  }
 
+  getArea():number {
+    const getHalf: number = (this.a + this.b + this.c) / 2;
```
dellachaise, line 41: You store in this variable `semiPerimetr` when I read `getHalf` I thought thet there should store a function that could help me to get a half. Because from verb starts names of the function. But here it isn't.

dellachaise (changes requested):
Good job 👍 

Just add improvements according to comments 👇

dellachaise (approved):
Well done 👍

PR #571, opened 2022-08-31
solution

KorniychukLesya:


max-kravchenko (approved):
Great solution!

src/figuresClasses.ts
```diff
@@ -1,19 +1,79 @@
...
+
+  constructor(
+    public color: Color,
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    if (a <= 0 || b <= 0 || c <= 0) {
+      throw new Error('side cannot be less than or equal to 0');
+    }
```
dellachaise, line 23: Better create a helper function that will accept arguments (any length) and determine if some have `value <= 0`.
```
(...args) => args.some() ....
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,79 @@
...
+
+    if (a + b <= c || b + c <= a || c + a <= b) {
+      throw new Error(`sides ${a}, ${b} and ${c} can't form a triangle`);
+    }
+  }
+
+  getArea(): number {
+    const s: number = (this.a + this.b + this.c) / 2;
```
dellachaise, line 31: `s` better rename to more obvious `semiPerimetr`.

KorniychukLesya:
Daria, thank you for your comments. I fixed everything☺

src/figuresClasses.ts
```diff
@@ -1,19 +1,101 @@
+// import { type } from "os";
```
dellachaise, line 1: Delete comment.

dellachaise (approved):
Well done 🚀

PR #574, opened 2022-08-31
solution 1

kostya-kosyuk:


src/figuresClasses.ts
```diff
@@ -1,19 +1,96 @@
-export interface Figure {
+type Shape = 'triangle' | 'circle' | 'rectangle';
```
andrii-dmytruk, line 1: better to use `enum` for Shape and Color

src/figuresClasses.ts
```diff
@@ -1,19 +1,96 @@
...
+      throw new Error(
+        'The longest side of a triangle is >= than a sum of two others',
+      );
+    }
+  }
+
+  getArea(): number {
+    const semiPerimeter = (this.a + this.b + this.c) / 2;
+    const area = Math.sqrt(semiPerimeter
+    * (semiPerimeter - this.a)
```
andrii-dmytruk, line 51: fix indents

src/figuresClasses.ts
```diff
@@ -1,19 +1,96 @@
...
+  * (semiPerimeter - this.b)
+  * (semiPerimeter - this.c));
+
+    return +area.toFixed(2);
```
andrii-dmytruk, line 55: use math.floor and implement heper function to calculate this, because the same operation repeats 3 times

src/figuresClasses.ts
```diff
@@ -1,19 +1,96 @@
...
+  shape: Shape = 'triangle';
 
-}
+  constructor(
+    public color: Color,
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    if (a <= 0 || b <= 0 || c <= 0) {
```
andrii-dmytruk, line 19: you can carry out this condition in separate function to prevent repeating same code. For example: `(...args) => args.some(el => el <= 0)`

andrii-dmytruk (changes requested):

andrii-dmytruk (approved):
Well done!

dellachaise (approved):
Well done 👍

src/figuresClasses.ts
```diff
@@ -1,19 +1,117 @@
...
+
+    if (longestSide >= sides.reduce((
+      sum: number,
+      currentSide:number,
+    ) => sum + currentSide)) {
+      throw new Error(
+        'The longest side of a triangle is >= than a sum of two others',
+      );
+    }
+  }
```
dellachaise, line 63: interesting solution, but better do it with one if and 3 conditions inside(3 lines of code).

PR #576, opened 2022-09-01
task solution code

rboiar:


src/figuresClasses.ts
```diff
@@ -1,19 +1,95 @@
...
+      .some((parameter) => parameter <= 0);
+
+    if (rectangleCheck) {
+      throw new Error('ERROR: width and hight must be >= than 0');
+    }
+  }
 
-export function getInfo(figure) {
+  getArea(): number {
+    return this.width * this.height;
```
NataliyaAn, line 89: if the width or height is a fraction, you will need to round the value. but the tests passed - everything is ok

NataliyaAn (approved):
Good job!

src/figuresClasses.ts
```diff
@@ -1,19 +1,95 @@
...
+export class Triangle implements Figure {
+  shape: Shape = Shape.Triangle;
+
+  constructor(
+    public color: Color,
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    if ([a, b, c].some((side) => side <= 0)) {
```
dellachaise, line 28: Better create a helper function that will accept arguments (any length) and determine if some have `value <= 0`.
```
(...args) => args.some() ....
```
And you could use it for any figure you need to check. Now almost the same check you have in 3 places all over the code.

src/figuresClasses.ts
dellachaise: Generally - better round, it will hangle more cases. But in this task - up to you.

src/figuresClasses.ts
```diff
@@ -1,19 +1,95 @@
...
+  getArea():number {
+    const semiPerimeter: number = (this.a + this.b + this.c) / 2;
+    const area = Math.sqrt(
+      semiPerimeter
+        * (semiPerimeter - this.a)
+        * (semiPerimeter - this.b)
+        * (semiPerimeter - this.c),
+    );
+
+    return Math.round(area * 100) / 100;
```
dellachaise, line 51: Optional: you round with the same logic twice over the code. You also could create a separate helper function - `rounder` for example for this purpose.

dellachaise (changes requested):
Great job 👍 

Just add a minimum one helper function.

src/figuresClasses.ts
rboiar: Yes, this is a really cool improvement

src/figuresClasses.ts
rboiar: 👍

PR #577, opened 2022-09-01
TS solution Figures Classes

wfurs0w:


src/figuresClasses.ts
```diff
@@ -1,19 +1,80 @@
...
+    public height: number,
+  ) {
+    if (height <= 0 || width <= 0) {
+      throw new Error('Sides must be greater than zero');
+    }
+  }
 
-export function getInfo(figure) {
+  getArea():number {
+    return this.width * this.height;
```
rboiar, line 74: I was advised to round these values, so you can also do that for improvement (optional)

rboiar (approved):
Good solution 👍

src/figuresClasses.ts
```diff
@@ -1,19 +1,80 @@
+type Shape = 'triangle' | 'circle' | 'rectangle';
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,80 @@
...
+
+  constructor(
+    public color: Color,
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    if (a <= 0 || b <= 0 || c <= 0) {
+      throw new Error('Side must be greater than zero');
+    }
```
dellachaise, line 22: Better create a helper function that will accept arguments (any length) and determine if some have `value <= 0`.
```
(...args) => args.some() ....
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,80 @@
...
+    const halfOfPerimeter: number = (this.a + this.b + this.c) / 2;
 
+    const triangleArea = Math.sqrt(
+      halfOfPerimeter
+      * (halfOfPerimeter - this.a)
+      * (halfOfPerimeter - this.b)
+      * (halfOfPerimeter - this.c),
+    );
+
+    return Math.floor(triangleArea * 100) / 100;
```
dellachaise, line 39: Optional: you round with the same logic twice over the code. You also could create a separate helper function - `rounder` for example for this purpose.

dellachaise (changes requested):
Great job 👍 

Just add a few fixes according to comments 👇

src/figuresClasses.ts
```diff
@@ -1,19 +1,85 @@
...
+  Circle = 'circle',
+  Rectangle = 'rectangle',
+}
+
 export interface Figure {
+  shape: Shape;
+  color: Color;
 
+  getArea(): number;
 }
```
AndriiRut, line 14:
```suggestion
export interface Figure {
  shape: Shape;
  color: Color;
  getArea(): number;
}
```

AndriiRut (approved):
🚀

PR #578, opened 2022-09-01
Solution

Aleksei-Evtushenko:


src/figuresClasses.ts
```diff
@@ -1,19 +1,93 @@
-export interface Figure {
+function MathFloor(value: number): number {
+  return Math.floor(value * 100) / 100;
+}
```
SanyaBratashchuk, line 3: choose a better name for this function

src/figuresClasses.ts
```diff
@@ -1,19 +1,93 @@
...
+    public SideC: number,
+  ) {
+    if (this.SideA <= 0 || this.SideB <= 0 || this.SideC <= 0) {
+      throw new Error('Side length can`t be 0');
+    }
+
+    if (this.SideA + this.SideB <= this.SideC
+      || this.SideB + this.SideC <= this.SideB
+      || this.SideC + this.SideB <= this.SideA) {
+      throw new Error('Longest side can`t be less than sum of two other sides');
```
SanyaBratashchuk, line 35:
```suggestion
      throw new Error('Longest side can`t be longer than sum of two other sides');
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,93 @@
...
+export class Rectangle implements Figure {
+  shape: Shape = Shape.Rectangle;
+
+  constructor(
+    public color: Color,
+    public width: number,
+    public height: number,
+  ) {
+    if (this.width <= 0 || this.height <= 0) {
+      throw new Error('Side length can`t be 0');
```
SanyaBratashchuk, line 80:
```suggestion
      throw new Error('Side length should be positive');
```

SanyaBratashchuk (changes requested):

src/figuresClasses.ts
```diff
@@ -1,19 +1,95 @@
-export interface Figure {
+function Rounding(value: number): number {
```
Bohdan-Maksymiuk, line 1:
```suggestion
function getRoundedValue(value: number): number {
```
name of function should describe action - what it do or what it give

src/figuresClasses.ts
```diff
@@ -1,19 +1,95 @@
...
+
+  constructor(
+    public color: Color,
+    public SideA: number,
+    public SideB: number,
+    public SideC: number,
+  ) {
+    if (this.SideA <= 0 || this.SideB <= 0 || this.SideC <= 0) {
+      throw new Error('Side length can`t be 0');
+    }
```
Bohdan-Maksymiuk, line 30:
```suggestion
    if (this.SideA <= 0 || this.SideB <= 0 || this.SideC <= 0) {
      throw new Error('Side length should be positive);
    }
```
You have check not only for 0 here

src/figuresClasses.ts
```diff
@@ -1,19 +1,95 @@
...
-export class Rectangle {
+export class Circle implements Figure {
+  shape: Shape = Shape.Circle;
 
+  constructor(
+    public color: Color,
+    public radius: number,
+  ) {
+    if (this.radius <= 0) {
+      throw new Error('Radius of circle can`t be 0');
```
Bohdan-Maksymiuk, line 62: same

Bohdan-Maksymiuk (changes requested):

Bohdan-Maksymiuk (approved):

PR #579, opened 2022-09-01
solution

oksana-kyryienko:


anyakrainiuk (approved):
Well done!

src/figuresClasses.ts
```diff
@@ -1,19 +1,84 @@
...
+  shape = Shapes.Triangle;
+
+  constructor(
+    public color: Color,
+
+    public a: number,
+
+    public b: number,
+
+    public c: number,
```
dellachaise, line 25: Why do you need these page breaks? The same in other constructors.
```suggestion
  constructor(
    public color: Color,
    public a: number,
    public b: number,
    public c: number,
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,84 @@
...
+  ) {
+    if (this.a <= 0 || this.b <= 0 || this.c <= 0) {
```
dellachaise, line 27: You don't need to use `this` inside of the constructor. It is redundant. The same in other constructors.
```suggestion
    if (a <= 0 || b <= 0 || c <= 0) {
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,84 @@
...
+  shape = Shapes.Circle;
+
+  constructor(
+    public color: Color,
 
+    public radius: number,
+  ) {
+    if (radius <= 0) {
+      throw new Error('Error radius <= 0');
+    }
```
dellachaise, line 54: Better create a helper function that will accept arguments (any length) and determine if some have `value <= 0`.
```
(...args) => args.some() ....
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,84 @@
...
+  }
+
+  getArea(): number {
+    return Math.floor(Math.PI * this.radius ** 2 * 100) / 100;
```
dellachaise, line 58: Optional: you round 2 times over the code, better use the same logic for it.  You also could create a separate helper function - `rounder` for example for this purpose.

dellachaise (changes requested):
Great job 👍 

Just add improvements according to comments 👇

AndriiRut (approved):
🔥

PR #580, opened 2022-09-01
Figures classes

Sasha-Gordijchuk:


Anastasiia-Khmelovska (approved):
gj

src/figuresClasses.ts
```diff
@@ -1,19 +1,80 @@
...
+    public sideA: number,
+    public sideB: number,
+    public sideC: number,
+  ) {
+    if (sideA <= 0
+      || sideB <= 0
+      || sideC <= 0
+    ) {
+      throw new Error('side must be greater than zero');
+    }
```
dellachaise, line 22: Better create a helper function that will accept arguments (any length) and determine if some have `value <= 0`.
```
(...args) => args.some() ....
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,80 @@
...
+      || sideC >= sideA + sideB
+    ) {
+      throw new Error(
+        'side of the triangle must be less than the sum of the other two',
+      );
+    }
+  }
+
+  getArea(): number {
+    const p = (this.sideA + this.sideB + this.sideC) / 2;
```
dellachaise, line 35: Variable names should be meaningful, that's why `p` is better rename to `semiPerimetr``.

src/figuresClasses.ts
```diff
@@ -1,19 +1,80 @@
...
+
+    return Math.floor(Math.sqrt(
+      p * (p - this.sideA) * (p - this.sideB) * (p - this.sideC),
+    ) * 100) / 100;
```
dellachaise, line 39: Optional: you round with the same logic twice over the code. You also could create a separate helper function - `rounder` for example for this purpose.

dellachaise (changes requested):
Great start 👍 

AndriiRut (approved):
good job! 🔥

PR #583, opened 2022-09-01
Figure Class solution

anyakrainiuk:


oksana-kyryienko (approved):

oksana-kyryienko:
Well done!

src/figuresClasses.ts
```diff
@@ -1,19 +1,87 @@
...
+  shape: Shape = 'triangle';
 
-}
+  constructor(
+    public color: Color,
+    public side1: number,
+    public side2: number,
+    public side3: number,
+  ) {
+    if (this.side1 <= 0 || this.side2 <= 0 || this.side3 <= 0) {
```
dellachaise, line 23: You don't need to write `this` inside of the constructor, it's redundant. The same in other constructors.
```suggestion
    if (side1 <= 0 || side2 <= 0 || side3 <= 0) {
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,87 @@
...
+
+    if (this.side1 >= (this.side2 + this.side3)
+    || this.side2 >= (this.side3 + this.side1)
+    || this.side3 >= (this.side1 + this.side2)) {
+      throw new Error('Sides 1, 2 and 3 can not form a triangle');
+    }
+  }
+
+  getArea(): number {
+    const halfPer = (this.side1 + this.side2 + this.side3) / 2;
```
dellachaise, line 35: This name is more obvious and meaningful.
```suggestion
    const halfPerimetr = (this.side1 + this.side2 + this.side3) / 2;
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,87 @@
...
+
+    const perimetr = halfPer * (halfPer - this.side1)
+    * (halfPer - this.side2) * (halfPer - this.side3);
 
-export class Circle {
+    const resultPer: number = Math.sqrt(perimetr);
 
+    return Number(resultPer.toFixed(2));
```
dellachaise, line 42: Optional: you round 3 times all over the code, better use for it one logic(for now you have 3 different ways). You also could create a separate helper function - `rounder` for example for this purpose.

dellachaise (changes requested):
Great job 👍 

src/figuresClasses.ts
anyakrainiuk: I did before without public

PR #587, opened 2022-09-01
Solution

Anastasiia-Khmelovska:


Sasha-Gordijchuk (approved):
Perfect Work!

src/figuresClasses.ts
```diff
@@ -1,19 +1,79 @@
-export interface Figure {
+type Shape = 'triangle' | 'circle' | 'rectangle';
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,79 @@
...
+  public shape: Shape = 'triangle';
 
-}
+  constructor(
+    public color: Color,
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    if (this.a <= 0 || this.b <= 0 || this.c <= 0
```
dellachaise, line 19: You don't need to use `this` inside of the constructor. The same all over the code.
```suggestion
    if (a <= 0 || b <= 0 || c <= 0
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,79 @@
...
+export class Circle implements Figure {
+  public shape: Shape = 'circle';
+
+  constructor(
+    public color: Color,
+    public radius: number,
+  ) {
+    if (this.radius <= 0) {
+      throw new Error('Radius must be the correct length');
+    }
```
dellachaise, line 47: Better create a helper function that will accept arguments (any length) and determine if some have `value <= 0`.
```
(...args) => args.some() ....
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,79 @@
...
+  ) {
+    if (this.width <= 0 || this.height <= 0) {
+      throw new Error('Width and height must be the correct length');
+    }
+  }
+
+  getArea(): number {
+    const area: number = this.width * this.height;
+
+    return Number(area.toFixed(2));
```
dellachaise, line 73: Optional: you round with the same logic twice over the code. You also could create a separate helper function - `rounder` for example for this purpose. And better use one logic to round all over the code, for now you do it 2 ways.

dellachaise (changes requested):
Great job 👍 

AndriiRut (approved):
great job! 🚀

PR #589, opened 2022-09-02
solution

Andrii-Kuzmenko:


src/figuresClasses.ts
```diff
@@ -1,19 +1,92 @@
+// import { type } from "os";
+
+type Shape = 'triangle' | 'circle' | 'rectangle';
```
alexandra-protyanova, line 3: it is better to use enum for shapes

src/figuresClasses.ts
```diff
@@ -1,19 +1,92 @@
...
+    public radius: number,
+    public shape: Shape = 'circle',
+  ) {
+    if (this.radius <= 0) {
+      throw new Error('Enter valid data');
+    }
+  }
 
+  getArea(): number {
+    return Math.floor((Math.PI * this.radius ** 2) * 100) / 100;
```
alexandra-protyanova, line 69: it would be great ti create separate function for rounding, because u r doing it 3 times

alexandra-protyanova (changes requested):

alexandra-protyanova (approved):
GJ!

PR #591, opened 2022-09-03
solution

vadym-zahazei:


src/figuresClasses.ts
```diff
@@ -1,19 +1,78 @@
+type Shape = 'triangle' | 'circle' | 'rectangle';
```
Desertiks, line 1: I believe that better use `enum` for more convenient use.

Desertiks (approved):
⚡Good work! ⚡

src/figuresClasses.ts
vadym-zahazei: Optimized with `enum` type ⚙

AlinaMatuschak (approved):

PR #592, opened 2022-09-03
ts_figures_classes

yurii-kozlov:


src/figuresClasses.ts
```diff
@@ -1,19 +1,78 @@
...
+  }
 
-export class Circle {
+  getArea(): number {
+    const p: number = (this.a + this.b + this.c) / 2;
+    const squareOfTriangle : number = Math.sqrt(
+      p * (p - this.a) * (p - this.b) * (p - this.c),
+    );
 
+    return Math.floor(squareOfTriangle * 100) / 100;
```
alexandra-protyanova, line 37: let us move the logic of rounding to separate function and use it everywhere

alexandra-protyanova (changes requested):

alexandra-protyanova (approved):
Well done!

PR #595, opened 2022-09-03
solution

maryna-mazur:


src/figuresClasses.ts
```diff
@@ -1,19 +1,69 @@
...
-
+  shape: 'triangle' | 'circle' | 'rectangle';
+  color: 'red' | 'green' | 'blue';
+  getArea(): number;
 }
 
-export class Triangle {
+export class Triangle implements Figure {
+  constructor(
+    public color: 'red' | 'green' | 'blue',
```
qoa11a, line 9: You can create type for possible colors to not repeat it everywhere.

src/figuresClasses.ts
```diff
@@ -1,19 +1,69 @@
...
+    public a: number,
```
qoa11a, line 10: You could solve task without making public fields, as you should pass instance of class into getInfo() function, not the entire class.

qoa11a (changes requested):

PR #596, opened 2022-09-03
"Figures" classes were implemented

Anton-Liada:


src/figuresClasses.ts
```diff
@@ -1,19 +1,96 @@
...
+    public radius: number,
+  ) {
+    super(Shape.Circle, color);
+    checkValue(this.radius);
+  }
 
+  getArea():number {
+    const area = Math.floor((Math.PI * this.radius ** 2) * 100) / 100;
+
+    return area;
```
tiserett, line 75:
```suggestion
   return Math.floor((Math.PI * this.radius ** 2) * 100) / 100;
```

tiserett (changes requested):
you can try to get rid of BaseFigure

tiserett (approved):
good job ;)

src/figuresClasses.ts
```diff
@@ -1,19 +1,84 @@
...
+
+  constructor(
+    public color: Color,
+    public radius: number,
+  ) {
+    checkValue(this.radius);
+  }
 
+  getArea():number {
+    return Math.floor((Math.PI * this.radius ** 2) * 100) / 100;
```
AlinaMatuschak, line 62: you round down to hundredths on each getArea, maybe better create a separate method for it?

AlinaMatuschak (approved):

PR #599, opened 2022-09-04
figures_classes solution

SergeyPistelev:


src/figuresClasses.ts
```diff
@@ -1,19 +1,90 @@
...
 }
 
 export class Rectangle {
+  shape: Shape = 'rectangle';
 
-}
+  constructor(
+    public color: Color,
+    public width: number,
+    public high: number,
```
Ohiiko, line 76: I think this is not the best name, it's better to write the "height"

Ohiiko (approved):

src/figuresClasses.ts
```diff
@@ -1,19 +1,90 @@
+type Color = 'red' | 'green' | 'blue';
+type Shape = 'triangle' | 'circle' | 'rectangle';
+
 export interface Figure {
+  shape: Shape;
+  color: Color;
+  getArea(): number;
+}
+
+function badTriangle(
```
AlinaMatuschak, line 10:
```suggestion
function checkAvailableTriangle(
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,90 @@
...
+    public color: Color,
+    public radius: number,
+  ) {
+    if (errorChecker(radius)) {
+      throw new Error('Radius must be greater than zero');
+    }
+  }
 
+  getArea(): number {
+    return Math.floor(Math.PI * (this.radius ** 2) * 100) / 100;
```
AlinaMatuschak, line 66: you round down to hundredths on each getArea, maybe better create a separate method for it?

AlinaMatuschak (approved):

PR #600, opened 2022-09-04
Solution

Ohiiko:


src/figuresClasses.ts
```diff
@@ -1,19 +1,79 @@
+type Shape ='triangle' | 'circle' | 'rectangle';
```
anyakrainiuk, line 1: Better use enum instead of both types

src/figuresClasses.ts
```diff
@@ -1,19 +1,79 @@
...
+function validates(...args: number[]): boolean {
+  return args.some((param : number) => param <= 0);
 }
 
 export class Triangle {
+  shape: Shape = 'triangle';
+
+  constructor(
+    public color: Color,
+    public a: number,
```
anyakrainiuk, line 19: Better use meaningful names for variables)

src/figuresClasses.ts
```diff
@@ -1,19 +1,79 @@
...
+    public b: number,
+    public c: number,
+  ) {
+    if (validates(this.a, this.b, this.c)) {
```
anyakrainiuk, line 23: You don't need to write 'this' inside of the constructor, it's redundant. The same in other constructors.

anyakrainiuk (changes requested):

anyakrainiuk (approved):
Good Job🚀

danylo-onishchuk (approved):

PR #602, opened 2022-09-04
solution

adrqq:


src/figuresClasses.ts
```diff
@@ -1,19 +1,72 @@
...
+  shape: Shape = Shape.Triangle;
 
-}
+  constructor(
+    public color: 'red' | 'green' | 'blue',
+    public a:number,
+    public b:number,
+    public c:number,
+  ) {
+    if (a >= b + c || b >= a + c || c >= a + b || a <= 0 || b <= 0 || c <= 0) {
```
Andrii-Dunets, line 22: may be better to rewrite each condition on a new line

Andrii-Dunets (approved):
Good job)

src/figuresClasses.ts
AndriiRut: and create a separate function for checking is the side's size is below zero. And use this function in other classes.

AndriiRut (changes requested):

src/figuresClasses.ts
```diff
@@ -1,19 +1,82 @@
...
-export class Triangle {
+enum Shape {
+  Triangle = 'triangle',
+  Circle = 'circle',
+  Rectangle = 'rectangle',
+}
 
+function belowZero(property:number): void {
+  if (property <= 0) {
+    throw new Error('Error: radius must be greater than 0!!!');
```
AndriiRut, line 15: you've used this function to the Triangle and if one of the sides of the triangle will be 0 or above user will see a message about radius? something wrong here

AndriiRut (changes requested):

src/figuresClasses.ts
```diff
@@ -10,12 +10,31 @@ enum Shape {
   Rectangle = 'rectangle',
 }
 
-function belowZero(property:number): void {
-  if (property <= 0) {
+function circleDataChecker(radius:number): void {
+  if (radius <= 0) {
     throw new Error('Error: radius must be greater than 0!!!');
```
denys-danyliuk, line 15: you definitely should fix it according to the comment above

denys-danyliuk (changes requested):
all good, just specify an error you throw

denys-danyliuk (approved):
my bad, you already done it, good job🔥

PR #603, opened 2022-09-04
Figure classes

Alej4ndro1:


src/figuresClasses.ts
```diff
@@ -1,19 +1,95 @@
...
+      a >= (b + c)
+      || b >= (a + c)
+      || c >= (a + b)
+    ) {
+      throw new Error('The longest side is longer than sum of two others');
+    }
+  }
+
+  getArea(): number {
+    const s = (this.a + this.b + this.c) / 2;
```
ostUp, line 44: It could be for example `semiPeriment` instead of s.

src/figuresClasses.ts
```diff
@@ -1,19 +1,95 @@
...
+    public color: Color,
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    if (
+      a <= 0
+      || b <= 0
+      || c <= 0
+    ) {
```
ostUp, line 32: Better create a helper function that will accept arguments (any length) and determine if some have value <= 0.

`(...args) => args.some() ....`

src/figuresClasses.ts
```diff
@@ -1,19 +1,95 @@
...
+  }
+
+  getArea(): number {
+    const s = (this.a + this.b + this.c) / 2;
+
+    const areaOfTriangle = s * (s - this.a) * (s - this.b) * (s - this.c);
 
+    const result = Math.sqrt(areaOfTriangle);
+
+    return Math.floor(result * 100) / 100;
```
ostUp, line 50: Optional: you round with the same logic twice over the code. You also could create a separate helper function - `rounder` for example for this purpose.

ostUp (changes requested):

ostUp (approved):

danylo-onishchuk (approved):

PR #610, opened 2022-09-08
Solution

G0odvin:


src/figuresClasses.ts
```diff
@@ -1,19 +1,86 @@
...
+  ) {
+    if (this.radius <= 0) {
+      throw new Error('Invalid parameters');
+    }
+  }
+
+  getArea(): number {
+    const areaCircle = Math.PI * this.radius ** 2;
+
+    return Math.floor(areaCircle * 100) / 100;
```
alexandra-protyanova, line 60: it would be great to create separate function for rounding and to call where needed and maybe for throwing errors

alexandra-protyanova (approved):
Well done!

PR #611, opened 2022-09-11
solution

frontend112:


src/figuresClasses.ts
```diff
@@ -1,19 +1,99 @@
...
+      throw new Error('Each side should be bigger than 0');
+    }
+
+    if (a + b <= c || a + c <= b || c + b <= a) {
+      throw new Error(`the longest side of a triangle 
+      is >= than a sum of two others`);
+    }
+  }
+
+  getArea():number {
```
amitpatiljc, line 38: we can reduce number of `this` by destructuring

amitpatiljc (changes requested):

amitpatiljc (approved):
Good Job!

PR #613, opened 2022-09-14
solution

HusakOleh:


Reptiloid044 (approved):
Nice job!

Reptiloid044 (approved):

src/figuresClasses.ts
```diff
@@ -1,19 +1,94 @@
-export interface Figure {
-
+function errorParams(...args: number[]): boolean {
+  return args.some((el: number): boolean => {
+    return el <= 0;
+  });
 }
 
-export class Triangle {
+type Form = 'triangle' | 'circle' | 'rectangle';
```
dellachaise, line 7: Better use `enum` here - it will give you more control and flexibility for `Shape` type.

src/figuresClasses.ts
```diff
@@ -1,19 +1,94 @@
-export interface Figure {
-
+function errorParams(...args: number[]): boolean {
+  return args.some((el: number): boolean => {
```
dellachaise, line 2: `el` or `res` is not a good name for the variable. Better name `value` for example.

src/figuresClasses.ts
```diff
@@ -1,19 +1,94 @@
...
+  }
 
+  getArea(): number {
+    const semiPerimeter = (this.a + this.b + this.c) / 2;
+    const area = Math.sqrt(semiPerimeter
+      * (semiPerimeter - this.a)
+      * (semiPerimeter - this.b)
+      * (semiPerimeter - this.c));
+
+    return Math.floor(area * 100) / 100;
```
dellachaise, line 49: Optional: You do almost the same calculation 3 times. Better create helper function for this purpose.

dellachaise (approved):
Good job 👍

PR #615, opened 2022-09-16
Solution

Serhii-Vovk:


src/figuresClasses.ts
```diff
@@ -1,19 +1,82 @@
...
+  constructor(
+    public color: Color,
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    const sides = [this.a, this.b, this.c];
+    const sortedSide = sides.sort((x, y) => y - x);
+
+    if (sides.some((x) => x < 0)) {
```
SanyaBratashchuk, line 22:
```suggestion
    if (sides.some((x) => x <= 0)) {
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,82 @@
...
 
-}
+  constructor(
+    public color: Color,
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    const sides = [this.a, this.b, this.c];
+    const sortedSide = sides.sort((x, y) => y - x);
```
SanyaBratashchuk, line 20:
```suggestion
    const sortedSide = sides.sort((x, y) => y - x);
```
It is useless to create this variable if you don't make a copy of sides array because of sort mutate an array

src/figuresClasses.ts
```diff
@@ -1,19 +1,82 @@
...
+    public color: Color,
+    public radius: number,
+  ) {
+    if (this.radius <= 0) {
+      throw new Error('Radius must be > 0');
+    }
+  }
+
+  getArea(): number {
+    const areaCircle = Math.PI * (this.radius * this.radius);
```
SanyaBratashchuk, line 54:
```suggestion
    const areaCircle = Math.PI * (this.radius ** 2);
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,82 @@
...
+export class Rectangle implements Figure {
+  shape: Shape = 'rectangle';
+
+  constructor(
+    public color: Color,
+    public height: number,
+    public width: number,
+  ) {
+    if (this.height <= 0 || this.width <= 0) {
+      throw new Error('Height or width must be > 0');
```
SanyaBratashchuk, line 69:
```suggestion
      throw new Error('Height and width must be > 0');
```

SanyaBratashchuk (changes requested):

src/figuresClasses.ts
```diff
@@ -1,19 +1,83 @@
...
+  shape: Shape = 'triangle';
 
-}
+  constructor(
+    public color: Color,
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    const sides = [this.a, this.b, this.c];
```
Bohdan-Maksymiuk, line 19: IMHO - useless constant - just duplicate

Bohdan-Maksymiuk (approved):
Good job 👍

PR #616, opened 2022-09-18
Solution

Viktorianeimesh:


src/figuresClasses.ts
```diff
@@ -1,19 +1,74 @@
...
+    public color: Color,
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    if ((a >= (b + c)) || (b >= (a + c)) || (c >= (a + b))
+    || (a <= 0 || b <= 0 || c <= 0)) {
+      throw new Error('error: sides 1, 2 and 3 cannot form a triangle');
+    }
+  }
```
Bohdan-Maksymiuk, line 23: better to make 2 checks and throw 2 different errors

src/figuresClasses.ts
```diff
@@ -1,19 +1,74 @@
...
+    public color: Color,
+    public radius: number,
+  ) {
+    if (radius <= 0) {
+      throw new Error('error: radius of the circle is less than 0');
+    }
+  }
+
+  getArea():number {
+    const result = (Math.PI * this.radius * this.radius);
```
Bohdan-Maksymiuk, line 48:
```suggestion
    const result = Math.PI * this.radius * this.radius;
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,74 @@
...
+  shape: Shape = 'rectangle';
 
-}
+  constructor(
+    public color: Color,
+    public width: number,
+    public height: number,
+  ) {
+    if ((width <= 0) || (height <= 0)) {
+      throw new Error('error: width and height cannot form a rectangle');
```
Bohdan-Maksymiuk, line 63: why? you can say that negative can’t and pass width and height into error message

src/figuresClasses.ts
```diff
@@ -1,19 +1,74 @@
...
+    }
+  }
 
-export function getInfo(figure) {
+  getArea():number {
+    return +(this.width * this.height).toFixed(2);
```
Bohdan-Maksymiuk, line 68: IMHO it is more readable to use `Number()`

Bohdan-Maksymiuk (changes requested):
Good job, fix comments ⏬

src/figuresClasses.ts
Viktorianeimesh: Thanks, replaced to Number()

src/figuresClasses.ts
Viktorianeimesh: Thanks, I removed brackets

src/figuresClasses.ts
Viktorianeimesh: Thanks, I changed to 2 checks

src/figuresClasses.ts
Viktorianeimesh: Sorry, I don't understand how to implement it, please help

Bohdan-Maksymiuk: ask in fe_chat instead of waiting next review)
Through interpolation and show user the values of width and height

Bohdan-Maksymiuk (approved):
Good job 👍 
Leaved last comment on your own 😉

PR #617, opened 2022-09-19
solution

ElinaBahirova:


src/figuresClasses.ts
```diff
@@ -1,19 +1,107 @@
...
+    if (b >= a + c) {
+      return false;
+    }
+  }
 
+  if (c > a && c > b) {
+    if (c >= a + b) {
+      return false;
+    }
+  }
```
liliia-livandovskaya, line 24: let's rewrite it without using nested if

liliia-livandovskaya (changes requested):
Great job! Lets work on fix

src/figuresClasses.ts
```diff
@@ -1,19 +1,101 @@
...
+    return false;
+  }
+
+  if (b > a && b > c && b >= a + c) {
+    return false;
+  }
 
+  if (c > a && c > b && c >= a + b) {
+    return false;
+  }
```
Bohdan-Maksymiuk, line 18: better use separate variable for each condition and then
if `var1 || var2 || var3` return false

Bohdan-Maksymiuk: or you can make array with sides, sort it, and use only one check

Bohdan-Maksymiuk (changes requested):

src/figuresClasses.ts
```diff
@@ -1,19 +1,97 @@
...
+  let sides: number[] = [a, b, c];
+
+  sides = sides.sort((x:number, y: number) => x - y);
 
+  if (sides[2] >= sides[1] + sides[0]) {
+    return false;
+  }
+
+  return true;
 }
```
Bohdan-Maksymiuk, line 17: 🔥

Bohdan-Maksymiuk (approved):

PR #618, opened 2022-09-19
Done

Oleksii-Blyzniuk:


src/figuresClasses.ts
```diff
@@ -1,19 +1,91 @@
...
+    public b: number,
+    public c: number,
+  ) {
+    this.color = color;
+    this.a = a;
+    this.b = b;
+    this.c = c;
 
-export class Circle {
+    if (a === 0 || b === 0 || c === 0) {
```
TarasHoliuk, line 25: check the case if the side is lower than 0: `a <= 0`

src/figuresClasses.ts
```diff
@@ -1,19 +1,91 @@
...
+  constructor(
+    public color: Color,
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    this.color = color;
+    this.a = a;
+    this.b = b;
+    this.c = c;
```
TarasHoliuk, line 23: you can remove it fi declare properties as you do in the constructor

TarasHoliuk (changes requested):
Almost done! Left two small comments

liliia-livandovskaya (approved):

PR #622, opened 2022-09-22
Solution

andriimelnyq:


src/figuresClasses.ts
```diff
@@ -1,19 +1,86 @@
...
+  }
 }
 
-export class Triangle {
+function roundArea(area: number): number {
+  return Math.floor(area * 100) / 100;
+}
 
+export interface Figure {
+  shape: string;
```
ivanchuk-yurii, line 24: move `'triangle' | 'circle' | 'rectangle'` to the type and use here

src/figuresClasses.ts
```diff
@@ -1,19 +1,86 @@
...
 
-export class Rectangle {
+export class Circle implements Figure {
+  shape = 'circle';
+
+  constructor(
+    public color: Color,
+    public radius: number,
+  ) {
+    catchError(radius);
```
ivanchuk-yurii, line 56: what is the reason of doing this for `Circle` and `Rectangle`?

ivanchuk-yurii (changes requested):

SanyaBratashchuk (changes requested):

ivanchuk-yurii (approved):

PR #630, opened 2022-09-29
Solution

MaxManis:


src/figuresClasses.ts
```diff
@@ -1,19 +1,120 @@
...
+  }
+
+  checkTriangle(): void {
+    const sides = [this.sizeA, this.sizeB, this.sizeC];
+    const aver = sides.reduce((a: number, b: number) => a + b, 0);
+
+    if (aver - Math.max(...sides) <= Math.max(...sides)
+      || sides.find((a: number) => (a <= 0))) {
+      throw new Error('Invalid shape settings');
+    }
```
silchyn, line 50: Good point)

silchyn: But it is not 'aver'

silchyn (changes requested):

silchyn (approved):
Cool!

PR #632, opened 2022-10-01
Solution

AntonVaida:


src/figuresClasses.ts
```diff
@@ -1,19 +1,112 @@
...
-export class Triangle {
+export class Triangle implements Figure {
+  color: Color;
+
+  a:number;
+
+  b:number;
+
+  c:number;
+
```
Anna-Tkachenko, line 20: let's use the next syntax in constructor for creating new properties instead of repeating the same code 2 times (declaring props here, and added them in constructor): `public color: Color`
just add input values in such way, and you don't need to declare it and assign a value 2 times

src/figuresClasses.ts
```diff
@@ -1,19 +1,112 @@
...
+  }
 }
 
-export class Rectangle {
+export class Rectangle implements Figure {
+  color: Color;
+
+  width:number;
+
+  height:number;
```
Anna-Tkachenko, line 84: same here

Anna-Tkachenko (changes requested):

src/figuresClasses.ts
```diff
@@ -1,19 +1,102 @@
...
+
+  a:number;
+
+  b:number;
+
+  c:number;
+
+  shape:Shape = 'triangle';
+
+  constructor(color:Color, a:number, b:number, c:number) {
```
Anna-Tkachenko, line 23: you didn't fix the comment
rewrite receiving and setting props in constructor using private modifier
example: `public color: Color, `

Anna-Tkachenko (changes requested):

Denchepornyuk (approved):

PR #635, opened 2022-10-03
Develop

grafit-off:
feat: interface and classes were implemented

### DIFF
--- a/src/figuresClasses.ts
+++ b/src/figuresClasses.ts
@@ -1,19 +1,97 @@
+type Shape = 'triangle' | 'circle' | 'rectangle';
+type Color = 'red' | 'green' | 'blue';
+
 export interface Figure {
-
+  shape: Shape;
+  color: Color;
+  getArea(): number;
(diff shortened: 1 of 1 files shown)


AntonVaida (approved):
GJ!

src/figuresClasses.ts
```diff
@@ -1,19 +1,97 @@
...
-export class Triangle {
+export class Triangle implements Figure {
+  shape: Shape = 'triangle';
 
-}
+  constructor(
+    public color: Color,
+    public a: number,
+    public b: number,
+    public c: number,
```
silchyn, line 18: Is it really has to be public?

silchyn (changes requested):

src/figuresClasses.ts
grafit-off: @silchyn  We don't have requirements about this, so why not

silchyn (approved):
Cool!

PR #637, opened 2022-10-03
Figures

Vladosik1337:


src/figuresClasses.ts
```diff
@@ -1,19 +1,93 @@
...
+  }
 
-export class Circle {
+  getArea(): number {
+    const perimeter = (this.a + this.b + this.c) / 2;
 
+    return Math.round(Math.sqrt(perimeter
+      * (perimeter - this.a)
+      * (perimeter - this.b)
+      * (perimeter - this.c)) * 100) / 100;
```
ArturNagaycev, line 29: mathematician would say that it's not a perimeter, this is halfOfPerimeter))))

a-nahaitsev (approved):
Clear solution! Good job!

src/figuresClasses.ts
```diff
@@ -1,19 +1,93 @@
...
+export interface Figure {
+  shape: Shape;
+  color: Color;
+  getArea(): number;
+  checkSides(): void;
 }
 
-export class Triangle {
+export class Triangle implements Figure {
+  shape: Shape = 'triangle';
```
OleksiiBlezniuk, line 12: Let's use enum for that purpose

src/figuresClasses.ts
OleksiiBlezniuk, line 29:
```suggestion
    return Math.round(
      Math.sqrt(perimeter
        * (perimeter - this.a)
        * (perimeter - this.b)
        * (perimeter - this.c)
      ) * 100
    ) / 100;
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,93 @@
...
+    ) {
+      throw new Error('The longest side of a triangle'
+      + 'is >= than a sum of two others');
+    }
+  }
 }
 
-export class Rectangle {
+export class Circle implements Figure {
+  shape: Shape = 'circle';
```
OleksiiBlezniuk, line 49: enum

src/figuresClasses.ts
```diff
@@ -1,19 +1,93 @@
-export interface Figure {
+type Shape = 'triangle' | 'circle' | 'rectangle';
+type Color = 'red' | 'green'| 'blue';
```
OleksiiBlezniuk, line 2: Better to create enums, using enum we'll avoid magic strings

OleksiiBlezniuk (changes requested):

src/figuresClasses.ts
```diff
@@ -1,19 +1,101 @@
-export interface Figure {
-
+enum Shape {
+  Triangle ='triangle',
+  Circle = 'circle',
+  Rectangle = 'rectangle',
 }
 
-export class Triangle {
+type Color = 'red' | 'green'| 'blue';
```
OleksiiBlezniuk, line 7: Also you could create enum for this values too

OleksiiBlezniuk (approved):

PR #638, opened 2022-10-03
solution

vladimir-dublya:


AniaHna (commented):
Good job!

src/figuresClasses.ts
```diff
@@ -1,19 +1,77 @@
...
+    public a: number,
+    public b: number,
+    public c: number,
+    public shape: Shape = Shape.triangle,
+  ) {
+    if (this.a + this.b <= this.c
+      || this.a + this.c <= this.b
+      || this.b + this.c <= this.a) {
+      throw new Error('Triangle is invalid');
+    }
```
naziamoff, line 27: It's better to make the conditions like that to a separate variable

src/figuresClasses.ts
```diff
@@ -1,19 +1,77 @@
...
 }
 
-export class Circle {
+export class Triangle implements Figure {
+  constructor(
+    public color: Color,
+    public a: number,
+    public b: number,
+    public c: number,
+    public shape: Shape = Shape.triangle,
```
naziamoff, line 21: the `shape` should not be in the constructor props

naziamoff: because it will always be a triangle, we never pass it explicitly

naziamoff (approved):
Well done, but I strongly recommend you to fix comments

PR #640, opened 2022-10-03
Figure Classes TS

Anna-Borodaienko:


src/figuresClasses.ts
```diff
@@ -1,19 +1,81 @@
-export interface Figure {
+type Shape = 'triangle' | 'circle' | 'rectangle';
 
+type Color = 'red' | 'green' | 'blue';
+export interface Figure {
+  color: Color;
+  shape?: Shape;
```
TheVovchik, line 6: we have shape in every figure, I think it's neccesary parametr

src/figuresClasses.ts
```diff
@@ -1,19 +1,81 @@
...
+  getArea(): number;
 }
 
-export class Triangle {
+export class Triangle implements Figure {
+  constructor(
+    public color: Color,
+    public sideA: number,
+    public sideB: number,
+    public sideC: number,
```
TheVovchik, line 15: we not use sides in public space, better use private instead

src/figuresClasses.ts
```diff
@@ -1,19 +1,81 @@
...
+    public shape?: Shape,
+  ) {
+    this.shape = 'triangle';
```
TheVovchik, line 18: we can use this implementetion in class body, because all triangles will have this propery

src/figuresClasses.ts
```diff
@@ -1,19 +1,81 @@
...
 
-}
+    if (sideA <= 0 || sideA <= 0 || sideC <= 0) {
+      throw new Error('Incorrect values for triangle sides');
```
TheVovchik, line 21: we haven't this in task but it's good decision 🔥

src/figuresClasses.ts
```diff
@@ -1,19 +1,81 @@
...
+    const maxSide: number = Math.max(sideA, Math.max(sideB, sideC));
+
+    if (maxSide >= (sideA + sideB + sideC - maxSide)) {
+      throw new Error('It is not a triangle');
+    }
+  }
 
-export class Circle {
+  getArea(): number {
+    const sunSides: number = this.sideA + this.sideB + this.sideC;
```
TheVovchik, line 32: what is sunSides?) totalSidesLength instead?

src/figuresClasses.ts
```diff
@@ -1,19 +1,81 @@
...
+      * (semiPerimeter - this.sideB)
+      * (semiPerimeter - this.sideC)) * 100) / 100;
+  }
 }
 
-export class Rectangle {
+export class Circle implements Figure {
+  constructor(
+    public color: Color,
+    public radius: number,
```
TheVovchik, line 45: also looks like private value, that don't used in public space

src/figuresClasses.ts
```diff
@@ -1,19 +1,81 @@
...
+    public shape?: Shape,
+  ) {
+    this.shape = 'circle';
```
TheVovchik, line 48: we can use this implementetion in class body, because all circles will have this propery

src/figuresClasses.ts
```diff
@@ -1,19 +1,81 @@
...
+    return Math.floor((Math.PI * this.radius ** 2) * 100) / 100;
+  }
 }
 
-export function getInfo(figure) {
+export class Rectangle implements Figure {
+  constructor(
+    public color: Color,
+    public width: number,
+    public height: number,
```
TheVovchik, line 64: same as radius and sizes

src/figuresClasses.ts
```diff
@@ -1,19 +1,81 @@
...
+    public shape?: Shape,
+  ) {
+    this.shape = 'rectangle';
```
TheVovchik, line 67: we can use this implementetion in class body, because all rectangles will have this propery

src/figuresClasses.ts
```diff
@@ -1,19 +1,81 @@
...
+
+    if (width <= 0 || height <= 0) {
+      throw new Error('Incorrect values for rectangle sides');
+    }
+  }
+
+  getArea(): number {
+    return this.width * this.height;
```
TheVovchik, line 75: 4.12 * 5.15 what result we will have here? 
not rounded down to hundredths

src/figuresClasses.ts
```diff
@@ -1,19 +1,81 @@
...
+  }
+}
 
+export function getInfo(figure: Figure): string {
+  return `A ${figure.color} ${figure.shape} - ${figure.getArea()}`;
```
TheVovchik, line 80: great practice 👍

TheVovchik (changes requested):
good work, leave some comments belowe

src/figuresClasses.ts
Anna-Borodaienko: Agree, fixed

src/figuresClasses.ts
Anna-Borodaienko: There are not clear requirements for these variables )

src/figuresClasses.ts
Anna-Borodaienko: fixed

src/figuresClasses.ts
Anna-Borodaienko: done

src/figuresClasses.ts
Anna-Borodaienko: done

src/figuresClasses.ts
Anna-Borodaienko: fixed

src/figuresClasses.ts
```diff
@@ -1,19 +1,78 @@
...
-export class Triangle {
+export class Triangle implements Figure {
+  shape: Shape = 'triangle';
 
-}
+  constructor(
+    public color: Color,
+    public sideA: number,
+    public sideB: number,
+    public sideC: number,
```
silchyn, line 17: It is bad practice to make fields public when unnecessary because it violates the encapsulation OOP principle. Even accessing shape and color properties from outside the class is not so good, would be nice to have a method for printing such information, but task requires that

silchyn (changes requested):

src/figuresClasses.ts
```diff
@@ -1,19 +1,78 @@
...
+    public color: Color,
+    private sideA: number,
+    private sideB: number,
+    private sideC: number,
+  ) {
+    if (sideA <= 0 || sideA <= 0 || sideC <= 0) {
+      throw new Error('Incorrect values for triangle sides');
+    }
+
+    const maxSide: number = Math.max(sideA, Math.max(sideB, sideC));
```
silchyn, line 23:
```suggestion
    const maxSide: number = Math.max(sideA, sideB, sideC);
```

silchyn (changes requested):

src/figuresClasses.ts
Anna-Borodaienko: I forgot that this method works with any quantity of numbers (

PR #641, opened 2022-10-04
classes

TheVovchik:


src/figuresClasses.ts
```diff
@@ -1,19 +1,102 @@
...
+    case 'circle': {
+      return `A ${figure.color} ${figure.shape} - ${figure.getArea()}`;
+    }
+
+    case 'rectangle': {
+      return `A ${figure.color} ${figure.shape} - ${figure.getArea()}`;
+    }
+
+    case 'triangle': {
+      return `A ${figure.color} ${figure.shape} - ${figure.getArea()}`;
```
Anna-Borodaienko, line 96: Do you really need to repeat the same code 3 times?

src/figuresClasses.ts
```diff
@@ -1,19 +1,102 @@
...
+    return Number(Math.sqrt(semiPer * A * B * C).toFixed(2));
+  }
+
+  checkProperties(): void {
+    const maxSide = Math.max(this.sideA, this.sideB, this.sideC);
+    const twoOtherSides = this.sideA + this.sideB + this.sideC - maxSide;
+
+    if (maxSide >= twoOtherSides) {
+      throw new Error('sides can\'t form a triangle');
+    }
```
Anna-Borodaienko, line 38: Don't you check if all sides are more than 0?

src/figuresClasses.ts
TheVovchik: you right! 👍

src/figuresClasses.ts
TheVovchik: also agree that need to chech sides

src/figuresClasses.test.ts
```diff
@@ -116,7 +116,7 @@ describe('Method "getArea" of class "Circle"', () => {
     const circle = new Circle('red', 6);
 
     expect(circle.getArea())
-      .toBe(113.09);
+      .toBe(113.10);
   });
```
OleksiiBlezniuk, line 120: Well, nice try 😅
have you tried other ways to round number?

src/figuresClasses.ts
```diff
@@ -1,19 +1,90 @@
+type Shape = 'triangle' | 'circle' | 'rectangle';
```
OleksiiBlezniuk, line 1: Let's use enum for that purpose, using enum we'll avoid magic strings

src/figuresClasses.ts
```diff
@@ -1,19 +1,90 @@
...
+  }
+
+  getArea(): number {
+    const semiPer = (this.sideA + this.sideB + this.sideC) / 2;
+    const A = semiPer - this.sideA;
+    const B = semiPer - this.sideB;
+    const C = semiPer - this.sideC;
 
-export class Circle {
+    return Number(Math.sqrt(semiPer * A * B * C).toFixed(2));
```
OleksiiBlezniuk, line 29: Probably it's not the way of rounding that tests expected 🙂

OleksiiBlezniuk (changes requested):

src/figuresClasses.ts
TheVovchik: "Я, кстати, энум не использовал ни разу)" - comment of middle in Epum)))

src/figuresClasses.test.ts
TheVovchik: totally not agry! because Before round it's square is 113.097 which is equal 113.1!
"returns the area of the figure rounded down to hundredths"

It's a mistake in test

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/toFixed

src/figuresClasses.ts
TheVovchik: Task expect that "returns the area of the figure rounded down to hundredths". 
I use method instead of Math.trunc(Math.sqrt(semiPer * A * B * C) * 100) / 100 because this one is not round
"returns the area of the figure to **second number after point**".

src/figuresClasses.test.ts
OleksiiBlezniuk: Okay, good point but there is a purpose in that `down` word of this sentence 🙂 

Here is a good sample of differences, check this link

https://www.quora.com/Whats-the-difference-between-round-up-round-down-and-round-off

src/figuresClasses.ts
OleksiiBlezniuk: Check first comment please

src/figuresClasses.ts
OleksiiBlezniuk: Well, there are so many question for that middle guy =) I don't know how devs in Epam using this tool, but in our company (and AFAIK, not only in our) this tool used much often 🙂

src/figuresClasses.test.ts
OleksiiBlezniuk: Still, if you not agree with this suggestion, you may ask opinion from other mentors/coaches 🙂  I can be wrong

TheVovchik: Lets open wikipedia https://ru.wikipedia.org/wiki/%D0%9E%D0%BA%D1%80%D1%83%D0%B3%D0%BB%D0%B5%D0%BD%D0%B8%D0%B5
May be this task was created when method toFix wasn't implemented, so we have a tool to calculate properly.

src/figuresClasses.ts
TheVovchik: reworked it on enum

src/figuresClasses.ts
TheVovchik: **rounded down** crap!!!!!

src/figuresClasses.test.ts
TheVovchik: **rounded down**.

sorry mate, I was sure that this one is general round

OleksiiBlezniuk (approved):
👍

PR #644, opened 2022-10-04
Solution

jojnajbude:


xiimiik (commented):
GJ 🔥

src/figuresClasses.ts
```diff
@@ -1,19 +1,103 @@
...
 }
 
-export class Triangle {
+export class Triangle implements Figure {
+  public sideA: number;
 
-}
+  public sideB: number;
+
+  public sideC: number;
```
silchyn, line 16: It is bad practice to make fields public when unnecessary, it violates the encapsulation OOP principle

silchyn (changes requested):

src/figuresClasses.ts
```diff
@@ -1,19 +1,103 @@
...
-export class Triangle {
+export class Triangle implements Figure {
+  private sideA: number;
 
-}
+  private sideB: number;
+
+  private sideC: number;
+
+  public shape: Shape = 'triangle';
```
naziamoff, line 18:
```suggestion
  private sideB: number;
  private sideC: number;
  public shape: Shape = 'triangle';
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,103 @@
...
+    b: number,
+    c: number,
+  ) {
+    if (a >= b + c || b >= a + c || c >= a + b) {
+      throw new Error(`sides {${a}, ${b} and ${c} can't form a triangle}`);
+    }
+
+    if (!a || !b || !c) {
+      throw new Error('sides cannot be zero');
+    }
```
naziamoff, line 32: what if we pass some negative value there?

src/figuresClasses.ts
```diff
@@ -1,19 +1,103 @@
...
+
+    this.sideA = a;
+    this.sideB = b;
+    this.sideC = c;
+  }
+
+  getArea(): number {
+    const halfP: number = (this.sideA + this.sideB + this.sideC) / 2;
```
naziamoff, line 40: use full names pls, now it is unclear what is stored inside the variable

naziamoff (changes requested):

src/figuresClasses.ts
jojnajbude: linter says these does not cool

naziamoff (approved):
GJ!

PR #646, opened 2022-10-04
FiguresClasses was declared

Yevheniya-Deryka:
Used Math.trunc instead of Math.round in getArea() for Circle because of incorrect rounding in the test

### DIFF
--- a/src/figuresClasses.ts
+++ b/src/figuresClasses.ts
@@ -1,19 +1,92 @@
+type Color = 'red' | 'green' | 'blue';
+type Shape = 'triangle' | 'circle' | 'rectangle';
+
 export interface Figure {
-
+  shape: Shape;
+  color: Color;
+  getArea(): number;
+  checkInitialData(): void;
 }
 
-export class Triangle {
+export class Triangle implements Figure {
+  shape: Shape = 'triangle';
 
+  constructor(
+    public color: Color,
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    this.checkInitialData();
+  }
+
+  checkInitialData(): void {
+    if (this.a <= 0 || this.b <= 0 || this.c <= 0) {
+      throw new Error('All sides should be > than zero!');
+    }
+
+    const sides: number[] = [this.a, this.b, this.c]
+      .sort((n: number, m: number) => m - n);
+
+    if (sides[0] >= sides[1] + sides[2]) {
+      throw new Error(
+        'The longes side should be < then the sum of two others!',
+      );
+    }
+  }
+
+  getArea(): number {
+    const s = (this.a + this.b + this.c) / 2;
+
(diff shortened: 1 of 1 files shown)


src/figuresClasses.ts
```diff
@@ -1,19 +1,92 @@
...
+    if (sides[0] >= sides[1] + sides[2]) {
+      throw new Error(
+        'The longes side should be < then the sum of two others!',
+      );
+    }
+  }
 
-export class Circle {
+  getArea(): number {
+    const s = (this.a + this.b + this.c) / 2;
```
Apollonia5, line 39: you should change naming for sides and semiperimeter

src/figuresClasses.ts
```diff
@@ -1,19 +1,92 @@
...
+  shape: Shape = 'circle';
 
+  constructor(
+    public color: Color,
+    public radius: number,
+  ) {
+    this.checkInitialData();
+  }
+
+  checkInitialData(): void {
```
Apollonia5, line 57: you can check the parameters inside the constructor without creating a new method

src/figuresClasses.ts
```diff
@@ -1,19 +1,92 @@
...
+    if (this.width <= 0 || this.height <= 0) {
+      throw new Error('Both sides should be > than zero!');
+    }
+  }
+
+  getArea(): number {
+    return this.width * this.height;
```
Apollonia5, line 86: we need to round the result to hundredths in case the parameters are with thousandths after the decimal point

Yevheniya-Deryka: +

src/figuresClasses.ts
Yevheniya-Deryka: for sure I can) but I wanted to make the constructor more compact)

src/figuresClasses.ts
Yevheniya-Deryka: we have such naming in the task description and as to me it's absolutely clear in this case what are a, b, c about.
but maybe you're right)

src/figuresClasses.ts
```diff
@@ -1,19 +1,93 @@
...
-export class Triangle {
+export class Triangle implements Figure {
+  shape: Shape = 'triangle';
 
-}
+  constructor(
+    public color: Color,
+    public sideA: number,
+    public sideB: number,
+    public sideC: number,
```
silchyn, line 18: It is bad practice to make fields public unnecessary. Check OOP principles, especially the encapsulation

silchyn (changes requested):

src/figuresClasses.ts
Yevheniya-Deryka: Done)

silchyn (approved):
Cool!

PR #649, opened 2022-10-04
Task was solved

DenysKleimenov:


src/figuresClasses.ts
```diff
@@ -1,19 +1,102 @@
...
+  }
 }
 
-export class Rectangle {
+export class Circle implements Figure {
+  shape: Shapes;
+
+  constructor(public color: Colors, public radius: number) {
+    this.shape = 'circle';
+    this.radius = radius;
```
Andrew-Skoryk, line 66: you already made it public

src/figuresClasses.ts
```diff
@@ -1,19 +1,102 @@
...
-}
+  a: number;
+
+  b: number;
+
+  c: number;
+
+  constructor(public color: Colors, a: number, b: number, c: number) {
+    this.shape = 'triangle';
+    this.color = color;
```
Andrew-Skoryk, line 21: you already made it public

Andrew-Skoryk (approved):
GJ!

src/figuresClasses.ts
```diff
@@ -1,19 +1,100 @@
...
 
-}
+  a: number;
+
+  b: number;
+
+  c: number;
+
+  constructor(public color: Colors, a: number, b: number, c: number) {
+    this.shape = 'triangle';
```
OleksiiBlezniuk, line 20: Use enum for this purpose

src/figuresClasses.ts
```diff
@@ -1,19 +1,100 @@
...
+      semiPerim
+      * (semiPerim - this.a) * (semiPerim - this.b) * (semiPerim - this.c),
+    );
+
+    return +area.toFixed(2);
+  }
 
-export class Circle {
+  isTriangle(): void {
+    if ((this.a === 0 || this.b === 0 || this.c === 0)) {
```
OleksiiBlezniuk, line 40:
```suggestion
    if (!this.a || !this.b || !this.c) {
```

OleksiiBlezniuk (changes requested):

src/figuresClasses.ts
OleksiiBlezniuk: But better to check if it's less or equal to 0 in that case to prevent negative values

src/figuresClasses.ts
```diff
@@ -1,19 +1,105 @@
...
+  shape: Shapes;
+
+  a: number;
+
+  b: number;
+
+  c: number;
+
+  constructor(public color: Colors, a: number, b: number, c: number) {
+    this.shape = Shapes.Triangle;
```
OleksiiBlezniuk, line 25: just set this field as public

OleksiiBlezniuk (changes requested):

Andr1yK (approved):

PR #650, opened 2022-10-04
Solution

Apollonia5:


src/figuresClasses.ts
```diff
@@ -1,19 +1,74 @@
...
+    public bSide: number,
+    public cSide: number,
+  ) {
+    if (aSide <= 0 || bSide <= 0 || cSide <= 0) {
+      throw new Error('The side of triangle is less than or equal to zero');
+    }
+
+    if (aSide >= bSide + cSide || bSide >= aSide + cSide
+      || cSide >= aSide + bSide) {
+      throw new Error('Euclid disapproves: you got a degenerate triangle');
```
Yevheniya-Deryka, line 25: )))

Yevheniya-Deryka (approved):
GJ! 
Cool error-messages))

src/figuresClasses.ts
```diff
@@ -1,19 +1,74 @@
...
-}
+  constructor(
+    public color: Color,
+    public aSide: number,
+    public bSide: number,
+    public cSide: number,
+  ) {
+    if (aSide <= 0 || bSide <= 0 || cSide <= 0) {
+      throw new Error('The side of triangle is less than or equal to zero');
+    }
```
Anna-Tkachenko, line 21: you have the same functionality in each constructor: you checked that all input valued should be greater or equal to 0
let's create separate function for this check, and will call it for each figure

Anna-Tkachenko (approved):

PR #651, opened 2022-10-04
solution was added

ghost:


Yeliseiev (approved):
Good work. ᕦ⁠(⁠ò⁠_⁠ó⁠ˇ⁠)⁠ᕤ

Anna-Tkachenko (changes requested):
Check checklist

src/figuresClasses.ts
```diff
@@ -1,19 +1,101 @@
...
+  correctValuesOfFigure(): void {
+    if (this.a <= 0 || this.b <= 0 || this.c <= 0) {
+      throw new Error('Resulting value should be even number');
+    }
+
+    let sides: number[] = [this.a, this.b, this.c];
+
+    sides = sides.sort((a: number, b: number) => b - a);
 
+    const longestSide: number = sides[0];
```
Denchepornyuk, line 36: don't add type for primitive js types
```suggestion
    const longestSide = sides[0];
```

Denchepornyuk (approved):

PR #652, opened 2022-10-04
Solution

Yeliseiev:


Maksym845 (approved):
Cool👍

src/figuresClasses.ts
```diff
@@ -1,19 +1,91 @@
...
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    const sides: number[] = [a, b, c];
+    const perimeter: number = a + b + c;
 
+    if (sides.some((side: number) => side <= 0)) {
+      throw new Error('The sides values must be greater than 0');
+    }
```
Anna-Tkachenko, line 35: you checked input values for each figure
will be better to create separate function for this purpose

Anna-Tkachenko (approved):

PR #656, opened 2022-10-04
solution

nikitaovch99:


vmamchur (approved):
Looks fine to me.

src/figuresClasses.ts
```diff
@@ -1,19 +1,73 @@
...
+
+export interface Figure {
+  shape: Shape,
+  color: Color,
+  getArea(): number
 }
 
-export class Triangle {
+export class Triangle implements Figure {
+  public shape: Shape = 'triangle';
```
OleksiiBlezniuk, line 12: Let's use enum for that purpose

OleksiiBlezniuk (changes requested):

src/figuresClasses.ts
```diff
@@ -1,19 +1,78 @@
...
+  public shape: Shape = Shape.Rectangle;
+
+  constructor(
+    public color: Color,
+    public a: number,
+    public b: number,
+  ) {
+    if (a <= 0 || b <= 0) {
+      throw new Error('Sides can`t be zero');
+    }
```
OleksiiBlezniuk, line 68: BTW, you've used logic that checks if any value is greater than 0 in every class, so you could move it to helper funciton

OleksiiBlezniuk (approved):

PR #658, opened 2022-10-04
Soulution

anastasiia-nahaiska:


src/figuresClasses.ts
```diff
@@ -1,19 +1,99 @@
...
+
+    if (a >= b + c || b >= a + c || c >= a + b) {
+      throw new Error('The side length cannot be greater'
+      + 'than or equal to sum of the others');
+    }
+
+    this.shape = Shape.Triangle;
+    this.a = a;
+    this.b = b;
+    this.c = c;
```
myroslav-diiak, line 41: You don`t need it when you added public or private in constructor arguments

src/figuresClasses.ts
```diff
@@ -1,19 +1,99 @@
...
+    public color: Color,
+    public radius: number,
+  ) {
+    if (radius <= 0) {
+      throw new Error('The radius cannot be less than or equal to 0');
+    }
+
+    this.shape = Shape.Circle;
+    this.color = color;
+    this.radius = radius;
```
myroslav-diiak, line 65: same

src/figuresClasses.ts
```diff
@@ -1,19 +1,99 @@
...
+    public b: number,
+  ) {
+    if (a <= 0 || b <= 0) {
+      throw new Error('The side length cannot'
+        + 'be less than or equal to 0');
+    }
+
+    this.shape = Shape.Rectangle;
+    this.a = a;
+    this.b = b;
```
myroslav-diiak, line 88: same

myroslav-diiak (approved):
GJ

src/figuresClasses.ts
OleksiiBlezniuk: agree, fix it everywhere please

src/figuresClasses.ts
```diff
@@ -1,19 +1,99 @@
...
+    this.color = color;
+  }
+
+  getArea(): number {
+    return +(this.a * this.b).toFixed(2);
```
OleksiiBlezniuk, line 93: Ok for tests, but this is not that rounding that expected by task, you need to round value down

OleksiiBlezniuk (changes requested):

src/figuresClasses.ts
```diff
@@ -1,19 +1,91 @@
...
+
+  constructor(
+    public color: Color,
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    if (a <= 0 || b <= 0 || b <= 0) {
+      throw new Error('The side length cannot be less than or equal to 0');
+    }
```
OleksiiBlezniuk, line 31: You've used same logic in every class, so would be nice to move it to helper function

OleksiiBlezniuk (approved):

PR #659, opened 2022-10-04
solution

artem-musii:


src/figuresClasses.ts
```diff
@@ -1,19 +1,80 @@
...
+  getArea(): number {
+    const s = Math.PI * (this.radius ** 2);
 
+    return roundArea(s);
+  }
 }
 
 export class Rectangle {
+  constructor(
+    public color: 'red' | 'green' | 'blue',
```
imaksweb, line 61: You have type `Color` why you don't use it as you do it in the `Triangle` class?

src/figuresClasses.ts
```diff
@@ -1,19 +1,80 @@
...
+    const p = (a + b + c) / 2;
+    const s = Math.sqrt(p * (p - a) * (p - b) * (p - c));
+
+    return roundArea(s);
+  }
 }
 
 export class Circle {
+  constructor(
+    public color: 'red' | 'green' | 'blue',
```
imaksweb, line 43: You have type `Color` why you don't use it as you do it in the `Triangle` class?

src/figuresClasses.ts
```diff
@@ -1,19 +1,80 @@
...
 
+    return roundArea(s);
+  }
 }
 
 export class Rectangle {
+  constructor(
+    public color: 'red' | 'green' | 'blue',
+    public a: number,
+    public b: number,
```
imaksweb, line 63: I think it could be better to use more descriptive names like `width` and `height`

imaksweb (approved):
Good job 👍

src/figuresClasses.ts
artem-musii: thanks, I didn't pay attention

OleksiiBlezniuk (changes requested):

OleksiiBlezniuk (approved):

src/figuresClasses.ts
```diff
@@ -1,19 +1,93 @@
...
 export class Circle {
+  public shape: Shape;
 
+  constructor(
+    public color: Color,
+    public radius: number,
+  ) {
+    this.shape = Shape.Circle;
+
+    if (radius < 0) {
```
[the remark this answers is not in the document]
artem-musii: corrected

PR #661, opened 2022-10-04
Solution

ArtemHulkoDev:


src/figuresClasses.ts
```diff
@@ -1,19 +1,73 @@
 export interface Figure {
-
+  shape: string;
+  color: string;
```
Andrew-Skoryk, line 3: Please be more specific, it's not a random string... an enum would be a great solution for this

Andrew-Skoryk (changes requested):
Great work! Add specific types for the Figure interface

src/figuresClasses.ts
ArtemHulkoDev: done)

Andrew-Skoryk (approved):

src/figuresClasses.ts
```diff
@@ -1,19 +1,76 @@
...
+export class Circle implements Figure {
+  shape: Shape = 'circle';
 
+  constructor(
+    public color: Color,
+    public radius: number,
+  ) {
+    if (this.radius <= 0) {
+      throw new Error('Radius is not a positive number');
+    }
```
Anna-Tkachenko, line 48: you check input value for figures sizes in each class
will be better to create separate function for this purpose and call this function in each class

Anna-Tkachenko (approved):

PR #663, opened 2022-10-04
Solution

Rina07:


src/figuresClasses.ts
```diff
@@ -1,19 +1,102 @@
...
+
 export interface Figure {
+  color: 'red' | 'green' | 'blue';
+  shape: 'triangle' | 'circle' | 'rectangle';
 
+  getArea: () => number;
 }
 
 export class Triangle {
+  shape: string = 'triangle';
```
korniienkokarina, line 11: you can omit the string type as you declare triangle type

src/figuresClasses.ts
```diff
@@ -1,19 +1,102 @@
...
+          * (this.semiperimeter - this.a)
+          * (this.semiperimeter - this.b)
+          * (this.semiperimeter - this.c),
+    );
+
+    if (this.a <= 0 || this.b <= 0 || this.c <= 0) {
+      throw new Error('Every side must be bigger than 0!');
+    }
+
+    if (!this.area) {
```
korniienkokarina, line 41: i guess it would be better to check if the longest side of a triangle is >= than a sum of two others literally like this:
`if (this.a >= this.b + this.c || this.b >= this.a + this.c
      || this.c >= this.a + this.b) {
      throw new Error('side is longer than the sum of two others');
    }` 

or you can add every side to an array and check the longest one

korniienkokarina (commented):
good job, left few comment to optimize the code 😊

src/figuresClasses.ts
iryna-borniak: have added enum for this line :)

src/figuresClasses.ts
iryna-borniak: thanks, have used this check)

korniienkokarina (approved):

src/figuresClasses.ts
```diff
@@ -1,19 +1,96 @@
-export interface Figure {
+// import { moveEmitHelpers } from 'typescript';
```
naziamoff, line 1:
```suggestion

```

src/figuresClasses.ts
```diff
@@ -1,19 +1,96 @@
...
+  return Math.floor(num * 100) / 100;
+};
+
+const checkParameters = (...parameters: number[]): void => {
+  if (parameters.some((num) => num <= 0)) {
+    throw new Error('Enter number bigger than 0!');
+  }
+};
+
 export class Triangle {
```
naziamoff, line 25: actually it would be better to explicitly mention what interface we implement here.
```suggestion
export class Triangle implements Figure {
```
It is used to verify that a class conforms to a specific interface

naziamoff (approved):
Well done 🔥

PR #664, opened 2022-10-04
solution

myroslav-diiak:


src/figuresClasses.ts
```diff
@@ -1,19 +1,80 @@
+type Shape = 'triangle' | 'circle' | 'rectangle';
```
anastasiia-nahaiska, line 1: As I  understand `enum` is better than literal types

anastasiia-nahaiska (approved):
GJ!)

silchyn (changes requested):

src/figuresClasses.ts
```diff
@@ -1,19 +1,77 @@
...
+      throw Error('All sides should be > 0');
+    }
+
+    if (a + b <= c || a + c <= b || c + b <= a) {
+      throw Error('The longes side should be < then the sum of two others!');
+    }
+  }
+
+  getArea(): number {
+    const perimetr: number = (this.a + this.b + this.c) / 2;
```
silchyn, line 27: It is not perimeter

src/figuresClasses.ts
```diff
@@ -1,19 +1,77 @@
...
+  }
 }
 
-export class Circle {
+export class Circle implements Figure {
+  shape: Shape = 'circle';
 
+  constructor(
+    public color: string,
+    private a: number,
```
silchyn, line 43: Shouldn't it be called 'radius'?

silchyn (changes requested):

naziamoff (approved):
🔥

PR #668, opened 2022-10-05
solution to figures classes

dsydorenko534:


iPanda177 (approved):
Nice work

src/figuresClasses.ts
```diff
@@ -1,19 +1,83 @@
...
+    const largest: number = Math.max(this.a, this.b, this.c);
 
-export class Circle {
+    if ((this.a + this.b + this.c - largest) <= largest) {
+      throw new Error('Sides 1, 2 and 3 can not form a triangle');
+    }
+  }
 
+  getArea(): number {
+    const semiPerimeter : number = (this.a + this.b + this.c) / 2;
```
Denchepornyuk, line 39: don't add type for primitive js types
```suggestion
    const semiPerimeter = (this.a + this.b + this.c) / 2;
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,83 @@
...
+    public radius: number,
+  ) {
+    if (invalidValue(this.radius)) {
+      throw new Error('Invalid value');
+    }
+  }
 
+  getArea(): number {
+    return Math.floor(Math.PI * (this.radius ** 2) * 100) / 100;
+  }
```
Denchepornyuk, line 60: the same for functions
```suggestion
  getArea() {
    return Math.floor(Math.PI * (this.radius ** 2) * 100) / 100;
  }
```

Denchepornyuk (approved):

PR #669, opened 2022-10-05
ts_figures_classes

xiimiik:


src/figuresClasses.ts
```diff
@@ -1,19 +1,83 @@
...
 
+export interface Figure {
+  shape: Shape;
+  color: Color;
+  getArea(): number
 }
 
-export class Triangle {
+export class Triangle implements Figure {
+  shape: Shape = 'triangle';
```
jojnajbude, line 11:
```suggestion
  public shape: Shape = 'triangle';
```

jojnajbude (approved):

src/figuresClasses.ts
OleksiiBlezniuk, line 11: Better to use enum instead of magic string

src/figuresClasses.ts
```diff
@@ -1,19 +1,83 @@
...
-}
+  constructor(
+    public color: Color,
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    if (this.a <= 0 || this.b <= 0 || this.c <= 0) {
+      throw new Error('Triangle sides should be more than 0');
+    }
```
OleksiiBlezniuk, line 21: Or you can create helper function that will check if values are valid 🙂

src/figuresClasses.ts
```diff
@@ -1,19 +1,83 @@
...
+    const semiPerimeter: number = (this.a + this.b + this.c) / 2;
 
-export class Circle {
+    const area = Math.sqrt(
+      semiPerimeter * (semiPerimeter - this.a)
+      * (semiPerimeter - this.b)
+      * (semiPerimeter - this.c),
+    );
 
+    return Math.round(area * 100) / 100;
```
OleksiiBlezniuk, line 40: Task expect other rounding here

OleksiiBlezniuk: Screenshot 2022-10-07 at 15 27 19

src/figuresClasses.ts
```diff
@@ -1,19 +1,83 @@
...
+    public width: number,
+    public height: number,
+  ) {
+    if (width <= 0 || height <= 0) {
+      throw new Error('Triangle radius should be more than 0');
+    }
+  }
+
+  getArea(): number {
+    return this.width * this.height;
```
OleksiiBlezniuk, line 77: And what about rounding this value? Even if tests are passed, this method still works incorrect 🙂

src/figuresClasses.ts
```diff
@@ -1,19 +1,83 @@
...
+  ) {
+    if (radius <= 0) {
+      throw new Error('Triangle radius should be more than 0');
+    }
+  }
+
+  getArea(): number {
+    const area = Math.PI * this.radius ** 2;
+
+    return Math.floor(area * 100) / 100;
```
OleksiiBlezniuk, line 59: You may move this function to helper function and reuse for all figures

OleksiiBlezniuk (changes requested):

src/figuresClasses.ts
```diff
@@ -1,19 +1,88 @@
...
+};
+
+export class Triangle implements Figure {
+  shape = Shape.Triangle;
+
+  constructor(
+    public color: Color,
+    public a: number,
+    public b: number,
+    public c: number,
```
silchyn, line 32: It is bad practice to make fields public when unnecessary. This is a part of the encapsulation OOP principle

silchyn (changes requested):

src/figuresClasses.ts
xiimiik: fixed

silchyn (changes requested):
'readonly' keyword does not make them private

silchyn (approved):
Cool!

PR #671, opened 2022-10-05
Solution

OleksandrGlulhovskiy:


grafit-off (approved):
Good job! 🔥

src/figuresClasses.ts
```diff
@@ -1,19 +1,96 @@
...
-export class Triangle {
+export class Triangle implements Figure {
+  shape: Shape = 'triangle';
 
-}
+  constructor(
+    public color: Color,
+    public a: number,
+    public b: number,
+    public c: number,
```
silchyn, line 18: It is bad practice to make fields public when unnecessary, it violates the encapsulation OOP principle

silchyn (changes requested):

PR #673, opened 2022-10-05
figureClasses

Arsen-Shpak:


src/figuresClasses.ts
```diff
@@ -1,19 +1,85 @@
...
+    private b: number,
+    private c: number,
+  ) {
+    if (
+      this.a + this.b <= this.c
+      || this.b + this.c <= this.a
+      || this.c + this.a <= this.b
+    ) {
+      throw new Error('This is not a triangle');
+    }
```
Dikey945, line 31: In my opinion, better to put this check to function and just call it in constructor

src/figuresClasses.ts
```diff
@@ -1,19 +1,85 @@
...
+  shape: Shape = Shape.Rectangle;
+
+  constructor(
+    public color: Color,
+    private width: number,
+    private height: number,
+  ) {
+    if (this.width <= 0 || this.height <= 0) {
+      throw new Error('this is not a Rectangle');
+    }
```
Dikey945, line 73: same here

src/figuresClasses.ts
```diff
@@ -1,19 +1,85 @@
...
 }
 
-export class Rectangle {
+export class Circle implements Figure {
+  shape: Shape = Shape.Circle;
+
+  constructor(public color: Color, private radius: number) {
+    if (this.radius <= 0) {
+      throw new Error('this is not a circle');
+    }
```
Dikey945, line 53: same here

Dikey945:
Good job 🔥

src/figuresClasses.ts
Anna-Tkachenko: agree, will be better to create separate function for this check

src/figuresClasses.ts
Anna-Tkachenko: you can create separate method, that will accept input arguments and will check that all of them is greater than 0
and you can use this function for checking input arguments for each constructor

Anna-Tkachenko (approved):

PR #677, opened 2022-10-05
solution of figure classes

Dikey945:


tatyana22 (approved):
GJ!

src/figuresClasses.ts
```diff
@@ -1,19 +1,125 @@
...
+  sideC: number;
 
-export class Circle {
+  constructor(
+    color: Color,
+    sideA: number,
+    sideB: number,
+    sideC: number,
+  ) {
+    this.shape = 'triangle';
```
OleksiiBlezniuk, line 27: Let's set this value as class field, outside constructor function, no need to set it here

src/figuresClasses.ts
```diff
@@ -1,19 +1,125 @@
...
+    this.color = color;
+    this.sideA = sideA;
+    this.sideB = sideB;
+    this.sideC = sideC;
```
OleksiiBlezniuk, line 31: Actually there is no need to write this values here 🙂  Just set this fields as public inside constructor arguments

src/figuresClasses.ts
```diff
@@ -1,19 +1,125 @@
...
+
+  getArea(): number {
+    const semiPerimetr: number
+      = (this.sideA + this.sideB + this.sideC) / 2;
+
+    const area: number
+      = Math.sqrt(semiPerimetr * (semiPerimetr - this.sideA)
+      * (semiPerimetr - this.sideB) * (semiPerimetr - this.sideC));
+
+    return Math.round(area * 100) / 100;
```
OleksiiBlezniuk, line 56: Task expects Math.floor here

src/figuresClasses.ts
OleksiiBlezniuk: Please fix everywhere

OleksiiBlezniuk (changes requested):

src/figuresClasses.ts
```diff
@@ -1,19 +1,96 @@
...
 
+export interface Figure {
+  shape: Shape;
+  color: Color
+  getArea(): number;
 }
 
-export class Triangle {
+export class Triangle implements Figure {
+  shape: Shape = 'triangle';
```
OleksiiBlezniuk, line 11: would be nice to use enum for that purpose

OleksiiBlezniuk (approved):

PR #679, opened 2022-10-05
Solution

tatyana22:


src/figuresClasses.ts
```diff
@@ -1,19 +1,73 @@
-export interface Figure {
+type Color = 'red'| 'green' | 'blue';
+
+enum Shape {
+  Triangle = 'triangle',
+  Circle = 'circle',
+  Rectangle = 'rectangle',
+}
```
Dikey945, line 7: I think we can make do this with union literal type
type Shape = 'triangle' | 'circle' | 'rectangle'

src/figuresClasses.ts
```diff
@@ -1,19 +1,73 @@
...
+    public color: Color,
+    public a: number,
+    public b: number,
+    public c: number,
+    public shape: Shape = Shape.Triangle,
+  ) {
+    if (this.a + this.b <= this.c
+      || this.a + this.c <= this.b
+      || this.b + this.c <= this.a) {
+      throw new Error('Triangle is invalid');
```
Dikey945, line 26: For more clear code better to transfer this check to a function and call it in constructor.

src/figuresClasses.ts
```diff
@@ -1,19 +1,73 @@
...
+    }
+  }
+
+  public getArea(): number {
+    const p: number = (this.a + this.b + this.c) / 2;
 
+    return Math.round(Math
+      .sqrt(p * (p - this.a) * (p - this.b) * (p - this.c)) * 100) / 100;
```
Dikey945, line 34: try to use more meaningful variables names

src/figuresClasses.ts
```diff
@@ -1,19 +1,73 @@
...
 
 export class Circle {
+  constructor(
+    public color: Color,
+    public radius: number,
+    public shape: Shape = Shape.Circle,
+  ) {
+    if (this.radius <= 0) {
+      throw new Error('Error. Radius must be > 0');
+    }
```
Dikey945, line 46: For more clear code better to transfer this check to a function and call it in constructor.

src/figuresClasses.ts
```diff
@@ -1,19 +1,73 @@
...
 export class Rectangle {
+  constructor(
+    public color: Color,
+    public width: number,
+    public height: number,
+    public shape: Shape = Shape.Rectangle,
+  ) {
+    if (this.width <= 0 || this.height <= 0) {
+      throw new Error('Error. Sides must be > 0');
+    }
```
Dikey945, line 63: For more clear code better to transfer this check to a function and call it in constructor.

Dikey945:
Good Job 🔥  Just left few comments

src/figuresClasses.ts
Anna-Tkachenko: fix @Dikey945 comment
create separate function that will accept the lengths of the figure sides and will throw an error if some of them is < 0
and use this function for all figures

Anna-Tkachenko (changes requested):

src/figuresClasses.ts
tatyana22: it would be better to use enum

src/figuresClasses.ts
tatyana22: this function will be used only one time

Denchepornyuk (approved):

PR #689, opened 2022-10-09
Solution

Proph7000:


src/figuresClasses.ts
```diff
@@ -1,19 +1,104 @@
...
+    color: Color,
+    a: number,
+    b: number,
+    c: number,
+  ) {
+    if (a <= 0 || b <= 0 || c <= 0) {
+      throw new Error('All sides must be greater than 0');
+    }
+
+    if ((a + b <= c) || (c + b <= a) || (a + c <= b)) {
```
TarasHoliuk, line 30: would be better to check the max number before, and then compare other numbers with the max number

src/figuresClasses.ts
```diff
@@ -1,19 +1,104 @@
...
+  a: number;
+
+  b: number;
+
+  constructor(
+    color: Color,
+    a: number,
+    b: number,
+  ) {
+    if (a <= 0 || b <= 0) {
```
TarasHoliuk, line 87: you are doing almost the same 3 times
create a function for it

TarasHoliuk (changes requested):

nazarmatsevych (approved):

PR #691, opened 2022-10-09
complete the task

barsuchellos:
image


Andrew-Skoryk (approved):
GJ!

src/figuresClasses.ts
```diff
@@ -1,19 +1,85 @@
...
+    public sideC: number,
+  ) {
+    if (sideA <= 0 || sideB <= 0 || sideC <= 0) {
+      throw new Error('Side must be >= 0');
+    }
 
+    if (this.sideB + this.sideC <= this.sideA
+      || this.sideA + this.sideC <= this.sideB
+      || this.sideA + this.sideB <= this.sideC) {
+      throw new Error('Invalid');
```
Anna-Tkachenko, line 31: check point 3 in checklist

src/figuresClasses.ts
```diff
@@ -1,19 +1,85 @@
...
 export class Circle {
+  shape: Shape = Shape.Circle;
+
+  constructor(
+    public color: Color,
+    public radius: number,
+  ) {
+    if (this.radius <= 0) {
+      throw new Error('Radius must be >= 0');
+    }
```
Anna-Tkachenko, line 57: you check input values for figures sizes in each class, will be better to create separate function for this purpose and call this function in all classes

Anna-Tkachenko (changes requested):

PR #693, opened 2022-10-10
solution

vitalii-babiienko:


src/figuresClasses.ts
```diff
@@ -1,19 +1,87 @@
...
+      throw new Error('Invalid input params!');
+    }
+
+    if ((a + b) <= c || (a + c) <= b || (b + c) <= a) {
+      throw new Error('It is not a triangle!');
+    }
+  }
 
+  getArea(): number {
+    const p: number = (this.a + this.b + this.c) / 2;
```
tetiana-ivanchenko, line 42: What is "p"? if you use math definitions, where "p" is perimeter - in your variable only half of it. I mean: let's try to use more explicit naming)

tetiana-ivanchenko (changes requested):

AlisaVlasova (approved):

PR #696, opened 2022-10-13
solution

mayuces:


kamal-tufekcic (approved):

src/figuresClasses.ts
```diff
@@ -1,19 +1,73 @@
...
+    public b: number,
+    public c: number,
+  ) {
+    if (a <= 0 || b <= 0 || c <= 0) {
+      throw new Error('Error... for triangle please check');
+    }
+
+    if (a + b <= c || a + c <= b || b + c <= a) {
+      throw new Error('Error... for triangle please check');
+    }
```
ihor-jpeg, line 24: maybe write a few private methods instead of doing checks in the constructor?

ihor-jpeg (approved):

PR #701, opened 2022-10-16
Figures classes

Yuriy-Komolov:


Andrew-Skoryk (approved):
GJ!

src/figuresClasses.ts
```diff
@@ -1,19 +1,95 @@
...
 
-export class Triangle {
+export class Triangle implements Figure {
+  shape: Shape = 'triangle';
+
+  constructor(
+    public color: Color,
+    public a: number,
+    public b: number,
+    public c: number,
```
silchyn, line 17: It is better not to make fields public when unnecessary, read more on the 'encapsulation OOP principle'

silchyn (changes requested):

src/figuresClasses.ts
```diff
@@ -1,19 +1,95 @@
...
+    private a: number,
+    private b: number,
+    private c: number,
+    public color: Color,
+  ) {
+    this.findErrors();
+  }
+
+  getArea(): number {
+    const halfPerim = (this.a + this.b + this.c) / 2;
```
naziamoff, line 23:
```suggestion
    const semiPerimeter = (this.a + this.b + this.c) / 2;
```

naziamoff (approved):

PR #704, opened 2022-10-18
added classes of figures

igor-1987:


src/figuresClasses.ts
```diff
@@ -1,19 +1,87 @@
...
 export class Circle {
+  shape = Shape.Circle;
 
+  constructor(
+    public color: Color,
+    private radius: number,
+  ) {
+    if (this.radius <= 0) {
+      throw new Error('Wrong radius!');
+    }
```
Anna-Tkachenko, line 59: you check input values for figure sizes in each constructor
will be better to create separate function for this purpose and call this function in each class

Anna-Tkachenko (approved):

PR #707, opened 2022-10-19
First commit

Bibicutu:


fatihsamur (approved):

src/figuresClasses.ts
```diff
@@ -1,19 +1,113 @@
...
+  shape: Shape;
 
-}
+  color: Color;
+
+  sideA: number;
+
+  sideB: number;
+
+  sideC: number;
```
VinnieJ-2k20, line 19: Use `public` in constructor to avoid double initiation.

src/figuresClasses.ts
```diff
@@ -1,19 +1,113 @@
...
+    color: Color,
+    sideA: number,
+    sideB: number,
+    sideC: number,
+  ) {
+    this.shape = 'triangle';
+    this.color = color;
+    this.sideA = sideA;
+    this.sideB = sideB;
+    this.sideC = sideC;
```
VinnieJ-2k20, line 31: Same. With `public`, this is not needed.

VinnieJ-2k20 (changes requested):
Fix children classes declarations.

ihor-jpeg (approved):

PR #721, opened 2022-11-03
Implement FiguresClasses task

anastasiia-tilikina:


src/figuresClasses.ts
```diff
@@ -1,19 +1,90 @@
+type Shape = 'triangle' | 'circle' | 'rectangle';
```
AlisaVlasova, line 1: let`s make this type enum

src/figuresClasses.ts
```diff
@@ -1,19 +1,90 @@
...
+  shape: Shape;
+  color: Color;
+  getArea(): number;
+}
 
+function round(area: number): number {
+  return Math.trunc(area * 100) / 100;
 }
+export class Triangle implements Figure {
+  shape: Shape = 'triangle';
```
AlisaVlasova, line 13: and then you can use it like this
```suggestion
  shape = Shape.TRIANGLE;
```

AlisaVlasova (approved):
goog job)

PR #723, opened 2022-11-06
solution

metinbicaksiz:


src/figuresClasses.ts
```diff
@@ -1,19 +1,111 @@
...
+
+    if (a >= b + c || b >= a + c || c >= b + a) {
+      throw new Error('Not A triangle!');
+    }
+  }
+
+  getArea(): number {
+    const size = (this.a + this.b + this.c) / 2;
+
+    return Math.floor((Math.sqrt(size
```
amitpatiljc, line 41: seems legit to introduce a rounding helper function to use it 3 times in your code

src/figuresClasses.ts
```diff
@@ -1,19 +1,111 @@
...
+      throw new Error('Negative sides!');
+    }
+
+    if (a >= b + c || b >= a + c || c >= b + a) {
+      throw new Error('Not A triangle!');
+    }
+  }
+
+  getArea(): number {
+    const size = (this.a + this.b + this.c) / 2;
```
amitpatiljc, line 39: wondering, would de-structuring reduce the character-count:
const { a, b, c } = this;

src/figuresClasses.ts
```diff
@@ -1,19 +1,111 @@
...
+
+  if (figure instanceof Rectangle) {
+    return `A ${figure.color} rectangle - ${figure.getArea()}`;
+  }
+
+  if (figure instanceof Triangle) {
+    return `A ${figure.color} triangle - ${figure.getArea()}`;
+  }
 
+  return `A ${figure.color} triangle - ${figure.getArea()}`;
```
amitpatiljc, line 110: instead of all these you can do it like this , how ? find out .
if stuck let me know 
```suggestion
export function getInfo(figure: Figure): string {
  return `A ${figure.color} ${figure.shape} - ${figure.getArea()}`;
```

amitpatiljc (changes requested):

src/figuresClasses.ts
amitpatiljc: Not applied , please apply this comment

amitpatiljc (changes requested):

metinbicaksiz:
Actualy, I did the changes but I dont see it in the pushed file. I made a minor change and pushed again.

getArea(): number {
    const { a, b, c } = this;
    const size = (a + b + c) / 2;

    return Math.floor((Math.sqrt(size
      * (size - a) * (size - b) * (size - c))) * 100) / 100;
  }

amitpatiljc (changes requested):
introduce rounding function somthing like this
`function rounding(number: number): number {
  return Math.floor(number * 100) / 100;
}`

qoa11a (approved):
Good job! 👍

PR #724, opened 2022-11-07
Solution

bokorzita:


src/figuresClasses.ts
```diff
@@ -1,19 +1,109 @@
...
+  }
+
+  getArea(): number {
+    const perimeter = (this.a + this.b + this.c) / 2;
+    const area = Math.sqrt(
+      perimeter
+      * (perimeter - this.a) * (perimeter - this.b) * (perimeter - this.c),
+    );
+
+    return Math.floor(area * 100) / 100;
```
amitpatiljc, line 54: seems legit to introduce a rounding helper function to use it 3 times in your code

src/figuresClasses.ts
```diff
@@ -1,19 +1,109 @@
...
+    this.a = a;
+    this.b = b;
+    this.c = c;
+  }
+
+  getArea(): number {
+    const perimeter = (this.a + this.b + this.c) / 2;
+    const area = Math.sqrt(
+      perimeter
+      * (perimeter - this.a) * (perimeter - this.b) * (perimeter - this.c),
```
amitpatiljc, line 51: wondering, would de-structuring reduce the character-count:
const { a, b, c } = this;

amitpatiljc (changes requested):

PR #725, opened 2022-11-08
Solution

SergIats:


src/figuresClasses.ts
```diff
@@ -1,19 +1,75 @@
...
+    return Math.floor(triangleSquare * 100) / 100;
+  }
 }
 
-export class Rectangle {
+export class Circle implements Figure {
+  shape: 'circle' = 'circle';
+
+  constructor(
+    public color: 'red'|'green'|'blue',
```
liliia-livandovskaya, line 38: Let's create `type color` and use it instead of duplicating `'red'|'green'|'blue'`

src/figuresClasses.ts
```diff
@@ -1,19 +1,75 @@
...
+    const triangleSquare: number
+     = ((sp * (sp - this.a) * (sp - this.b) * (sp - this.c)) ** 0.5);
 
+    return Math.floor(triangleSquare * 100) / 100;
+  }
 }
 
-export class Rectangle {
+export class Circle implements Figure {
+  shape: 'circle' = 'circle';
```
liliia-livandovskaya, line 35: Let's create a `type shape` and use it there instead of hardcoding type in each class

src/figuresClasses.ts
SergIats: Lilia, hi! Thank you for the review, I have implemented your recommendations!

src/figuresClasses.ts
```diff
@@ -1,19 +1,78 @@
...
+  ) {
+    if (radius <= 0) {
+      throw new Error('Radius should be more than zero');
+    }
+  }
 
+  getArea(): number {
+    const circleSquare: number = Math.PI * (this.radius ** 2);
+
+    return Math.floor(circleSquare * 100) / 100;
```
vlad-demchuk, line 52: you do the same actions many times `Math.floor(...`. let's create a separate function to avoid repetitions

vlad-demchuk (changes requested):

src/figuresClasses.ts
```diff
@@ -1,19 +1,78 @@
...
+    }
 
-export class Circle {
+    if (a >= (b + c) || b >= (a + c) || c >= (b + a)) {
+      throw new Error('Any side must be less than the sum of two other sides');
+    }
+  }
 
+  getArea(): number {
+    const sp: number = (this.a + this.b + this.c) / 2;
```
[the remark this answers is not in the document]
SergIats: Hi, Vlad! 'sp' is 'SemiPerimetr') You're right - the whole world is more understandable - I have changed it.

src/figuresClasses.ts
SergIats: No problem! Thanks for recommendations!

liliia-livandovskaya (approved):
🔥 🔥 🔥 🔥 Great job 🔥 🔥 🔥 🔥

PR #726, opened 2022-11-08
solution

MaxFedak:


src/figuresClasses.ts
```diff
@@ -1,19 +1,82 @@
...
+export class Circle implements Figure {
+  shape: Shape = 'circle';
 
-}
+  constructor(
+    public color: Color,
+    public radius: number,
+  ) {
+    if (radius <= 0) {
+      throw new Error('Radius cant be 0 or less');
```
iryna-lelyk, line 22: You can add one single quote like this `\'`

iryna-lelyk (approved):
Great job!

src/figuresClasses.ts
MaxFedak: yeah, thanks - just missed, that i need an apostrophe)

src/figuresClasses.ts
```diff
@@ -1,19 +1,82 @@
+type Color = 'red' | 'blue' | 'green';
+type Shape = 'circle' | 'rectangle' | 'triangle';
```
Olexandr6, line 2: You use Shape values in several places in your code, so it makes sense to make it `Enum`, not just `type`.
Use `Enum` here instead

src/figuresClasses.ts
```diff
@@ -1,19 +1,82 @@
...
+  getArea(): number;
+}
 
+export function roundNum(num : number) :number {
+  return Math.floor(num * 100) / 100;
 }
 
-export class Triangle {
+export class Circle implements Figure {
+  shape: Shape = 'circle';
```
Olexandr6, line 15: Enum

src/figuresClasses.ts
```diff
@@ -1,19 +1,82 @@
...
+    }
+  }
+
+  getArea(): number {
+    const semiPerimetr = (this.a + this.b + this.c) / 2;
+    const area = Math.sqrt(semiPerimetr * (semiPerimetr - this.a)
+    * (semiPerimetr - this.b) * (semiPerimetr - this.c));
+
+    return roundNum(area);
+  }
```
Olexandr6, line 77: Let's use object destructuring here:
```suggestion
  getArea(): number {
    const {
      a,
      b,
      c
    } = this;
    
    const semiPerimetr = (a + .b + c) / 2;
    const area = Math.sqrt(semiPerimetr * (semiPerimetr - a)
      * (semiPerimetr - b) * (semiPerimetr - c));

    return roundNum(area);
  }
```

Olexandr6 (changes requested):
Good job, overall 👍 Let's work on improvements:

PR #733, opened 2022-11-10
Develop

andreyolin:


denys-danyliuk (approved):
I agree with your mate comment, you can create separate function for that

PR #743, opened 2022-11-11
figureClasses

sh1gatsu:


src/figuresClasses.ts
```diff
@@ -1,19 +1,80 @@
...
+    public height: number,
+  ) {
+    if (width <= 0 || height <= 0) {
+      throw new Error('Bruh, we dont have at least one parametr');
+    }
+  }
 
+  getArea(): number {
+    return this.width * this.height;
+  }
```
Stepico, line 73: Why didn't you use floor here, we can set width or/and height like 45.63 and so on

Stepico (approved):
Good job, but in my opinion, you should write more precise things in your errors

src/figuresClasses.ts
sh1gatsu: my bad, didnt admit that <3

src/figuresClasses.ts
```diff
@@ -1,19 +1,82 @@
+type Shape = 'triangle' | 'circle' | 'rectangle';
```
illiarerun, line 1: use `enum` for Shape

src/figuresClasses.ts
```diff
@@ -1,19 +1,82 @@
...
+  getArea(): number;
+}
 
+function getRoundedDigit(digit: number): number {
+  return Math.floor(digit * 100) / 100;
 }
 
-export class Triangle {
+export class Triangle implements Figure {
+  public shape: Shape = 'triangle';
```
illiarerun, line 15: and afterwards, use it here, also redundant public modifier, fields are public by default, fix it everywhere
```suggestion
  shape: Shape = Shape.Triangle;
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,82 @@
...
+
+  constructor(
+    public color: Color,
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    if (a <= 0 || b <= 0 || c <= 0) {
+      throw new Error('We dont have one or maybe more sides!');
```
illiarerun, line 24: this doesn't mean that values hasn't been passed, this only means that values are lower than 0

illiarerun (changes requested):

illiarerun (approved):

PR #748, opened 2022-11-11
Solution

andrii-bassarab:


src/figuresClasses.ts
```diff
@@ -1,19 +1,87 @@
...
+      .reduce((sum, curr) => sum + curr) - biggestSide;
+
+    if (biggestSide >= sumOfSmallerSides) {
+      throw new Error('The longest side of the triangle must be less'
+      + 'than the sum of the other two');
+    }
+
+    if (this.a < 0 || this.b < 0 || this.c < 0
+      || biggestSide >= sumOfSmallerSides) {
+      throw new Error('Sides of the triangle must be bigger than 0');
```
Largoxoxo, line 30:
```suggestion
    if (this.a < 0 || this.b < 0 || this.c < 0) {
      throw new Error('Sides of the triangle must be bigger than 0');
```
Redundant check, you check for side biggest side size on line 23 already.

src/figuresClasses.ts
```diff
@@ -1,19 +1,87 @@
...
+    }
+  }
+
+  shape: Shape;
```
Largoxoxo, line 34: Move this declaration before the constructor.

src/figuresClasses.ts
```diff
@@ -1,19 +1,87 @@
...
 
-export class Circle {
+  getArea(): number {
+    const avergeSide = (this.a + this.b + this.c) / 2;
 
+    return Math.floor(Math.sqrt(
+      avergeSide * (avergeSide - this.a)
+      * (avergeSide - this.b) * (avergeSide - this.c),
+    ) * 100) / 100;
```
Largoxoxo, line 42: Consider using a function for rounding since you apply it three times.

Largoxoxo (approved):
GJ! Minor tweaks needed.

src/figuresClasses.ts
```diff
@@ -1,19 +1,87 @@
...
+    public color: Color,
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    this.shape = 'triangle';
 
-}
+    const biggestSide = Math.max(this.a, this.b, this.c);
+    const sumOfSmallerSides = [this.a, this.b, this.c]
```
mykola-koval, line 20: let's use object destructuring to avoid using `this` everywhere

src/figuresClasses.ts
```diff
@@ -1,19 +1,87 @@
...
+      .reduce((sum, curr) => sum + curr) - biggestSide;
+
+    if (biggestSide >= sumOfSmallerSides) {
+      throw new Error('The longest side of the triangle must be less'
+      + 'than the sum of the other two');
+    }
+
+    if (this.a < 0 || this.b < 0 || this.c < 0
+      || biggestSide >= sumOfSmallerSides) {
```
mykola-koval, line 29:
```suggestion
      ) {
```

src/figuresClasses.ts
mykola-koval: agree

src/figuresClasses.ts
mykola-koval: agree

mykola-koval (changes requested):

src/figuresClasses.ts
```diff
@@ -1,19 +1,97 @@
...
+    public color: Color,
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    this.shape = 'triangle';
+
+    const biggestSide = Math.max(a, b, c);
+    const sumOfSmallerSides = [a, b, c]
+      .reduce((sum, curr) => sum + curr) - biggestSide;
```
mykola-koval, line 27: do you really need a reduce here ? :)

mykola-koval (approved):

PR #749, opened 2022-11-12
solution

mitattt:


src/figuresClasses.ts
```diff
@@ -1,19 +1,84 @@
...
+    if (a >= b + c || b >= a + c || c >= a + b) {
+      throw new Error('Can not build the triangle!');
+    }
+  }
 
-export class Circle {
+  getArea(): number {
+    const semiPerimeter = (this.a + this.b + this.c) / 2;
+    const area = Math.sqrt((semiPerimeter - this.a) * (semiPerimeter - this.b)
+      * (semiPerimeter - this.c) * semiPerimeter);
```
Anna-Shapovalova, line 40: U can destructure the object to not repeat 'this' such many times)

src/figuresClasses.ts
```diff
@@ -1,19 +1,84 @@
...
 
+    return roundedNum(area);
+  }
 }
 
-export class Rectangle {
+export class Circle implements Figure {
+  shape: Shape = Shape.Circle;
 
+  constructor(public color: Color, public radius: number) {
```
Anna-Shapovalova, line 49:
```suggestion
  constructor(
    public color: Color, 
    public radius: number
  ) {
```
Maybe in this way, it looks more readable) Do U agree?

Anna-Shapovalova (approved):
Well done!) 🚀

src/figuresClasses.ts
mykola-koval, line 40:
```suggestion
    const { a, b, c } = this; 
    const area = Math.sqrt((semiPerimeter - a) * (semiPerimeter - b)
      * (semiPerimeter - c) * semiPerimeter);
```

src/figuresClasses.ts
mykola-koval: agree

mkyloa (changes requested):

src/figuresClasses.ts
```diff
@@ -1,19 +1,88 @@
...
+    }
+
+    if (a >= b + c || b >= a + c || c >= a + b) {
+      throw new Error('Can not build the triangle!');
+    }
+  }
 
-export class Circle {
+  getArea(): number {
+    const semiPerimeter = (this.a + this.b + this.c) / 2;
```
mykola-koval, line 38: you can declare it AFTER object destructuring and use values from there

mkyloa (approved):

PR #756, opened 2022-11-13
Solution

MikhailKucherina:


src/figuresClasses.ts
```diff
@@ -1,19 +1,79 @@
-export interface Figure {
+type Shape = 'triangle' | 'circle' | 'rectangle';
```
taraselnik, line 1: you can try with enum here

src/figuresClasses.ts
```diff
@@ -1,19 +1,79 @@
...
 
+export interface Figure {
+  shape: Shape;
+  color: Color;
+  getArea(): number;
 }
 
-export class Triangle {
+export class Triangle implements Figure {
+  shape: Shape = 'triangle';
```
taraselnik, line 11: and take a 'value by Key from enum' here

taraselnik (approved):

mykola-koval (changes requested):
let's use enums

mykola-koval (approved):

PR #774, opened 2022-11-18
Added a task solution

oleksusov:


src/figuresClasses.ts
```diff
@@ -1,19 +1,92 @@
...
+  public shape: Shape = 'circle';
+
+  constructor(
+    public color: Color,
+    public radius: number,
+  ) {
+    if (radius <= 0) {
+      throw new Error(`
+        The radius must be > 0
+      `);
```
Anastasiia-Svintsova, line 52: I think in this case it'll be ok

```suggestion
      throw new Error(`The radius must be > 0`);
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,92 @@
...
+    } else if (!(b + c > a) || !(a + c > b) || !(a + b > c)) {
+      throw new Error(`
+        The longest side of a triangle must be >= than the sum of the other two
+      `);
+    } else {
+      this.a = a;
+      this.b = b;
+      this.c = c;
+    }
```
Anastasiia-Svintsova, line 30: You don't need this else statement here

src/figuresClasses.ts
```diff
@@ -1,19 +1,92 @@
...
+  }
 
-export class Circle {
+  getArea(): number {
+    const s: number = (this.a + this.b + this.c) / 2;
```
Anastasiia-Svintsova, line 34: let's use object destructuring to avoid using `this` everywhere

src/figuresClasses.ts
```diff
@@ -1,19 +1,92 @@
...
+    public color: Color,
+    public radius: number,
+  ) {
+    if (radius <= 0) {
+      throw new Error(`
+        The radius must be > 0
+      `);
+    } else {
+      this.radius = radius;
+    }
```
Anastasiia-Svintsova, line 55: Same about `else` here. You already have it

src/figuresClasses.ts
```diff
@@ -1,19 +1,92 @@
...
+    public height: number,
+  ) {
+    if (width <= 0 || height <= 0) {
+      throw new Error(`
+      The length of any side must be > 0
+      `);
+    } else {
+      this.width = width;
+      this.height = height;
+    }
```
Anastasiia-Svintsova, line 80: Same

src/figuresClasses.ts
```diff
@@ -1,19 +1,92 @@
-export interface Figure {
+export type Shape = 'triangle' | 'circle' | 'rectangle';
+export type Color = 'red' | 'green' | 'blue';
 
+export interface Figure {
+  shape: Shape;
+  color: Color;
 }
```
Anastasiia-Svintsova, line 7: Forgot about getArea() value here

src/figuresClasses.ts
```diff
@@ -1,19 +1,92 @@
...
+  }
+
+  getArea(): number {
+    const area: number = this.width * this.height;
+
+    return Math.floor(area * 100) / 100;
+  }
+}
 
+export function getInfo(figure: Circle | Triangle | Rectangle): string {
```
Anastasiia-Svintsova, line 90:
```suggestion
export function getInfo(figure: Figure): string {
```

Anastasiia-Svintsova (changes requested):

TarasHoliuk (approved):

PR #779, opened 2022-11-25
solved

Olek-Lytvynenko:


src/figuresClasses.ts
```diff
@@ -1,19 +1,87 @@
...
+    public color: Color,
+    public width: number,
+    public height: number,
+  ) {
+    if (width <= 0 || height <= 0) {
+      throw new Error(
+        'A side of a rectangle cannot be equal to 0 or be negative.',
+      );
+    }
+  }
```
DarkMistyRoom, line 78: Better move all validation logic to a separate function.

src/figuresClasses.ts
```diff
@@ -1,19 +1,87 @@
...
+
+  getArea(): number {
+    return Math.floor(this.width * this.height * 100) / 100;
```
DarkMistyRoom, line 81: Rounding logic also can be moved to function

src/figuresClasses.ts
```diff
@@ -1,19 +1,87 @@
...
+        'A side of a triangle cannot be equal to 0 or be negative.',
+      );
+    }
+
+    const maxSide: number = Math.max(a, b, c);
+
+    if (maxSide >= (a + b + c) - maxSide) {
+      throw new Error(
+        'the longest side of the triangle must'
+         + 'be greater than the sum of the other two',
```
DarkMistyRoom, line 30:
```suggestion
        'the longest side of the triangle must '
         + 'be greater than the sum of the other two',
```

DarkMistyRoom: to have the space in the final string

DarkMistyRoom (changes requested):

SanyaBratashchuk (approved):

PR #781, opened 2022-11-26
Solution

UmizDemud:


src/figuresClasses.ts
```diff
@@ -1,19 +1,111 @@
...
+    if (_a < 0 || _b < 0 || _c < 0) {
+      throw new Error('Side of a triangle can not be negative');
+    }
+
+    if (_a >= _b + _c
+        || _b >= _a + _c
+        || _c >= _b + _a) {
+      throw new Error('the longest side of a triangle must be greater'
+        + 'than a sum of two others');
+    }
```
naziamoff, line 36: let's separate the condition into a variable

src/figuresClasses.ts
```diff
@@ -1,19 +1,111 @@
...
-export class Triangle {
+export class Triangle implements Figure {
+  color: Color;
 
-}
+  a: number;
+
+  b: number;
+
+  c: number;
```
naziamoff, line 17:
```suggestion
  a: number;
  b: number;
  c: number;
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,111 @@
...
 
-export class Circle {
+  shape: Shape;
 
+  constructor(
+    _color: Color,
+    _a: number,
+    _b: number,
+    _c: number,
```
naziamoff, line 25: let's use `private` here instead of `_`

naziamoff (changes requested):

src/figuresClasses.ts
UmizDemud: My linter was insistent about this

naziamoff (approved):
🔥

PR #782, opened 2022-11-27
TS_Figures task was done

illia-kots:


src/figuresClasses.ts
```diff
@@ -1,19 +1,81 @@
...
 
-}
+  constructor(
+    public color: Color,
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    if (a <= 0 || b <= 0 || c <= 0) {
+      throw new Error('your error message');
```
Bohdan-Maksymiuk, line 20:
```suggestion
      throw new Error('*********');
```
mb add some meaningful messages everywhere?

Bohdan-Maksymiuk (changes requested):

PR #785, opened 2022-12-01
ts_figures_classes

kolya-movchan:


kolya-movchan:
finished task with git commit --no-verify -am considering the error not fatal 
image
image

tsconfig.json
```diff
@@ -1,4 +1,4 @@
-{
+https://github.com/kolya-movchan/ts_figures_classes.git{
```
TarasHoliuk, line 1: why did you do it? keep this file untouched
that causes a linter error and other tests can't run

src/figuresClasses.ts
```diff
@@ -1,19 +1,85 @@
-export interface Figure {
+function convertNumber(number: number): number {
+  return Math.trunc(number * 100) / 100;
+}
 
+export interface Figure {
+  shape: 'triangle' | 'circle' | 'rectangle',
+  color: 'red' | 'green' | 'blue',
```
TarasHoliuk, line 7:
```suggestion
  shape: 'triangle' | 'circle' | 'rectangle',
  color: 'red' | 'green' | 'blue',
```
Create types `Color` and `Shape` above and use them here and below

src/figuresClasses.ts
```diff
@@ -1,19 +1,85 @@
...
+  getArea(): number
 }
 
 export class Triangle {
+  public shape: Figure['shape'] = 'triangle';
+
+  constructor(
+    public color: Figure,
```
TarasHoliuk, line 15:
```suggestion
    public color: Color,
```

TarasHoliuk (changes requested):

tsconfig.json
kolya-movchan: Literally, I don't know how that happened :) 
That was an incident I suppose

tetiana-ivanchenko (approved):

PR #786, opened 2022-12-01
Solution

Julia2063:


src/figuresClasses.ts
```diff
@@ -1,19 +1,90 @@
...
+    public c: number,
+  ) {
+    const arr = [a, b, c];
+
+    if (arr.find((el) => el <= 0)) {
+      throw new Error('your error message');
+    }
 
+    const maxSide = arr.sort((x, y) => x - y).pop();
+    const restSidesSum = arr.reduce((x, y) => x + y);
```
loralevitska, line 35: it is good practice to pass the initialValue to reduce

src/figuresClasses.ts
```diff
@@ -1,19 +1,90 @@
...
+
+    if (maxSide && maxSide >= restSidesSum) {
+      throw new Error('your error message');
+    }
+  }
+
+  getArea(): number {
+    const sPer = (this.a + this.b + this.c) / 2;
```
loralevitska, line 43: don't cut variable name. Try to change for more meaningful name

loralevitska (changes requested):
almost done!

loralevitska (approved):
good job!

PR #788, opened 2022-12-04
Solution by DK

DimaK123:


src/figuresClasses.ts
```diff
@@ -1,19 +1,68 @@
...
 }
 
-export class Rectangle {
+export class Circle implements Figure {
+  constructor(
+    public color: string,
+    public radius: number,
+  ) {
+    if (radius <= 0) {
+      throw new Error('your error message');
```
loralevitska, line 35: change the text ``your error message'' to something more meaningful. Fix it everywhere

loralevitska (changes requested):

src/figuresClasses.ts
DimaK123: Done

PR #793, opened 2022-12-10
`Figure`

smiyka:


src/figuresClasses.ts
```diff
@@ -1,19 +1,78 @@
-export interface Figure {
+type Shape = 'triangle' | 'circle' | 'rectangle';
```
AlisaVlasova, line 1: letzs make it enum

src/figuresClasses.ts
```diff
@@ -1,19 +1,78 @@
...
 
+export interface Figure {
+  shape: Shape;
+  color: Color;
+  getArea(): number,
 }
 
-export class Triangle {
+export class Triangle implements Figure {
+  public shape: Shape = 'triangle';
```
AlisaVlasova, line 11: so you can use it here
```suggestion
  public shape: Shape = Shape.Triangle;
```

AlisaVlasova (changes requested):

PR #794, opened 2022-12-11
ts_figures_classes

ysrckr:


src/figuresClasses.ts
```diff
@@ -1,19 +1,88 @@
-export interface Figure {
 
+type Color = 'red' | 'green' | 'blue';
+type Shape = 'triangle' | 'circle' | 'rectangle';
```
naziamoff, line 3: would be better to make Shape a enum

naziamoff (approved):
LGTM!

src/figuresClasses.ts
ysrckr: I think both would benefit from enum. In real project there might be much more shapes and colors. I just made it type for simplicity.

PR #795, opened 2022-12-13
Solution

s3rhiienko:


src/figuresClasses.ts
```diff
@@ -1,19 +1,68 @@
...
 }
 
-export class Rectangle {
+export class Circle implements Figure {
+  constructor(
+    public color: string,
+    public radius: number,
+  ) {
+    if (radius <= 0) {
+      throw new Error('Error! This is not a circle');
```
polosanya, line 35: Would be great to clarify why exactly it's not a circle

src/figuresClasses.ts
```diff
@@ -1,19 +1,68 @@
...
 
-export function getInfo(figure) {
+export class Rectangle implements Figure {
+  constructor(
+    public color: string,
+    public a: number,
+    public b: number,
+  ) {
+    if (a <= 0 || b <= 0) {
+      throw new Error('Error! This is not rectangle');
```
polosanya, line 55: Same here

polosanya (changes requested):
Almost done! Let's improve your code a bit)

Anastasiia-Svintsova (approved):

PR #796, opened 2022-12-13
Solution

vadiimvooo:


nikachu404 (approved):
Looks nice, good job!=)

src/figuresClasses.ts
```diff
@@ -1,19 +1,95 @@
...
+    public b: number,
+    public c: number,
+  ) {
+    if (this.isValidTriangle()) {
+      throw new Error(`Sides ${a}, ${b} and ${c} can't form a triangle`);
+    }
+  }
 
-export class Circle {
+  isValidTriangle: () => boolean = () => {
```
solaryasha, line 24: Good method, let's make it private, cause we are using it only internally

src/figuresClasses.ts
```diff
@@ -1,19 +1,95 @@
...
+    return this.a === 0
+    || this.b === 0
+    || this.c === 0
+    || this.a + this.b <= this.c
+    || this.b + this.c <= this.a
+    || this.c + this.a <= this.b;
```
solaryasha, line 30: Let's pack the conditions into the array and check it with `some` method

src/figuresClasses.ts
solaryasha: If the method has name `isValidTriangle`, it should return `true` if it's value, and `false` if it's invalid

src/figuresClasses.ts
```diff
@@ -1,19 +1,95 @@
...
+    public color: Color,
+    public radius: number,
+  ) {
+    if (this.isValidCircle()) {
+      throw new Error(`Radius ${radius} can't form the circle`);
+    }
+  }
+
+  isValidCircle: () => boolean = () => {
+    return this.radius < 1;
```
solaryasha, line 59: What if we have a radius of 0.9 meters?

src/figuresClasses.ts
```diff
@@ -1,19 +1,95 @@
...
+  }
+
+  getArea(): number {
+    const area = this.a * this.b;
+
+    return Math.floor(100 * area) / 100;
+  }
+
+  isValidRectangle: () => boolean = () => {
+    return this.a < 1 || this.b < 1;
```
solaryasha, line 89: What if we have sides of 0.9 meters?

solaryasha (changes requested):

solaryasha (approved):
👍

PR #802, opened 2022-12-15
ts figures classes

kalmykov-mykhailo:


src/figuresClasses.ts
```diff
@@ -1,19 +1,73 @@
...
 
-}
+  constructor(
+    public color: Color,
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    if (a <= 0 || b <= 0 || c <= 0 || a + b <= c || a + c <= b || c + b <= a) {
+      throw new Error('your error message');
```
vlad-demchuk, line 20: specify type of erro. eg `One or more sides less than 0` and so on

src/figuresClasses.ts
```diff
@@ -1,19 +1,73 @@
-export interface Figure {
+type Shape = 'triangle' | 'circle' | 'rectangle';
+type Color = 'red' | 'green' | 'blue';
```
vlad-demchuk, line 2: use enums here

vlad-demchuk (changes requested):

tetiana-ivanchenko (approved):

PR #803, opened 2022-12-15
solution

fallowlone:


src/figuresClasses.ts
```diff
@@ -1,19 +1,104 @@
...
+  }
+
+  isValidParameters(): void {
+    if (
+      this.a <= 0
+      || this.b <= 0
+      || this.c <= 0
+    ) {
+      throw new Error('side(s) can\'t be less then 0');
+    }
```
Nazar-Brunarskyi, line 30: it seems like you can do these operations inside a constructor because every class has a different implementation of this function

Nazar-Brunarskyi (approved):

src/figuresClasses.ts
```diff
@@ -1,19 +1,104 @@
...
 
-export class Circle {
+    if (
+      this.a + this.b <= this.c
+      || this.a + this.c <= this.b
+      || this.b + this.c <= this.a
```
Ihor-Karpyn, line 35: Let's move to method `isValidTriangle`

src/figuresClasses.ts
```diff
@@ -1,19 +1,104 @@
...
+    public color: Color,
+    public radius: number,
+  ) {
+    this.shape = 'circle';
+    this.isValidParameters();
+  }
 
+  isValidParameters(): void {
+    if (this.radius <= 0) {
+      throw new Error('radius can\'n be less then 0');
```
Ihor-Karpyn, line 64:
```suggestion
      throw new Error('Radius should be greater than 0');
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,104 @@
...
+  constructor(
+    public color: Color,
+    public width: number,
+    public height: number,
+  ) {
+    this.shape = 'rectangle';
+    this.isValidParameters();
+  }
+
+  isValidParameters(): void {
```
Ihor-Karpyn, line 87: if the function start from `is` it should return boolean

```suggestion
  validateParameters(): void {
```

Ihor-Karpyn (changes requested):

src/figuresClasses.ts
ivan-kadykalo: Fix this comment

src/figuresClasses.ts
```diff
@@ -1,19 +1,113 @@
...
+  ) {
+    this.shape = 'circle';
+    this.isValidParameters();
+  }
+
+  isValidParameters(): void {
+    if (this.radius <= 0) {
+      throw new Error('radius should be greater then 0');
+    }
+  }
```
ivan-kadykalo, line 75: The same

ivan-kadykalo (changes requested):
You are almost there 💪

ivan-kadykalo (approved):
Good job 🔥

PR #807, opened 2022-12-16
solution

AntonMartynes:


oleksandrnikora (approved):
nice)

src/figuresClasses.ts
```diff
@@ -1,19 +1,100 @@
...
+  color: Color;
+  getArea(): number;
 }
 
-export class Triangle {
+export class Triangle implements Figure {
+  shape: Shape;
 
-}
+  isValidTriangle(): boolean {
```
solaryasha, line 13: If the method has a name `isValidTriangle` - we should return true if it's valid

solaryasha: Let's make the private method, cause you are using it only internally

src/figuresClasses.ts
```diff
@@ -1,19 +1,100 @@
...
+    return (
+      this.a <= 0
+      || this.b <= 0
+      || this.c <= 0
+      || this.a + this.b <= this.c
+      || this.a + this.c <= this.b
+      || this.b + this.c <= this.a
```
solaryasha, line 20: Let's store these conditions in the array and check it with the `some` method

src/figuresClasses.ts
```diff
@@ -1,19 +1,100 @@
...
 
+export interface Figure {
+  shape: Shape;
+  color: Color;
+  getArea(): number;
 }
 
-export class Triangle {
+export class Triangle implements Figure {
+  shape: Shape;
```
```suggestion
  shape: Shape = 'triangle';
```

solaryasha (changes requested):

src/figuresClasses.ts
```diff
@@ -1,19 +1,95 @@
...
+  getArea(): number;
 }
 
-export class Triangle {
+export class Triangle implements Figure {
+  shape: Shape = 'triangle';
 
-}
+  private isValidTriangle(): boolean {
+    const triangleConditions = [
```
solaryasha, line 14:
```suggestion
    const invalidConditions = [
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,95 @@
...
+    return Math.floor(area * 100) / 100;
+  }
 }
 
-export function getInfo(figure) {
+export class Rectangle implements Figure {
+  shape: Shape = 'rectangle';
+
+  private isValideRectangle(): boolean {
+    return (this.width >= 0 && this.height >= 0);
```
solaryasha, line 73:
```suggestion
    return this.width >= 0 && this.height >= 0;
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,95 @@
...
+  }
+
+  getArea(): number {
+    const area: number = this.width * this.height;
```
solaryasha, line 77:
```suggestion
    return this.width * this.height;
```

PR #808, opened 2022-12-16
solution

lefteris-L:


src/figuresClasses.ts
```diff
@@ -1,19 +1,74 @@
-export interface Figure {
+export type Shape = 'circle' | 'triangle' | 'rectangle';
+export type Color = 'red' | 'green' | 'blue';
```
ysrckr, line 2: Just a suggestion -> You can use `enum` type here.

ysrckr (approved):
LGTM 🤟🏻 Only one suggestion that would make it easier to expand the code.

src/figuresClasses.ts
lefteris-L: Works like a charm. thanks

meforoma (approved):
charming 👍

PR #809, opened 2022-12-17
solution

anastasiia-roiek:


src/figuresClasses.ts
```diff
@@ -1,19 +1,76 @@
-export interface Figure {
+type Shape = 'triangle' | 'circle' | 'rectangle';
```
4ebupel, line 1: Let's practice enum (optional)

4ebupel (approved):
Good job, not confident about your `if` block inside a `triangle class`, I made a function for that purpose, stored all the conditions inside an array and used `some` method to get a `boolean`, but to my opinion your implementation is still good tho🔥

src/figuresClasses.ts
```diff
@@ -1,19 +1,76 @@
...
+    public c: number,
+  ) {
+    if (
+      a <= 0
+      || b <= 0
+      || c <= 0
+      || a + b <= c
+      || b + c <= a
+      || a + c <= b
+    ) {
```
solaryasha, line 26: Let's pack all of the conditions to the array and check it with `some` method

src/figuresClasses.ts
```diff
@@ -1,19 +1,76 @@
...
+      throw new Error('Incorrect numbers!');
```
solaryasha, line 27: Why are these numbers incorrect?

solaryasha (changes requested):

solaryasha (approved):

PR #811, opened 2022-12-17
Solution

nurdanokcu:


src/figuresClasses.ts
```diff
@@ -1,19 +1,85 @@
...
-export class Triangle {
+enum Color {
+  red = 'red',
+  green = 'green',
+  blue = 'blue'
+}
 
+export interface Figure {
+  shape: string;
+  color: string;
```
ysrckr, line 15:
```suggestion
  shape: Shape;
  color: Color;
```
You can do this.

src/figuresClasses.ts
```diff
@@ -1,19 +1,85 @@
...
+  getArea: () => number;
 }
 
-export class Circle {
+export class Triangle implements Figure {
+  public shape: Shape = Shape.triangle;
```
ysrckr, line 20:
```suggestion
  shape = Shape.triangle;
```
If you follow suggestion above, you can assign this way. 
---
You don't need `public` here. `public` is default. In `constructor` we write it because otherwise we have to assign it in the `constructor`.

ysrckr (changes requested):
Just a small change after that everything looks perfect.

src/figuresClasses.ts
nurdanokcu: Thank you Yasar.  I get it now.

src/figuresClasses.ts
```diff
@@ -1,19 +1,85 @@
...
+    private b: number,
+    private c: number,
+  ) {
+    if (a <= 0 || b <= 0 || c <= 0) {
+      throw new Error('Sides of a triangle must be pozitive numbers');
+    }
+
+    if (a >= b + c || b >= a + c || c >= a + b) {
+      throw new Error('Sides '
+      + `${this.a}, ${this.b} and ${this.c} `
```
meforoma, line 34: wondering - in conditional block you use a,b,c without `this.`,
but in string literal with `this.` 🙂

meforoma (approved):
nicely done

src/figuresClasses.ts
nurdanokcu: I didn't realize it , I will give it a look and fix it. 👍
Thank you.

PR #814, opened 2022-12-18
Solution

vrima1221:


src/figuresClasses.ts
```diff
@@ -1,19 +1,87 @@
...
+  color: Color,
+  getArea(): number,
 }
 
-export class Triangle {
+export class Triangle implements Figure {
+  public shape: Shape = 'triangle';
 
-}
+  checkTriangle(): boolean {
```
4ebupel, line 13: Nice method, but let's make it private 'cause you are using it only internally

src/figuresClasses.ts
```diff
@@ -1,19 +1,87 @@
...
+    return this.a <= 0
+      || this.b <= 0
+      || this.c <= 0
+      || this.a >= this.b + this.c
+      || this.b >= this.a + this.c
+      || this.c >= this.a + this.b;
+  }
```
4ebupel, line 20: Let's store these conditions in the array and check it with the Array.prototype.some() method

src/figuresClasses.ts
```diff
@@ -1,19 +1,87 @@
...
+    public color: Color,
+    public radius: number,
+  ) {
+    if (this.radius <= 0) {
+      throw new Error('Radius <= 0');
+    }
+  }
 
+  getArea(): number {
+    const area = Math.floor(Math.PI * this.radius * this.radius * 100) / 100;
```
4ebupel, line 59: maybe it would be better to write it like this: `this.radius ** 2`

src/figuresClasses.ts
```diff
@@ -1,19 +1,87 @@
-export interface Figure {
+type Shape = 'triangle' | 'circle' | 'rectangle';
```
4ebupel, line 1: You also may try the enum if u want

4ebupel (changes requested):
In general, very good work, just a few fixes

PR #815, opened 2022-12-18
Solution

4ebupel:


src/figuresClasses.ts
```diff
@@ -1,19 +1,94 @@
...
+  ) {
+    this.shape = 'circle';
+
+    if (this.radius <= 0) {
+      throw new Error('You cannot make a circle with such radius');
+    }
+  }
 
+  getArea(): number {
+    const area = Math.PI * this.radius * this.radius;
```
anastasiia-bilska, line 64: maybe it would be better to write it like this: ```this.radius ** 2```

anastasiia-bilska (approved):
Great!

src/figuresClasses.ts
```diff
@@ -1,19 +1,94 @@
+type Shape = 'triangle' | 'circle' | 'rectangle';
```
solaryasha, line 1: Let's practice the `enum`

src/figuresClasses.ts
```diff
@@ -1,19 +1,94 @@
...
+    public c: number,
+  ) {
+    this.shape = 'triangle';
+
+    if (this.isValidTriangle()) {
+      throw new Error('You cannot make a triangle with such sides');
+    }
+  }
+
+  isValidTriangle(): boolean {
```
solaryasha, line 27: It is a cool method and I like it. Let's make it private, cause you are using it only internally

src/figuresClasses.ts
```diff
@@ -1,19 +1,94 @@
...
+    return (
+      this.a <= 0
+      || this.b <= 0
+      || this.c <= 0
+      || this.a + this.b <= this.c
+      || this.b + this.c <= this.a
+      || this.c + this.a <= this.b);
```
solaryasha, line 34: Let's store these conditions in the array and check it with the some method

src/figuresClasses.ts
solaryasha: It's a good point, let's fix it

solaryasha (changes requested):

src/figuresClasses.ts
4ebupel: fixed

src/figuresClasses.ts
4ebupel: fixed

src/figuresClasses.ts
4ebupel: fixed

src/figuresClasses.ts
```diff
@@ -1,19 +1,102 @@
...
+    const conditions = [
+      this.a <= 0,
+      this.b <= 0,
+      this.c <= 0,
+      this.a + this.b <= this.c,
+      this.b + this.c <= this.a,
+      this.c + this.a <= this.b,
+    ];
 
+    return conditions.some((condition) => condition);
```
solaryasha, line 42:
```suggestion
    return !conditions.some(Boolean);
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,102 @@
...
-}
+  constructor(
+    public color: Color,
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    this.shape = Shape.Triangle;
+
+    if (this.isValidTriangle()) {
```
```suggestion
    if (!this.isValidTriangle()) {
```

src/figuresClasses.ts
solaryasha: If the method has a name `isValidTriangle` - we should return true if it's valid

solaryasha (changes requested):

src/figuresClasses.ts
4ebupel: Good point, fixed

src/figuresClasses.ts
```diff
@@ -1,19 +1,102 @@
...
+      * (halfPerimeter - this.b) * (halfPerimeter - this.c),
+    );
+
+    return Number(area.toFixed(2));
+  }
 }
 
-export class Rectangle {
+export class Circle implements Figure {
+  shape: Shape;
```
solaryasha, line 58:
```suggestion
  shape = Shape.Circle;
```

solaryasha (approved):

PR #816, opened 2022-12-18
Solution

NadiiaKoch:


anastasiia-roiek (approved):
ok!

src/figuresClasses.ts
```diff
@@ -1,19 +1,81 @@
...
+    this.shape = 'triangle';
 
-export class Circle {
+    if (this.a <= 0
+      || this.b <= 0
+      || this.c <= 0
+      || this.a + this.b <= this.c
+      || this.a + this.c <= this.b
+      || this.b + this.c <= this.a) {
+      throw new Error('These sides can not form a triangle');
```
ivan-kadykalo, line 27: How as a user I should know which values are valid and which are not?
Type inside the error the reason. Imagine yourself as a user, will what will you do after that message?

ivan-kadykalo (changes requested):
You are almost there 💪

src/figuresClasses.ts
NadiiaKoch: fix

ivan-kadykalo (approved):
Good job 🔥

PR #817, opened 2022-12-18
Solution

OlggaJul:


anastasiia-bilska (approved):
Great!

src/figuresClasses.ts
```diff
@@ -1,19 +1,97 @@
...
+  getArea(): number {
+    const semiPerimeter = (this.sideA + this.sideB + this.sideC) / 2;
+
+    return +Math.sqrt(Math.abs(semiPerimeter
+      * (semiPerimeter - this.sideA)
+      * (semiPerimeter - this.sideB)
+      * (semiPerimeter - this.sideC))).toFixed(2);
+  }
+
+  isInvalidShape(): boolean {
```
solaryasha, line 22: Wow, cool method, I like it! Let's make it private, cause you are using it only internally

src/figuresClasses.ts
```diff
@@ -1,19 +1,97 @@
...
+    const arr = [this.sideA, this.sideB, this.sideC];
```
solaryasha, line 23:
```suggestion
    const sides = [this.sideA, this.sideB, this.sideC];
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,97 @@
...
+
+    const maxSide = arr.sort((a, b) => a - b).pop();
```
solaryasha, line 25: `Math.min` will have a better performance, but for your case this is fine.

src/figuresClasses.ts
```diff
@@ -1,19 +1,97 @@
...
+    const twoSideSum = arr.reduce((sum, current) => sum + current);
+    const isInvalidValue = arr.some((side) => side <= 0);
+
+    if (!maxSide) {
+      return false;
+    }
```
solaryasha, line 31: Could this condition ever happen?

src/figuresClasses.ts
```diff
@@ -1,19 +1,97 @@
...
+      * (semiPerimeter - this.sideB)
+      * (semiPerimeter - this.sideC))).toFixed(2);
+  }
+
+  isInvalidShape(): boolean {
+    const arr = [this.sideA, this.sideB, this.sideC];
+
+    const maxSide = arr.sort((a, b) => a - b).pop();
+    const twoSideSum = arr.reduce((sum, current) => sum + current);
+    const isInvalidValue = arr.some((side) => side <= 0);
```
```suggestion
    const hasNegativeValue = arr.some((side) => side <= 0);
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,97 @@
...
+    return (maxSide >= twoSideSum) || isInvalidValue;
+  }
 
+  constructor(
+    public color: Colors,
+    public sideA: number,
+    public sideB: number,
+    public sideC: number,
+  ) {
+    this.shape = 'triangle';
```
solaryasha, line 42: Maybe, let's try to create an enum with the values.

src/figuresClasses.ts
```diff
@@ -1,19 +1,97 @@
...
+    }
+  }
 }
 
-export function getInfo(figure) {
+export class Rectangle implements Figure {
+  shape: Shapes;
+
+  getArea(): number {
+    return +(this.width * this.height).toFixed(2);
```
solaryasha, line 77:
```suggestion
    return Number((this.width * this.height).toFixed(2));
```

src/figuresClasses.ts
solaryasha, line 22:
```suggestion
  hasInvalidShape(): boolean {
```

solaryasha (changes requested):
Good job, I left you some thoughts about improvements.

src/figuresClasses.ts
OlggaJul: +

OlggaJul: '+'

src/figuresClasses.ts
OlggaJul: +

src/figuresClasses.ts
OlggaJul: +

src/figuresClasses.ts
OlggaJul: without this condition i have:
const maxSide: number | undefined
Object is possibly 'undefined'

src/figuresClasses.ts
OlggaJul: +

src/figuresClasses.ts
OlggaJul: +

src/figuresClasses.ts
solaryasha: You could create a type for `sides` as a tuple. (like [number, number, number]. It should be fine.

src/figuresClasses.ts
```diff
@@ -1,19 +1,101 @@
-export interface Figure {
+type Colors = 'red' | 'green' | 'blue';
 
+enum Shapes {
+  triangle = 'triangle',
```
solaryasha, line 4: By the convention enum keys always start with the capital letter
```suggestion
  Triangle = 'triangle',
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,101 @@
...
+    public color: Colors,
+    public sideA: number,
+    public sideB: number,
+    public sideC: number,
+  ) {
+    this.shape = Shapes.triangle;
+
+    if (this.hasInvalidShape()) {
+      throw new Error(
+        'It`s not possible to create a triangle with these parameters',
```
solaryasha, line 50: You could print out the params in the error

solaryasha (approved):

src/figuresClasses.ts
OlggaJul: +

src/figuresClasses.ts
OlggaJul: +

PR #822, opened 2022-12-19
Solution

batstolya:


src/figuresClasses.ts
```diff
@@ -1,19 +1,97 @@
...
+
+    if (a <= 0
+      || b <= 0
+      || c <= 0
+      || a + b <= c
+      || b + c <= a
+      || a + c <= b
+    ) {
+      throw new Error('side of triangle is wrong');
+    }
```
viktoriia-kravchenko, line 35: It would be great to explain what exactly is wrong with the side of the triangle, what the user should write to satisfy the condition.

src/figuresClasses.ts
```diff
@@ -1,19 +1,97 @@
...
+
+  radius: number;
 
+  constructor(color: Color, radius: number) {
+    this.color = color;
+    this.radius = radius;
+
+    if (this.radius <= 0) {
+      throw new Error('your radius is incorrect');
+    }
```
viktoriia-kravchenko, line 59: Same here.

src/figuresClasses.ts
```diff
@@ -1,19 +1,97 @@
...
+type Color = 'red' | 'green' | 'rectangle';
 
+export interface Figure {
+  shape: Shape,
+  color: Color,
+  getArea(): number,
 }
 
 export class Triangle {
+  shape: Shape = 'triangle';
```
viktoriia-kravchenko, line 11: Every class should implement Figure interface.
In this case, "export class Triangle implements Figure".

src/figuresClasses.ts
```diff
@@ -1,19 +1,97 @@
...
+  height: number;
+
+  constructor(color: Color, width: number, height: number) {
+    this.color = color;
+    this.width = width;
+    this.height = height;
+
+    if (width <= 0 || height <= 0) {
+      throw new Error('your width or height is incorrect');
+    }
```
viktoriia-kravchenko, line 85: And here.

viktoriia-kravchenko (changes requested):
GJ! Only minor improvements can be made.

viktoriia-kravchenko (approved):
🔥

src/figuresClasses.ts
```diff
@@ -1,19 +1,104 @@
...
+  shape: Shape = 'triangle';
 
-}
+  color: Color;
+
+  a: number;
+
+  b: number;
+
+  c: number;
```
ivan-kadykalo, line 19:
```suggestion
  a: number;
  b: number;
  c: number;
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,104 @@
...
 
+export interface Figure {
+  shape: Shape,
+  color: Color,
+  getArea(): number,
 }
 
-export class Triangle {
+export class Triangle implements Figure {
+  shape: Shape = 'triangle';
```
ivan-kadykalo, line 11: Better to use enum
```suggestion
  shape: Shape = Shape.Triangle;
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,104 @@
...
+    this.a = a;
+    this.b = b;
+    this.c = c;
 
+    if (
+      a <= 0
+      || b <= 0
+      || c <= 0
+    ) {
+      throw new Error('All sides of a triangle mush be positive numbers');
```
ivan-kadykalo, line 32:
```suggestion
      throw new Error('All sides of a triangle must be positive numbers');
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,104 @@
...
+    }
+
+    if (
+      a + b <= c
+      || b + c <= a
+      || a + c <= b
+    ) {
+      throw new Error('The longest side of a triangle must be smaller'
+      + 'than a sum of two others');
```
ivan-kadykalo, line 41: Don't split string with + . Will be hard to work or find it in the future
```suggestion
      throw new Error(
        'The longest side of a triangle must be smaller than a sum of two others'
      );
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,104 @@
...
+  color: Color;
+
+  radius: number;
 
+  constructor(color: Color, radius: number) {
+    this.color = color;
+    this.radius = radius;
+
+    if (this.radius <= 0) {
+      throw new Error('Radius of a circle mush be a positive number');
```
ivan-kadykalo, line 65:
```suggestion
      throw new Error('Radius of a circle must be a positive number');
```

ivan-kadykalo (changes requested):
You are almost there 💪

src/figuresClasses.ts
batstolya: image

ivan-kadykalo (approved):
Good job 🔥

PR #823, opened 2022-12-19
solution

oleksandrnikora:


src/figuresClasses.ts
```diff
@@ -1,19 +1,76 @@
...
 
-export class Circle {
+  getArea(): number {
+    const p = 0.5 * (this.a + this.b + this.c);
 
+    return +(Math.sqrt(p
+      * (p - this.a)
+      * (p - this.b)
+      * (p - this.c)))
+      .toFixed(2);
```
AntonMartynes, line 35: `return Number((Math.sqrt(
      p
      * (p - this.a)
      * (p - this.b)
      * (p - this.c),
    ).toFixed(2)
    ));`
    
    In my opinion, it will be better to use Number instead of '+'

AntonMartynes (approved):

solaryasha (approved):
👍

PR #824, opened 2022-12-19
Initial commit

artem-hirzhev:


src/figuresClasses.ts
```diff
@@ -1,19 +1,77 @@
...
+  ) {
+    if (radius <= 0) {
+      throw new Error('Wrong radius');
+    }
+  }
 
+  getArea(): number {
+    const area = Math.PI * (this.radius ** 2);
+
+    return Math.floor(area * 100) / 100;
```
vlad-demchuk, line 51: don't repeat yourself. use a separate function for `Math.floor`

vlad-demchuk (changes requested):

SanyaBratashchuk (approved):

PR #827, opened 2022-12-19
solution

illiashcherbak:


src/figuresClasses.ts
```diff
@@ -1,19 +1,95 @@
...
+  public shape: Shape = 'triangle';
+
+  isInvalidTriangle(): boolean {
+    return this.a === 0
+      || this.b === 0
+      || this.c === 0
+      || this.a + this.b <= this.c
+      || this.b + this.c <= this.a
+      || this.c + this.a <= this.b;
+  }
```
oleh-yablunovskyi, line 21: With destructurization there will be no need to use 'this' so many times. You can try it.
**const { a, b, c } = this;**

oleh-yablunovskyi (approved):
Good job!

src/figuresClasses.ts
illiashcherbak: Thank you!

src/figuresClasses.ts
```diff
@@ -1,19 +1,98 @@
...
+  color: Color;
 
+  getArea(): number;
 }
 
-export class Triangle {
+export class Triangle implements Figure {
+  public shape: Shape = 'triangle';
+
+  isInvalidTriangle(): boolean {
```
solaryasha, line 14: It is a cool method and I like it. Let's make it private, cause you are using it only internally

src/figuresClasses.ts
```diff
@@ -1,19 +1,98 @@
...
+    const { a, b, c } = this;
+
+    return a === 0
+      || b === 0
+      || c === 0
+      || a + b <= c
+      || b + c <= a
+      || c + a <= b;
```
solaryasha, line 22: Let's store these conditions in the array and check it with the `some` method

solaryasha (changes requested):

src/figuresClasses.ts
```diff
@@ -1,19 +1,101 @@
...
-export class Triangle {
+export class Triangle implements Figure {
+  public shape: Shape = 'triangle';
+
+  private isInvalidTriangle(): boolean {
+    const { a, b, c } = this;
+    const checks = [
+      a === 0,
+      b === 0,
+      c === 0,
```
solaryasha, line 19:
```suggestion
      a <= 0,
      b <= 0,
      c <= 0,
```

solaryasha (approved):

PR #830, opened 2022-12-19
Solution

alyonkaradchuk:


src/figuresClasses.ts
```diff
@@ -1,19 +1,93 @@
...
+  b: number;
 
-export class Circle {
+  c: number;
 
+  constructor(
+    public color: 'red' | 'green' | 'blue',
+    a: number,
+    b: number,
+    c: number,
```
solaryasha, line 20: You should use `public` keyword for adding values to `this` automatically.

src/figuresClasses.ts
```diff
@@ -1,19 +1,93 @@
...
+  ) {
+    if (a > 0 && b > 0 && c > 0) {
+      if (a + b <= c || a + c <= b || c + b <= a) {
```
solaryasha, line 23: Avoid using nested `if` statements

src/figuresClasses.ts
```diff
@@ -1,19 +1,93 @@
...
+
+  radius: number;
 
+  constructor(
+    public color: 'red' | 'green' | 'blue',
+    radius: number,
+  ) {
+    if (radius <= 0) {
+      throw new Error('Not valid data');
+    } else {
```
solaryasha, line 56: You don't need `else` here. The `throw` command stops code execution

solaryasha (changes requested):

Ihor-Karpyn (approved):

PR #831, opened 2022-12-19
solution

DimaKolesnyk:


Mykhailyk-Vadym (approved):
looks good

src/figuresClasses.ts
```diff
@@ -1,19 +1,79 @@
...
+    public c: number,
+  ) {
+    if (this.a <= 0
+    || this.b <= 0
+    || this.c <= 0
+    || this.a >= this.b + this.c
+    || this.b >= this.a + this.c
+    || this.c >= this.a + this.b
+    ) {
+      throw new Error('Wrong input data');
```
ivan-kadykalo, line 24: Checklist p3
How should the user understand what was wrong with his values?
Be more detailed, and write what data are invalid.

ivan-kadykalo (changes requested):
You are almost there 💪

src/figuresClasses.ts
ivan-kadykalo: p.s. In all errors

DimaKolesnyk: Got it, ty

ivan-kadykalo (approved):
Good job 🔥

PR #833, opened 2022-12-19
my math teacher would be proud of me

RedkoSerhii:


Sergii-Osedchuk (approved):
Everything is good!!!

src/figuresClasses.ts
```diff
@@ -1,19 +1,87 @@
...
+  shape: Shape;
 
-}
+  constructor(
+    public color: Color,
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    this.shape = 'triangle';
```
solaryasha, line 19: Why does this value differ from values in type`Shape`?

src/figuresClasses.ts
```diff
@@ -1,19 +1,87 @@
...
 
+export interface Figure {
+  shape: Shape;
+  color: Color;
+  getArea(): number;
 }
 
-export class Triangle {
+export class Triangle implements Figure {
+  shape: Shape;
```
```suggestion
  shape: Shape = 'Triangle';
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,87 @@
...
+    this.shape = 'triangle';
 
-export class Circle {
+    if (
+      this.a <= 0
+      || this.b <= 0
+      || this.c <= 0
+      || this.a + this.b <= this.c
+      || this.b + this.c <= this.a
+      || this.c + this.a <= this.b) {
```
solaryasha, line 27: Let's pack the conditions into an array and check it with `some` method

solaryasha (changes requested):

src/figuresClasses.ts
RedkoSerhii: fixed it

src/figuresClasses.ts
RedkoSerhii: fixed it, at all cases below too

src/figuresClasses.ts
RedkoSerhii: cool advice, i fixed it. But I found desision with metod.some only for <=0. I didnt find option in this metod for condition like this (this.a + this.b <= this.c). So I used metod.sort for this conditions.

src/figuresClasses.ts
```diff
@@ -1,19 +1,93 @@
...
+    if (checkSides[2] >= checkSides[1] + checkSides[0]) {
+      throw new Error('longest side cannot be >= sum of 2 others');
+    }
+    // if (
+    //   this.a <= 0
+    //   || this.b <= 0
+    //   || this.c <= 0
+    //   || this.a + this.b <= this.c
+    //   || this.b + this.c <= this.a
+    //   || this.c + this.a <= this.b) {
```
solaryasha, line 35: You could store these expressions as array elements
```
const conditions = [this.a <= 0, this.a + this.b <= this.c, and so on]
```
If one of them returns true, you should throw an error

src/figuresClasses.ts
```diff
@@ -1,19 +1,93 @@
...
 
-export class Rectangle {
+export class Circle implements Figure {
+  shape: Shape = 'circle';
+
+  constructor(
+    public color: Color,
+    public radius: number,
+  ) {
+    // this.shape = 'circle';
```
solaryasha, line 59: Don't leave the comments

solaryasha (changes requested):

src/figuresClasses.ts
RedkoSerhii: fixed it

src/figuresClasses.ts
RedkoSerhii: fixed it

src/figuresClasses.ts
```diff
@@ -1,19 +1,85 @@
...
+    const conditions = [
+      this.a <= 0,
+      this.b <= 0,
+      this.c <= 0,
+      this.a + this.b <= this.c,
+      this.b + this.c <= this.a,
+    ];
 
-export class Circle {
+    if (conditions.some((condition) => (condition))) {
```
```suggestion
    if (conditions.some(Boolean)) {
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,85 @@
...
+  shape: Shape = 'triangle';
 
-}
+  constructor(
+    public color: Color,
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    const conditions = [
```
solaryasha, line 19:
```suggestion
    const invalidConditions = [
```

src/figuresClasses.ts
solaryasha: No, it's still there

solaryasha (approved):

src/figuresClasses.ts
RedkoSerhii: fixed, I didn`t use this sintaxis early, now I know, thanks

src/figuresClasses.ts
RedkoSerhii: fixed it

PR #834, opened 2022-12-19
solution

Sergii-Osedchuk:


illiashcherbak (approved):
Looks great! Good job!

src/figuresClasses.ts
```diff
@@ -1,19 +1,87 @@
...
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    if (!this.checkTriangle()) {
+      throw new Error('There are no correct triangle sizes');
+    }
+  }
+
+  checkTriangle(): boolean {
```
solaryasha, line 24:
```suggestion
  isValidTriangle(): boolean {
```

solaryasha: It is a cool method and I like it. Let's make it private, cause you are using it only internally

src/figuresClasses.ts
```diff
@@ -1,19 +1,87 @@
...
+    const maxSize = Math.max(this.a, this.b, this.c);
+
+    if ((maxSize < this.a + this.b + this.c - maxSize)
+      && this.a > 0
+      && this.b > 0
+      && this.c > 0) {
```
solaryasha, line 30: Let's store these conditions in the array and check it with the `some` method

solaryasha (changes requested):

src/figuresClasses.ts
Sergii-Osedchuk: Fixed. Its a good way to check conditions. Thanks.

Ihor-Karpyn (approved):

PR #836, opened 2022-12-19
solution try1

vladimir-arkanov:


src/figuresClasses.ts
```diff
@@ -1,19 +1,96 @@
...
+  shape: Shape = 'triangle';
 
-}
+  color: Color = 'red';
+
+  a: number;
+
+  b: number;
+
+  c:number;
```
Anastasia4sik, line 19: Do you need so many spaces?

src/figuresClasses.ts
```diff
@@ -1,19 +1,96 @@
...
+    this.c = c;
 
+    if (a <= 0 || b <= 0 || c <= 0) {
+      throw new Error('ERROR, side length can not be 0 or negative');
+    }
+
+    if (a >= b + c || b >= a + c || c >= a + b) {
+      throw new Error('ERROR, it is not triangle');
+    }
+  }
```
Anastasia4sik, line 34: I think you can combine this into one if with one Error
But it is not necessary

src/figuresClasses.ts
```diff
@@ -1,19 +1,96 @@
-export interface Figure {
+export type Shape = 'triangle'|'circle'|'rectangle';
+export type Color = 'red'|'green'|'blue';
 
+export interface Figure {
+  shape: string;
+  color:string;
+  getArea():number;
```
Anastasia4sik, line 7: You forgot spaces between : and type
Check it in whole document
Check it with linter maybe

src/figuresClasses.ts
```diff
@@ -1,19 +1,96 @@
...
+  }
 }
 
-export class Rectangle {
+export class Circle implements Figure {
+  shape: Shape = 'circle';
+
+  color: Color = 'green';
+
+  radius:number;
```
Anastasia4sik, line 51: Do you need so many spaces?

Anastasia4sik (changes requested):
Good job but few mistakes

src/figuresClasses.ts
vladimir-arkanov: I put two different errors because we have two checks

src/figuresClasses.ts
```diff
@@ -1,19 +1,96 @@
...
+  color: string;
+  getArea(): number;
 }
 
-export class Triangle {
+export class Triangle implements Figure {
+  shape: Shape = 'triangle';
 
-}
+  color: Color = 'red';
```
ivan-kadykalo, line 13: Better to use enums here
`shape = Shape.Triangle`
`color = Color.Red`

ivan-kadykalo (approved):
Good job 🔥

PR #837, opened 2022-12-19
figures classes solution

DiDemkiv:


src/figuresClasses.ts
```diff
@@ -1,19 +1,76 @@
...
+    ) {
+      throw new Error('Wrong sides!');
+    }
+  }
 
-export class Circle {
+  getArea(): number {
+    const halfPerim = (this.a + this.b + this.c) / 2;
+    const area = Math.sqrt(halfPerim * (halfPerim - this.a)
+      * (halfPerim - this.b) * (halfPerim - this.c));
```
batstolya, line 33: you can use destructuring for your values, but this is optional;

```suggestion
		{a, b , c} = this;
    const halfPerim = (a + b + c) / 2;
    const area = Math.sqrt(halfPerim * (halfPerim - a)
      * (halfPerim - b) * (halfPerim - c));
```

batstolya:
rest looks good 🏆

src/figuresClasses.ts
```diff
@@ -1,19 +1,77 @@
...
+    public c: number,
+  ) {
+    if (a <= 0
+      || b <= 0
+      || c <= 0
+      || (a + b) <= c
+      || (c + a) <= b
+      || (c + b) <= a
+    ) {
+      throw new Error('Wrong sides!');
```
ivan-kadykalo, line 26: Checklist p3 
The rest is the same

How users should understand what exactly was wrong and what data are valid?

ivan-kadykalo (changes requested):
You are almost there 💪

ivan-kadykalo (approved):
Good job 🔥

PR #840, opened 2022-12-20
Solution

YanaEvaGonchar:


src/figuresClasses.ts
```diff
@@ -1,19 +1,103 @@
...
 
+export interface Figure {
+  shape: Shape,
+  color: Color,
+  getArea(): number,
 }
 
-export class Triangle {
+export class Triangle implements Figure {
+  shape: Shape;
```
kreisberg, line 11: we can skip `this.shape = 'triangle`
and write it here:
`shape: Shape = 'triangle'`


it's about all types of shapes

src/figuresClasses.ts
```diff
@@ -1,19 +1,103 @@
...
 
-}
+  constructor(
+    public color: Color,
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    this.shape = 'triangle';
```
kreisberg, line 19: OR make enum here:
`this.shape = Shape.Triangle`

src/figuresClasses.ts
YanaEvaGonchar: done

PR #842, opened 2022-12-20
solution

OlexandraKorol:
Figures classes


KrisPlanka (approved):
looks good)

src/figuresClasses.ts
```diff
@@ -1,19 +1,83 @@
...
+
+  constructor(
+    public color: Color,
+    public width: number,
+    public height: number,
+  ) {
+    this.shape = 'rectangle';
+
+    if (this.width <= 0 || this.height <= 0) {
+      throw new Error('Wrong length of side');
```
ivan-kadykalo, line 72: You should write why is it wrong. 
For example `Value can't be less than 1`
The same in all other errors

ivan-kadykalo (changes requested):
You are almost there 💪

ivan-kadykalo (approved):
Good job 🔥

PR #843, opened 2022-12-20
solution

APetrinko:


src/figuresClasses.ts
```diff
@@ -1,19 +1,86 @@
...
 
-export class Triangle {
+export class Triangle implements Figure {
+  shape: Shape = 'triangle';
 
-}
+  isValidTriangle(): boolean {
+    return this.a === 0
+      || this.b === 0
+      || this.c === 0
```
olena-yanovska, line 16: we should check if any length is <= 0, not just = 0

olena-yanovska (changes requested):
looks good to me, just fix the check of triangle sizes

olena-yanovska (approved):

src/figuresClasses.ts
```diff
@@ -1,19 +1,88 @@
...
+  }
 
-export class Circle {
+  constructor(
+    public color: Color,
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    if (this.isValidTriangle()) {
```
Ihor-Karpyn, line 30: Why you throw an error if triangle is valid `if (this.isValidTriangle()) {`

Ihor-Karpyn (changes requested):

src/figuresClasses.ts
```diff
@@ -1,19 +1,83 @@
...
+
+  constructor(
+    public color: Color,
+    public wigth: number,
+    public height: number,
+  ) {
+    const isValidRectangle = wigth > 0 && height > 0;
+
+    if (!isValidRectangle) {
+      throw new Error('The entered sides can\t form a rectangle');
```
ivan-kadykalo, line 72: You should explain what data is valid. Fix here and in other places
```suggestion
      throw new Error('The entered sides can\t form a rectangle. Values should be greater than 0!');
```

ivan-kadykalo (changes requested):
You are almost there 💪

src/figuresClasses.ts
```diff
@@ -1,19 +1,84 @@
...
+  constructor(
+    public color: Color,
+    public wigth: number,
+    public height: number,
+  ) {
+    const isValidRectangle = wigth > 0 && height > 0;
+
+    if (!isValidRectangle) {
+      throw new Error('The entered sides can\t form a rectangle. Values should'
+      + ' be greater than 0!');
```
ivan-kadykalo, line 73: No needs to break text to a couple lines
```suggestion
      throw new Error(
      	'The entered sides can\t form a rectangle. Values should be greater than 0!'
      );
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,84 @@
...
-export class Rectangle {
+export class Circle implements Figure {
+  shape: Shape = 'circle';
+
+  constructor(
+    public color: Color,
+    public radius: number,
+  ) {
+    if (radius <= 0) {
+      throw new Error('The entered radius can\'t form a circle');
```
ivan-kadykalo, line 50: You need to describe here the same as at the bottom. Why this radius can't form a circle? Fix it in all places!!!

src/figuresClasses.ts
```diff
@@ -1,19 +1,84 @@
...
+  ) {
+    if (this.a <= 0
+      || this.b <= 0
+      || this.c <= 0
+      || this.a + this.b <= this.c
+      || this.b + this.c <= this.a
+      || this.a + this.c <= this.b
+    ) {
+      throw new Error(
+        `sides ${this.a}, ${this.b} and ${this.c} can't form a triangle`,
```
ivan-kadykalo, line 27: Here the same
image

ivan-kadykalo (changes requested):

PR #845, opened 2022-12-20
added shape classes, figure interface, color and shape types, toFixed…

TheOneThatKnowsNothing:
… method

### DIFF
--- a/src/figuresClasses.ts
+++ b/src/figuresClasses.ts
@@ -1,19 +1,104 @@
+type Shape = 'triangle' | 'circle' | 'rectangle';
+type Color = 'red' | 'green' | 'blue';
+
+function toFixed(num: number, fixed: number): number {
+  if (fixed <= 0) {
+    return +num.toString();
+  }
+
+  const numToStr = num.toString();
+  const pointIndex = numToStr.split('').indexOf('.');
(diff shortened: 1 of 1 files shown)


src/figuresClasses.ts
```diff
@@ -1,19 +1,104 @@
-export interface Figure {
+type Shape = 'triangle' | 'circle' | 'rectangle';
+type Color = 'red' | 'green' | 'blue';
+
+function toFixed(num: number, fixed: number): number {
+  if (fixed <= 0) {
+    return +num.toString();
```
solaryasha, line 6:
```suggestion
    return Number(num.toString());
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,104 @@
...
+
+  const numToStr = num.toString();
+  const pointIndex = numToStr.split('').indexOf('.');
+  let result = numToStr;
 
+  if (pointIndex > 0) {
+    result = numToStr.slice(0, pointIndex + 1 + fixed);
+  }
+
+  return +result;
```
solaryasha, line 17:
```suggestion
  return Number(result);
```

solaryasha (approved):
👍

PR #847, opened 2022-12-20
Solution

novytskyi-d:


src/figuresClasses.ts
```diff
@@ -1,19 +1,98 @@
...
 
+export interface Figure {
+  shape: Shape;
+  color: Color;
+  getArea(): number;
 }
 
-export class Triangle {
+export class Triangle implements Figure {
+  shape: Shape;
```
```suggestion
  shape: Shape = 'triangle';
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,98 @@
...
-}
+  constructor(
+    public color: Color,
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    this.shape = 'triangle';
+
+    if (this.isValidTriangle()) {
```
solaryasha, line 21: It is a cool method and I like it. Let's make it private, cause you are using it only internally

src/figuresClasses.ts
```diff
@@ -1,19 +1,98 @@
...
-export class Circle {
+  isValidTriangle(): boolean {
+    return (
+      this.a <= 0
+      || this.b <= 0
+      || this.c <= 0
+      || this.a + this.b <= this.c
+      || this.a + this.c <= this.b
+      || this.b + this.c <= this.a
+    );
```
solaryasha, line 34: Let's store these conditions in the array and check it with the some method

src/figuresClasses.ts
```diff
@@ -1,19 +1,98 @@
...
+  ) {
+    this.shape = 'triangle';
+
+    if (this.isValidTriangle()) {
+      throw new Error('Something went wrong with the sides of the triangle.');
+    }
+  }
 
-export class Circle {
+  isValidTriangle(): boolean {
```
solaryasha, line 26: If the method has a name isValidTriangle - we should return true if it's valid

solaryasha (changes requested):

PR #849, opened 2022-12-20
solution

kreisberg:


src/figuresClasses.ts
```diff
@@ -1,19 +1,100 @@
...
+  shape: Shape;
 
-}
+  constructor(
+    public color: Color,
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    this.shape = 'triangle';
```
RedkoSerhii, line 20: you can unit   
this.shape = 'triangle' 
and 
shape: Shape; 
Do it at all casses.

RedkoSerhii (approved):

src/figuresClasses.ts
```diff
@@ -1,19 +1,100 @@
...
+  }
+
+  isTriangle(): boolean {
+    return (
+      this.a + this.b > this.c
+        && this.a + this.c > this.b
+        && this.b + this.c > this.a
+        && this.a !== 0
+        && this.b !== 0
+        && this.c !== 0
```
ivan-kadykalo, line 35: Keep indentation
```suggestion
      && this.a + this.c > this.b
      && this.b + this.c > this.a
      && this.a !== 0
      && this.b !== 0
      && this.c !== 0
```

src/figuresClasses.ts
ivan-kadykalo, line 20: Make enum for shapes
`this.shape = Shape.Triangle`

src/figuresClasses.ts
```diff
@@ -1,19 +1,100 @@
...
+  shape: Shape;
+
+  constructor(
+    public color: Color,
+    public radius: number,
+  ) {
+    this.shape = 'circle';
+
+    if (radius <= 0) {
+      throw new Error('Is not a circle');
```
ivan-kadykalo, line 61: Checklist p3
How use should understand what is wrong and what data are valid?
Fix here and in all the rest errors

ivan-kadykalo (changes requested):
You are almost there 💪

src/figuresClasses.ts
kreisberg: ✔️

src/figuresClasses.ts
kreisberg: ✔️

src/figuresClasses.ts
kreisberg: cut without using 'this':
`  shape: Shape = 'triangle'; `

src/figuresClasses.ts
```diff
@@ -1,19 +1,94 @@
...
+export class Rectangle implements Figure {
+  shape: Shape = 'rectangle';
+
+  constructor(
+    public color: Color,
+    public width: number,
+    public height: number,
+  ) {
+    if (this.isRectangle() === true) {
+      throw new Error('Uncorrect rectangle sides size');
```
ivan-kadykalo, line 77: Incorrect, but which is correct?
For example
`Values should be greater than 0`
You should explain in the errors what exactly are incorrect

Fix in other places

ivan-kadykalo (changes requested):
You are almost there 💪

src/figuresClasses.ts
kreisberg: ✔️

ivan-kadykalo (approved):
Good job 🔥

PR #850, opened 2022-12-20
Solution

CharAnagn:


src/figuresClasses.ts
```diff
@@ -1,19 +1,87 @@
...
+    public color: Color,
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    if ((a && b && c) <= 0) {
+      throw new Error('All sides must be higher than zero');
+    }
+
+    if (a + b <= c || a + c <= b || b + c <= a) {
```
lefteris-L, line 34: A variable here would make it even clearer what the check is like `failsTriangleInequality` or something like that, but the message probably covers it

lefteris-L (approved):
LGTM 💯 just a minor suggestion

meforoma (approved):
✅

PR #853, opened 2022-12-20
done1

panianchuk:


src/figuresClasses.ts
```diff
@@ -1,19 +1,76 @@
...
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    if (this.a <= 0 || this.b <= 0 || this.c <= 0
+      || this.a >= (this.b + this.c)
+      || this.b >= (this.a + this.c)
+      || this.c >= (this.b + this.a)) {
+      throw new Error('check the sides of the triangle!');
+    }
```
DarkMistyRoom, line 24: You check that input values are bigger than 0 for each class
Will be better to create separate function, that will accept initial parameters, check it and return the corresponding error if not

DarkMistyRoom (changes requested):

nazarmatsevych (approved):
👍

PR #854, opened 2022-12-21
Solution

skorykdariia:


src/figuresClasses.ts
```diff
@@ -1,19 +1,99 @@
...
+    public color: Color,
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    this.shape = 'triangle';
+
+    if (this.isValidTriangle()) {
+      throw new Error('Impossible to make a triangle with such sides');
+    }
```
oleh-yablunovskyi, line 23:
```suggestion
    if (this.isInvalidTriangle()) {
      throw new Error('Impossible to make a triangle with such sides');
    }
```
I would rather rename the method name. Because here it sounds like "If this is valid triangle, I will throw Error". Which is kinda confusing in terms of human language. The same in other methods.

oleh-yablunovskyi (approved):
Good job!

src/figuresClasses.ts
solaryasha: I agree with Oleh, let's fix it.

src/figuresClasses.ts
```diff
@@ -1,19 +1,99 @@
...
+  }
 
-export class Circle {
+  isValidTriangle(): boolean {
```
solaryasha, line 26: It is a cool method and I like it. Let's make it private, cause you are using it only internally

src/figuresClasses.ts
```diff
@@ -1,19 +1,99 @@
...
+    return (
+      this.a <= 0
+      || this.b <= 0
+      || this.c <= 0
+      || this.a + this.b <= this.c
+      || this.a + this.c <= this.b
+      || this.b + this.c <= this.a
+    );
```
solaryasha, line 34: Let's store these conditions in the array and check it with the `some` method

src/figuresClasses.ts
```diff
@@ -1,19 +1,99 @@
...
 
+export interface Figure {
+  shape: Shape;
+  color: Color;
+  getArea(): number;
 }
 
-export class Triangle {
+export class Triangle implements Figure {
+  shape: Shape;
```
solaryasha, line 11: We don't need to set this value in the constructor
```suggestion
  shape: Shape = 'triangle';
```

solaryasha (changes requested):

solaryasha (changes requested):
.

src/figuresClasses.ts
```diff
@@ -1,19 +1,96 @@
...
+  ) {
+    if (this.isInvalidTriangle()) {
+      throw new Error('Impossible to make a triangle with such sides');
+    }
+  }
 
-export class Circle {
+  private isInvalidTriangle(): boolean {
+    const { a, b, c } = this;
+    const inspect = [
```
solaryasha, line 26:
```suggestion
    const checks = [
```

solaryasha (approved):

PR #855, opened 2022-12-21
Solution

YakovlewDan:


NadiiaKoch (approved):
Nice!

src/figuresClasses.ts
```diff
@@ -1,19 +1,99 @@
...
+  color: Color;
+  getArea(): number;
 }
 
-export class Triangle {
+export class Triangle implements Figure {
+  public shape: Shape = 'triangle';
 
-}
+  isValidTriangle(): boolean {
```
solaryasha, line 13: It is a cool method and I like it. Let's make it private, cause you are using it only internally

src/figuresClasses.ts
```diff
@@ -1,19 +1,99 @@
...
+    return this.a === 0
+      || this.b === 0
+      || this.c === 0
+      || this.a + this.b <= this.c
+      || this.b + this.c <= this.a
+      || this.a + this.c <= this.b;
```
solaryasha, line 19: Let's store these conditions in the array and check it with the some method

src/figuresClasses.ts
solaryasha, line 13: If the method has a name `isValidTriangle`- we should return true if it's valid

solaryasha (changes requested):

PR #856, opened 2022-12-21
solution

grslckr:


src/figuresClasses.ts
```diff
@@ -1,19 +1,72 @@
...
+  shape: Figure['shape'] = 'triangle';
 
-}
+  constructor(
+    public color: Figure['color'],
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    if (Math.min(...[a, b, c]) <= 0
```
UmizDemud, line 18: I suppose you can omit the square brackets, like:
```js
Math.min(a, b, c)
```

UmizDemud (approved):
well done 👍

src/figuresClasses.ts
```diff
@@ -1,19 +1,72 @@
-export interface Figure {
+const hackerAlert = new Error('Congratz! You\'ve hacked the system!');
 
+export interface Figure {
+  shape: 'triangle' | 'circle' | 'rectangle',
+  color: 'red' | 'green' | 'blue',
+  getArea: () => number,
 }
```
meforoma, line 7: relevant interface, but i still think you should use enums for shape and color
```suggestion
enum Shape {
  triangle,
  circle,
  rectangle,
}

enum Color {
  red,
  green,
  blue,
}

export interface Figure {
  shape: Shape,
  color: Color,
  getArea: () => number,
}
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,72 @@
...
 
-export class Triangle {
+export class Triangle implements Figure {
+  shape: Figure['shape'] = 'triangle';
```
meforoma, line 10: and then this would omit using strings - which are more prone to mistyping compared to enums
```suggestion
  shape: Figure.shape = Shape.triangle;
```

meforoma (changes requested):
let's use enums and consider @UmizDemud's comments - indeed `Math.max(a,b,c)` would do the job

meforoma (approved):
✅

PR #858, opened 2022-12-21
Solution

sercansimsek:


src/figuresClasses.ts
```diff
@@ -1,19 +1,76 @@
-export interface Figure {
+type Shape = 'triangle' | 'circle' | 'rectangle';
+type Color = 'red' | 'green' | 'blue';
```
Iksich, line 3: Use Enums, it is good to practice, and even in this code you can use them to make your code better.

src/figuresClasses.ts
```diff
@@ -1,19 +1,76 @@
...
 
+export interface Figure {
+  shape: Shape;
+  color: Color;
+  getArea(): number;
 }
 
-export class Triangle {
+export class Triangle implements Figure {
+  shape: Shape = 'triangle';
```
Iksich, line 11: Assign shape directly shape = Shape.Triangle it will shorten your code.

Iksich (changes requested):

meforoma (approved):
✅

PR #860, opened 2022-12-21
Shapes and Figures

Kavster92:


src/figuresClasses.ts
```diff
@@ -1,19 +1,81 @@
-export interface Figure {
+type Shape = 'triangle' | 'circle' | 'rectangle';
+type Color = 'red' | 'green' | 'blue';
```
JzSol, line 3: Would be better to use an enum for Color and Shape.

JzSol (approved):
Looks good, but check out the comments!

src/figuresClasses.ts
```diff
@@ -1,19 +1,81 @@
...
+    const triangleArea = Math.sqrt(halfOfPerimeter * (halfOfPerimeter - this.a)
+    * (halfOfPerimeter - this.b) * (halfOfPerimeter - this.c));
 
+    return Math.round(triangleArea * 100) / 100;
+  }
 }
 
-export class Rectangle {
+export class Circle implements Figure {
+  shape: Shape = 'circle';
```
meforoma, line 41: totally agree for the comment about using an enum
that's a common practice to avoid strings as much as possible
```suggestion
  shape: Shape = Shape.Triangle;
```

src/figuresClasses.ts
meforoma: totally. Convert to enum and adjust using enum in the following code.

meforoma (changes requested):
it's decent - just add an enum and you're all good.

naziamoff (approved):
well done!

PR #862, opened 2022-12-21
Solution

Grayfox147:


src/figuresClasses.ts
```diff
@@ -1,19 +1,80 @@
-export interface Figure {
+type Color = 'red' | 'green' | 'blue';
+type Shape = 'triangle' | 'circle' | 'rectangle';
```
Iksich, line 3: You don't need it in this task, but maybe better to use enum instead of just types.

src/figuresClasses.ts
```diff
@@ -1,19 +1,80 @@
...
+  shape: Shape;
 
-}
+  constructor(
+    public color: Color,
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    this.shape = 'triangle';
```
Iksich, line 19: They told me to change this on my review cause you can assign directly shape = Shape.Triangle, and with enums you can also skip the type after you implement them.

Iksich (approved):

src/figuresClasses.ts
meforoma: introduce enum

src/figuresClasses.ts
meforoma, line 19: use enum

meforoma (changes requested):
@Grayfox147 don't ignore peer comments.
introduce enum and use it - it's a common practice in production development.

meforoma (approved):
✅

PR #867, opened 2022-12-21
solution

mmahfuzi:


UmizDemud (approved):
🚀 good job

src/figuresClasses.ts
```diff
@@ -1,19 +1,64 @@
-export interface Figure {
 
+export interface Figure {
+  shape: string;
```
meforoma, line 3: let's use an enum for both shape and color
like
```javascript
enum Shape {
  triangle,
  rectangle,
  circle,
}
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,64 @@
...
+  color: string;
+  getArea(): number;
 }
 
-export class Triangle {
+export class Triangle implements Figure {
+  shape = 'triangle';
```
meforoma, line 9: use enum
```suggestion
  shape = Shape.triangle;
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,64 @@
...
+  }
 
+  getArea(): number {
+    return Math.floor((Math.PI * this.radius ** 2) * 100) / 100;
+  }
 }
 
-export function getInfo(figure) {
+export class Rectangle implements Figure {
+  shape = 'rectangle';
```
meforoma, line 45: enum

meforoma (changes requested):
let's use enums @mmahfuzi

meforoma (approved):
👏

PR #869, opened 2022-12-22
Solution

Rijo-J:


src/figuresClasses.ts
```diff
@@ -1,19 +1,85 @@
...
+    const sides = [this.side1, this.side2, this.side3]
+      .sort((a, b) => a - b);
+    const [cat1, cat2, hip] = sides;
 
-export class Circle {
+    return cat1 + cat2 <= hip;
+  }
 
+  getArea(): number {
+    const sides = [this.side1, this.side2, this.side3].sort();
```
UmizDemud, line 37: You don't have to sort this. Semiperimeter will be greater than any of the sides.

For any a, b, c that forms a triangle

a + b > c,
a + b + c > 2c
(a + b + c) / 2 > c
semiperimeter > c
semiperimeter > b
semiperimeter > a

cus same true for 
a + c > b,
b + c > a,

UmizDemud (approved):
👍

src/figuresClasses.ts
```diff
@@ -1,19 +1,85 @@
-export interface Figure {
+type Shape = 'triangle' | 'circle' | 'rectangle';
```
meforoma, line 1: let's better use an enum here

src/figuresClasses.ts
```diff
@@ -1,19 +1,85 @@
...
 
+export interface Figure {
+  shape: Shape;
+  color: Color;
+  getArea(): number;
 }
 
-export class Triangle {
+export class Triangle implements Figure {
+  shape: Shape = 'triangle';
```
meforoma, line 11: and use Shape enum where applicable

meforoma (changes requested):
overall good
but let's use an enum for Shapes, ok?

src/figuresClasses.ts
Rijo-J: ✔️

src/figuresClasses.ts
Rijo-J: ✔️

src/figuresClasses.ts
Rijo-J: ✔️

meforoma (approved):
✅

PR #871, opened 2022-12-22
Solution

RomeoKs:


UmizDemud (approved):
I think multiplying with 0.5 is faster than dividing by 2. I like that you used that. All around good 👍 

Tho please write a sentence to Errors

meforoma (approved):
👍

PR #876, opened 2022-12-25
Solution

Dmytro-Kravchenko:


src/figuresClasses.ts
```diff
@@ -1,19 +1,84 @@
...
+  shape = Shape.Triangle;
+
+  constructor(
+    public color: Color,
+    public side1: number,
+    public side2: number,
+    public side3: number,
+  ) {
+    if (side1 <= 0 || side2 <= 0 || side2 <= 0) {
+      throw new Error('your error message');
```
vlad-demchuk, line 26: specify type of error instead of `your error message`

vlad-demchuk: fix all cases

src/figuresClasses.ts
```diff
@@ -1,19 +1,84 @@
...
+    }
+
+    const sides: number[] = [this.side1, this.side2, this.side3];
+
+    const max: number = Math.max(...sides);
+
+    const sum = sides.filter((a) => a !== max).reduce((a, b) => a + b);
```
vlad-demchuk, line 33: simplify it. you can avoid so many actions. just sort it and use indexes for conditional

src/figuresClasses.ts
```diff
@@ -1,19 +1,84 @@
...
 
+    if (sum <= max) {
+      throw new Error('your error message');
+    }
+  }
+
+  getArea: GetArea = () => {
```
vlad-demchuk, line 40:
```suggestion
  getArea(): number  {
```

vlad-demchuk (changes requested):

tetiana-ivanchenko (approved):

PR #878, opened 2022-12-26
solution

tYuriy:


src/figuresClasses.ts
```diff
@@ -1,19 +1,84 @@
...
 }
 
-export class Triangle {
+export class Triangle implements Figure {
+  constructor(
+    public color: Colors,
+    public a: number,
+    public b: number,
+    public c: number,
+    public shape: Shapes = 'triangle',
```
TarasHoliuk, line 20: it shouldn't be in the `constructor`. Move it above the constructor

src/figuresClasses.ts
```diff
@@ -1,19 +1,84 @@
-export interface Figure {
+type Colors = 'red' | 'green' | 'blue';
+type Shapes = 'triangle' | 'circle' | 'rectangle';
```
TarasHoliuk, line 2:
```suggestion
type Shape = 'triangle' | 'circle' | 'rectangle';
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,84 @@
...
+  getArea(): number {
+    const s = (this.a + this.b + this.c) / 2;
+
+    return roundNumber(
+      Math.sqrt(s * (s - this.a) * (s - this.b) * (s - this.c)),
+    );
+  }
 }
 
 export class Circle {
```
TarasHoliuk, line 44: It also implements `Figure` (as well as `Rectangle`)

src/figuresClasses.ts
TarasHoliuk: the same in other figures

TarasHoliuk (changes requested):

nazarmatsevych (approved):

PR #879, opened 2022-12-28
Figures classes

maksym-kryvolap:


src/figuresClasses.ts
```diff
@@ -1,19 +1,85 @@
 export interface Figure {
-
+  shape: string;
+  color: string;
+  getArea(): number;
 }
 
-export class Triangle {
+export class Triangle implements Figure {
+  shape: string;
```
TarasHoliuk, line 8: you can define it here in a such way:

```suggestion
  public shape: string = 'triangle';
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,85 @@
 export interface Figure {
-
+  shape: string;
+  color: string;
```
TarasHoliuk, line 3: shape and color have defined possible values. let's create types for them (enum or union)

src/figuresClasses.ts
```diff
@@ -1,19 +1,85 @@
...
+
+    if (this.a <= 0 || this.b <= 0) {
+      throw new Error("Can't form a Rectangle");
+    }
+  }
+
+  getArea(): number {
+    const square = this.a * this.b;
+
+    return square;
```
TarasHoliuk, line 79:
```suggestion
    return this.a * this.b;
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,85 @@
...
+
+    if (this.a <= 0) {
+      throw new Error("Can't form a Circle");
+    }
+  }
 
+  getArea(): number {
+    const square = 100 * (Math.PI * (this.a * this.a));
+
+    return Math.floor(+square) / 100;
```
TarasHoliuk, line 57:
```suggestion
    return Math.floor(square) / 100;
```
it's always number

src/figuresClasses.ts
```diff
@@ -1,19 +1,85 @@
...
+    }
 
-export class Circle {
+    if (this.c >= this.b + this.a) {
+      throw new Error("sides 1, 2 and 3 can't form a triangle");
+    }
+  }
 
+  getArea(): number {
+    const s = 0.5 * (this.a + this.b + this.c);
```
TarasHoliuk, line 32: `s` is not the best name

TarasHoliuk (changes requested):

TarasHoliuk (approved):
:+1:

PR #880, opened 2022-12-29
solution

miamarichka:


src/figuresClasses.ts
```diff
@@ -1,19 +1,85 @@
...
+  ) {
+    this.shape = 'triangle';
+
+    if (a <= 0
+            || b <= 0
+            || c <= 0
+            || a + b <= c
+            || b + c <= a
+            || c + a <= b
+    ) {
```
loralevitska, line 25: combine logic and move to variable. Ex. `const isValidValue =  a > 0 && b > 0 && c > 0`. All values must be greater than zero, not just some one, right?

src/figuresClasses.ts
```diff
@@ -1,19 +1,85 @@
...
+      throw new Error('Invalid data.Value can\'t be less than 1.');
+    }
+  }
+
+  getArea():number {
+    const sum = (this.a + this.b + this.c) / 2;
+    const area = Math.sqrt(sum * (sum - this.a)
+          * (sum - this.b) * (sum - this.c));
```
loralevitska, line 33: move calculation to variable

src/figuresClasses.ts
```diff
@@ -1,19 +1,85 @@
...
+    this.shape = 'triangle';
+
+    if (a <= 0
+            || b <= 0
+            || c <= 0
+            || a + b <= c
+            || b + c <= a
+            || c + a <= b
+    ) {
+      throw new Error('Invalid data.Value can\'t be less than 1.');
```
loralevitska, line 26: repeatable message move to constants

loralevitska (changes requested):

src/figuresClasses.ts
miamarichka: || a + b <= c
|| b + c <= a
|| c + a <= b
this lasts cant make as an variable, test wont pass than, I tried it didnt work to me. First 3 alright

src/figuresClasses.ts
miamarichka: done

src/figuresClasses.ts
miamarichka: done

src/figuresClasses.ts
```diff
@@ -1,19 +1,87 @@
...
+  ) {
+    this.shape = 'triangle';
+
+    const isInvalidValue:boolean = this.a <= 0 && this.b <= 0 && this.c <= 0;
+
+    if (isInvalidValue || this.a + this.b <= this.c
+            || this.b + this.c <= this.a
+            || this.c + this.a <= this.b) {
+      throw new Error(anError);
+    }
```
SanyaBratashchuk, line 28:
```suggestion
    const sides = [a, b, c];
    const biggestSide = Math.max(...sides);
    const perimeter = sides.reduce((sum, side) => sum + side, 0);
    const hasTooShortSide = sides.some(side <= 0);
    const hasTooLongSide = biggestSide >= perimetr - biggestSide;
    
    if (hasTooShortSide || hasTooLongSide) {
       throw new Error('Invalid value')
    }
    
```

SanyaBratashchuk: In my opinion it is more readable

src/figuresClasses.ts
```diff
@@ -1,19 +1,87 @@
...
+  }
+
+  getArea():number {
+    const sum = (this.a + this.b + this.c) / 2;
```
SanyaBratashchuk, line 32:
```suggestion
    const semiperimeter = (this.a + this.b + this.c) / 2;
```

SanyaBratashchuk (changes requested):

src/figuresClasses.ts
```diff
@@ -1,19 +1,94 @@
...
+  constructor(
+    public color: Figure.color,
+    public width: number,
+    public height: number,
+  ) {
+    this.shape = 'rectangle';
 
-export function getInfo(figure) {
+    if (this.height <= 0
+            || this.width <= 0) {
```
loralevitska, line 80:
```suggestion
    if (this.height <= 0 || this.width <= 0) {
```
OR

```suggestion
    if (this.height <= 0 
         || this.width <= 0) {
```

loralevitska (approved):

PR #883, opened 2023-01-05
solution

euFarafonov:


src/figuresClasses.ts
```diff
@@ -1,19 +1,84 @@
...
 
+function getTruncNum(number: number): number {
+  return Math.trunc(number * 100) / 100;
 }
 
 export class Triangle {
+  public shape: Figure['shape'] = 'triangle';
+
+  constructor(
+    public color: Figure,
```
witflash, line 15: It should be the `Figure['color']` as I see

src/figuresClasses.ts
```diff
@@ -1,19 +1,84 @@
...
 
+    return getTruncNum(area);
+  }
 }
 
 export class Circle {
+  public shape: Figure['shape'] = 'circle';
 
+  constructor(
+    public color: Figure,
```
witflash, line 47: And here too

witflash (changes requested):
Well done!
Pay attention to the comments, please.

src/figuresClasses.ts
witflash: Which type for color do you have?
The same as you use type for shape.
'Figure' is the whole Interface but you need only color here.

euFarafonov:
code fixed

OlEzhA0 (approved):
GJ!

PR #890, opened 2023-01-10
Develop

kostasburnazaki:
On line 78 due to wrong rounding (to nearest hundredth) by JavaScript (instead of .09 it always rounds to .1) I have rounded to the nearest thousandth and chop off last digit.


src/figuresClasses.ts
```diff
@@ -1,19 +1,110 @@
...
+      if (side <= 0) {
+        throw new Error('One of the sides is too short');
+      }
+    });
 
+    this.sumOfSides = [...args].reduce((partialSum, a) => partialSum + a, 0);
+
+    if (this.longestSide >= (this.sumOfSides - this.longestSide)) {
+      throw new Error('Triangle proportion is incorrect');
+    }
```
DarkMistyRoom, line 47: Will be better to create separate function, that will accept initial parameters, check it and return the corresponding error if not

DarkMistyRoom (changes requested):

marina-tilniak (changes requested):

PR #892, opened 2023-01-10
Solution

FocusHuK:


src/figuresClasses.ts
```diff
@@ -1,19 +1,97 @@
...
+      throw new Error('you error message');
+    }
+
+    if (b <= 0) {
+      throw new Error('you error message');
+    }
+
+    if (c <= 0) {
+      throw new Error('you error message');
+    }
```
Anastasiia-Svintsova, line 35: You do the same action in 3 cases, so combine them

src/figuresClasses.ts
```diff
@@ -1,19 +1,97 @@
...
+  constructor(
+    public color: Color,
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    this.shape = 'triangle';
+
+    if (a <= 0) {
+      throw new Error('you error message');
```
Anastasiia-Svintsova, line 26: Change `'you error message'` to the right message everywhere, please

src/figuresClasses.ts
```diff
@@ -1,19 +1,97 @@
...
+
+export interface Figure {
+  shape: Shape;
+  color: Color;
+  getArea(): number;
 }
 
-export class Triangle {
+let p: number;
+let square: number;
```
Anastasiia-Svintsova, line 12: You don't need these variables in a global scope. You have different formulas for every figure, so use const inside getArea() methods

src/figuresClasses.ts
```diff
@@ -1,19 +1,97 @@
...
+  shape: Shape;
+
+  constructor(
+    public color: Color,
+    public width : number,
+    public height : number,
+  ) {
+    this.shape = 'rectangle';
+
+    if (width <= 0 || height <= 0) {
```
Anastasiia-Svintsova, line 83:
```suggestion
    if (!width || !height) {
```

Anastasiia-Svintsova (changes requested):

PR #899, opened 2023-01-16
Add figures and respective functionality

mykyta-krainik:


src/figuresClasses.ts
```diff
@@ -1,19 +1,115 @@
...
+  getArea(): number;
 }
 
-export class Circle {
+export class Triangle implements Figure {
+  readonly shape = 'triangle';
+
+  color: Color;
+
+  readonly sides: [number, number, number];
```
loralevitska, line 21: Combine `[number, number, number]` to avoid duplicate

src/figuresClasses.ts
```diff
@@ -1,19 +1,115 @@
...
 
-export function getInfo(figure) {
+export class Rectangle implements Figure {
+  readonly shape = 'rectangle';
+
+  color: Color;
+
+  width: number;
+
+  height: number;
```
loralevitska, line 94:
```suggestion
  color: Color;
  width: number;
  height: number;
```

loralevitska (changes requested):
Almost done)

src/figuresClasses.ts
```diff
@@ -1,19 +1,113 @@
-export interface Figure {
+/* eslint-disable @typescript-eslint/lines-between-class-members */
```
Anastasiia-Svintsova, line 1: Don't disable eslint. Try to fix the error

Anastasiia-Svintsova (approved):

PR #900, opened 2023-01-16
Solution

andrieieva-yuliia:


OlhaYevstifieieva (approved):
Iteration  methods are always a great solution!

src/figuresClasses.ts
```diff
@@ -1,19 +1,93 @@
...
+  getArea(): number;
+}
 
+function roundedArea(number: number): number {
+  return Math.floor(number * 100) / 100;
 }
 
-export class Triangle {
+export class Triangle implements Figure {
+  shape: Shape = 'triangle';
```
Anna-Tkachenko, line 15: will be better to create enum for corresponding values, because for now you can make error while specifying the string

src/figuresClasses.ts
```diff
@@ -1,19 +1,93 @@
...
 
-}
+  constructor(
+    public color: Color,
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    const triangleSides = [a, b, c];
+    const checkValidSides = triangleSides.every((side) => side > 0);
```
Anna-Tkachenko, line 24: probably we need to check for all figures that initial input arguments are bigger than 0
let's create separate function for this functionality, that will accept params that we'll pass to the figure constructor and will check are they bigger than 0 or no
and will throw an error if no

with this flow you'll be able to call one function for each figure and be sure that input params will be checked, and you don't need to provide separate functionality for this purpose for each class

src/figuresClasses.ts
```diff
@@ -1,19 +1,93 @@
...
+
+    if (longestSide >= sumOfShortestSides) {
+      throw new Error('Length of biggest side cannot '
+      + 'be bigger than sum of two oher sides.');
+    }
+  }
 
-export class Circle {
+  getArea(): number {
+    const s = (this.a + this.b + this.c) / 2;
```
Anna-Tkachenko, line 43: what is `s`?

Anna-Tkachenko (changes requested):

src/figuresClasses.ts
```diff
@@ -1,19 +1,105 @@
...
+  getArea(): number;
+}
 
+function getRoundedArea(area: number): number {
+  return Math.floor(area * 100) / 100;
 }
 
-export class Circle {
+function checkValidSides([...args]: number[]): boolean {
+  return args.every((side) => side > 0);
```
Anna-Tkachenko, line 24: 🔥

Anna-Tkachenko (approved):
Good job!

PR #901, opened 2023-01-17
figureClasses

Kukharuk:


src/figuresClasses.ts
```diff
@@ -1,19 +1,85 @@
...
+  ) {
+    if (a <= 0 || b <= 0 || c <= 0) {
+      throw new Error('Side length below 0');
+    }
+
+    const arr = [];
+
+    arr.push(a);
+    arr.push(b);
+    arr.push(c);
```
vlad-demchuk, line 27:
```suggestion
    const arr = [a, b, c];
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,85 @@
...
+
+    arr.sort((x, y) => y - x);
```
vlad-demchuk, line 29: use better variable names, like `side` here

src/figuresClasses.ts
```diff
@@ -1,19 +1,85 @@
...
+  ) {
+    if (radius <= 0) {
+      throw new Error('Radius length below 0');
+    }
+  }
+
+  getArea():number {
+    const square = Math.PI * this.radius * this.radius;
 
+    return Math.floor(square * 100) / 100;
```
vlad-demchuk, line 61: don't repeat yourself
use a separate function for rounding

vlad-demchuk (changes requested):

witflash (approved):
Well done 👍

PR #902, opened 2023-01-17
solution

Avichez:


oleh-holovnykh (approved):

src/figuresClasses.ts
```diff
@@ -1,19 +1,82 @@
...
+  ) {
+    const sides = [a, b, c];
+    const negativeLength = sides.some((side) => side <= 0);
+    const sumOfSides = sides.reduce((sum, side) => sum + side, 0);
+    const biggestSide = Math.max(...sides);
+    const isBiggest = biggestSide >= sumOfSides - biggestSide;
+
+    if (negativeLength) {
+      throw new Error('One of the sides is negative');
+    }
```
Anna-Tkachenko, line 27: you need to check that input arguments are bigger than 0 for each figure
instead of checking it in each class, let's create one separate function for this purpose, and let's call it for each figure

Anna-Tkachenko (approved):

PR #906, opened 2023-01-18
Figures - completed

GarryRocksman:
Have a nice day🔥

### DIFF
--- a/src/figuresClasses.ts
+++ b/src/figuresClasses.ts
@@ -1,19 +1,124 @@
+type Shape = 'triangle' | 'circle' | 'rectangle';
+type Color = 'red' | 'green' | 'blue';
+
+function isPositive(...args: number[]): boolean {
+  for (let i = 0; i < args.length; i += 1) {
+    if (args[i] < 0) {
+      return false;
+    }
+  }
+
+  return true;
+}
+
+function triangleSideChecker(...args: number[]): boolean {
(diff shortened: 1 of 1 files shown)


src/figuresClasses.ts
```diff
@@ -1,19 +1,124 @@
...
+
+  public b: number;
+
+  public c: number;
+
+  constructor(
+    color: Color,
+    a: number,
+    b: number,
+    c: number,
```
oleh-holovnykh, line 47: you can add public in constructor and delete it from shape

```suggestion
  shape: Shape = 'triangle';
  
  constructor(
    public color: Color,
    public a: number,
    public b: number,
    public c: number,
```

oleh-holovnykh (changes requested):
Good job! And you can move public to constructor and delete sum lines of code

src/figuresClasses.ts
```diff
@@ -1,19 +1,94 @@
-export interface Figure {
+type Shape = 'triangle' | 'circle' | 'rectangle';
+type Color = 'red' | 'green' | 'blue';
+
+function isPositive(...args: number[]): boolean {
+  for (let i = 0; i < args.length; i += 1) {
+    if (args[i] < 0) {
+      return false;
+    }
+  }
```
Anna-Tkachenko, line 9: will be better to use iteration method instead of for cycle

Anna-Tkachenko (approved):
⭐️

PR #907, opened 2023-01-18
Figure classes for Triangle, Circle and Rectangle was added

svitjojo:


src/figuresClasses.ts
```diff
@@ -1,19 +1,95 @@
-export interface Figure {
-
+enum Shape {
+  triangle = 'triangle',
+  circle = 'circle',
+  rectangle = 'rectangle',
 }
```
OlhaYevstifieieva, line 5: If you use enum, please, look at checklist (Enum constant names usually start with capital letter. Values can have any case that's necessary.)

OlhaYevstifieieva (changes requested):

src/figuresClasses.ts
svitjojo: Ohh... I knew it and for some reason didn't use it.

ArturErmolenko (approved):
Good job 🔥

PR #908, opened 2023-01-18
Solution

maasiimka:


OlhaYevstifieieva (approved):
GJ 👋🏻
Method 'to.fixed()' also can be useful in this task

src/figuresClasses.ts
```diff
@@ -1,19 +1,75 @@
...
+  color: string;
+  getArea():number
 }
 
-export class Triangle {
+export class Triangle implements Figure {
+  shape = 'triangle';
+
+  constructor(
+    public color: string,
```
ArturErmolenko, line 11: Here should be enum with colors
```suggestion
    public color: Color,
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,75 @@
...
+    return Math.floor(Math.PI * radius * radius * 100) / 100;
+  }
 }
 
-export function getInfo(figure) {
+export class Rectangle implements Figure {
+  shape = 'rectangle';
+
+  constructor(
+    public color: string,
```
ArturErmolenko, line 57:
```suggestion
    public color: Color,
```

ArturErmolenko (changes requested):
GJ 👍 
Let's add Color enum ⭐

src/figuresClasses.ts
maasiimka: I pushed what I mean, so you can look at this

ArturErmolenko: agree , your version is better 👍

ArturErmolenko (approved):
Good job 👍

PR #909, opened 2023-01-18
Solution was realised

anna-inozemets:


oleh-holovnykh (approved):
Well done!

src/figuresClasses.ts
```diff
@@ -1,19 +1,80 @@
...
-}
+  constructor(
+    public color: Color,
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    if (this.a <= 0 || this.b <= 0 || this.c <= 0) {
+      throw new Error('Size of side should be greater than 0');
+    }
```
Anna-Tkachenko, line 22: you need to check that input arguments are bigger than 0 for each figure
instead of checking it in each class, let's create one separate function for this purpose, and let's call it for each figure

Anna-Tkachenko (approved):
⭐️

PR #911, opened 2023-01-18
ts_figures_classes

rafmamedov:


src/figuresClasses.ts
```diff
@@ -1,19 +1,85 @@
-export interface Figure {
+type Shape = 'triangle' | 'circle' | 'rectangle';
+type Color = 'red' | 'green' | 'blue';
```
tetiana-ivanchenko, line 2: use enum

tetiana-ivanchenko (changes requested):

src/figuresClasses.ts
```diff
@@ -1,19 +1,94 @@
...
+  shape: Shape;
+  color: Color;
+  a?: number;
+  b?: number;
+  c?: number;
+  radius?: number;
+  width?: number;
+  height?: number;
+  getArea(): number;
 }
```
DarkMistyRoom, line 25: the Figure interface should contain only 3 properties
image
other will be added to each child's class

src/figuresClasses.ts
```diff
@@ -1,19 +1,94 @@
...
+  shape: Shape = Shape.rectangle;
+
+  constructor(
+    public color: Color,
+    public width: number,
+    public height: number,
+  ) {
+    if (width <= 0 || height <= 0) {
+      throw new Error('All values have to be more than 0');
+    }
```
DarkMistyRoom, line 84: now you check if all args are bigger than 0 in each class. to avoid repeating would be good to make a function with this logic

DarkMistyRoom (changes requested):

nazarmatsevych (approved):
Great job!

SanyaBratashchuk (approved):

PR #915, opened 2023-01-18
solution

victorionov:


Serghii29 (commented):
really cool work!

src/figuresClasses.ts
```diff
@@ -1,19 +1,84 @@
...
 export class Circle {
+  shape = Shape.Circle;
+
+  constructor(
+    public color: Color,
+    public radius: number,
+  ) {
+    if (this.radius <= 0) {
+      throw new Error('the radius is not correct');
+    }
```
Anna-Tkachenko, line 56: will be better to create separate function that will accept input arguments and will check if each of them is bigger than 0
and call this function for each figure

Anna-Tkachenko (approved):
Good job!

PR #917, opened 2023-01-18
solution

Oleksandr-Maly:


src/figuresClasses.ts
```diff
@@ -1,19 +1,90 @@
...
+  return Math.floor(area * 100) / 100;
+}
+
+function isTriangle(a:number, b:number, c:number):boolean {
+  const allSides = [a, b, c].sort((x:number, y:number) => x - y);
 
+  return allSides[2] < allSides[0] + allSides[1];
 }
 
 export class Triangle {
```
nkondd, line 24:
```suggestion
export class Triangle implements Figure {
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,90 @@
...
+}
+
+function roundedArea(area: number): number {
+  return Math.floor(area * 100) / 100;
+}
+
+function isTriangle(a:number, b:number, c:number):boolean {
+  const allSides = [a, b, c].sort((x:number, y:number) => x - y);
 
+  return allSides[2] < allSides[0] + allSides[1];
```
nkondd, line 21: I think that it would be better to use this function **inside the class**. 
```suggestion
  const allSides = [a, b, c].sort((x, y) => x - y);

  const isTriangle = allSides[2] < (allSides[0] + allSides[1]);
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,90 @@
...
 
+  constructor(
+    public color:Color,
+    public a:number,
+    public b:number,
+    public c:number,
+  ) {
+    if (a <= 0 || b <= 0 || c <= 0 || !isTriangle(a, b, c)) {
+      throw new Error('triangle with such sides does not exist');
+    }
```
nkondd, line 35: It would be better to divide Errors by 2 types:
1. All sides are positive (first `if` statement)
2. Triangle is valid (second `if` statement with `!isTriangle`)

nkondd (changes requested):
Good job :star:

src/figuresClasses.ts
Denchepornyuk: it would be great to make this feature useful for every shape of the figure

Denchepornyuk (approved):

PR #918, opened 2023-01-18
Solution

tania-boiar:


src/figuresClasses.ts
```diff
@@ -1,19 +1,86 @@
...
+  shape: 'triangle' | 'circle' | 'rectangle',
+  getArea(): number
 }
 
-export class Triangle {
+export class Triangle implements Figure {
+  public shape = 'triangle';
 
-}
+  constructor(public color: string, public a: number,
```
AndreyBerezyanskiy, line 10:
```suggestion
  constructor(
    public color: string, 
    public a: number,
    public b: number,
```

AndreyBerezyanskiy: You should write every key in a new row

src/figuresClasses.ts
```diff
@@ -1,19 +1,86 @@
...
+  getArea(): number {
+    return Math.floor((Math.PI * this.radius * this.radius) * 100) / 100;
+  }
 }
 
-export function getInfo(figure) {
+export class Rectangle implements Figure {
+  public shape = 'rectangle';
+
+  constructor(public color: string, public width: number,
```
AndreyBerezyanskiy, line 68:
```suggestion
  constructor(
    public color: string, 
    public width: number,
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,86 @@
...
+    return +Math.sqrt(s * ((s - this.a) * (s - this.b) * (s - this.c)))
+      .toFixed(2);
+  }
 }
 
-export class Rectangle {
+export class Circle implements Figure {
+  public shape = 'circle';
+
+  constructor(public color: string, public radius: number) {
```
AndreyBerezyanskiy, line 51:
```suggestion
  constructor(
    public color: string, 
    public radius: number
  ) {
```

AndreyBerezyanskiy (changes requested):
Good job! 👌

Just a few suggestions)

src/figuresClasses.ts
tania-boiar: My bad! Thanks 😊

src/figuresClasses.ts
```diff
@@ -1,19 +1,96 @@
 export interface Figure {
-
+  color: 'red' | 'green' | 'blue',
+  shape: 'triangle' | 'circle' | 'rectangle',
```
Anna-Tkachenko, line 3: will be better to create enum with possible values for color and shape properties

src/figuresClasses.ts
```diff
@@ -1,19 +1,96 @@
...
+    }
 
+    this.color = color;
+    this.a = a;
+    this.b = b;
+    this.c = c;
+  }
+
+  getArea(): number {
+    const s: number = (this.a + this.b + this.c) / 2;
```
Anna-Tkachenko, line 45: what is `s`?

src/figuresClasses.ts
```diff
@@ -1,19 +1,96 @@
...
+export class Circle implements Figure {
+  public shape = 'circle';
+
+  constructor(
+    public color: string,
+    public radius: number,
+  ) {
+    if (radius <= 0) {
+      throw new Error('Radius is not valid');
+    }
```
Anna-Tkachenko, line 61: you need to check input arguments for each figure
let's create separate function for this purpose, that will accept needed params, and will check that each of them is bigger than 0
and will call this function in each figure

Anna-Tkachenko (changes requested):

src/figuresClasses.ts
```diff
@@ -1,19 +1,111 @@
...
-export class Triangle {
+export interface Figure {
+  color: Color,
+  shape: Shape,
+  getArea(): number
+}
 
+function validateSides(...args: number[]): void {
+  if (args.some((side: number) => side <= 0)) {
+    throw new Error('Sides are not valid');
```
Anna-Tkachenko, line 21: will be better to specify the reason, why they're not valid

Anna-Tkachenko (approved):

PR #928, opened 2023-01-19
v-01

AndreyBerezyanskiy:


NVyrsta (approved):
Good job👌

src/figuresClasses.ts
```diff
@@ -1,19 +1,89 @@
...
 }
 
-export class Triangle {
+export interface Figure {
+  shape: Shape;
+  color: Color;
+  getArea(): number;
+}
 
+function isNegative(first: number, second: number, third: number = 1): void {
```
Anna-Tkachenko, line 19: function name should contain some action
check what you can do with this function and create corresponding function name

also, don't use separate argument for each variable
if you'll have some figure with 10 sides, will you create the variable for each of them?

let's move all of them to the one array and will process it

src/figuresClasses.ts
```diff
@@ -1,19 +1,89 @@
...
+export class Circle implements Figure {
+  shape: Shape = Shape.Circle;
+
+  constructor(
+    public color: Color,
+    public radius: number,
+  ) {
+    if (radius <= 0) {
+      throw new Error('Radius is negative. Input positive radius');
+    }
```
Anna-Tkachenko, line 61: you're already have the function that can check that input values are bigger than 0
let's call those function for each figure

Anna-Tkachenko (changes requested):

src/figuresClasses.ts
```diff
@@ -1,19 +1,87 @@
...
-export class Triangle {
+export interface Figure {
+  shape: Shape;
+  color: Color;
+  getArea(): number;
+}
 
+function CheckNegativeNumber(...args: number[]): void {
+  if (args.some((item: number) => item <= 0)) {
+    throw new Error('Numbers are negative. Input positive numbers');
```
alexandra-protyanova, line 21:
```suggestion
    throw new Error('Numbers are negative. Enter positive numbers');
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,87 @@
...
 }
 
-export class Triangle {
+export interface Figure {
+  shape: Shape;
+  color: Color;
+  getArea(): number;
+}
 
+function CheckNegativeNumber(...args: number[]): void {
```
alexandra-protyanova, line 19:
```suggestion
function checkNegativeNumber(...args: number[]): void {
```

alexandra-protyanova (approved):
Good job 👍

src/figuresClasses.ts
AndreyBerezyanskiy: Thanks)  I've fixed it.

PR #930, opened 2023-01-19
solution

Dashaverz:


src/figuresClasses.ts
```diff
@@ -1,19 +1,84 @@
...
+    public b: number,
+    public c: number,
+  ) {
+    if (this.a >= this.b + this.c
+      || this.b >= this.a + this.c
+      || this.c >= this.b + this.a
+      || this.a <= 0
+      || this.b <= 0
+      || this.c <= 0) {
+      throw new Error('The length of triangle is not correct');
```
nkondd, line 34: It's better to use array methods instead of `if` statement with multiple conditions.
For triangle here you have 2 conditions: 
1. All sides should be positive. If it's not true, then throw an Error.
`arr.every(side => side > 0)`
2. Triangle is valid (the longest side is bigger than sum of two another)
Find longest side from array (use array methods or `Math.max`). Find sum of all elements from array (`arr.reduce`). Then you can find validation of triangle with: 
`const validTriangle = longestSide < (sumOfSides - longestSide);`
If it's not true, then throw an error.

nkondd (changes requested):
Good job!

src/figuresClasses.ts
```diff
@@ -1,19 +1,90 @@
...
+  shape = Shape.Rectangle;
 
-}
+  constructor(
+    public color: Color,
+    public width: number,
+    public height: number,
+  ) {
+    if (this.height <= 0 || this.width <= 0) {
+      throw new Error('the input is not correct');
```
alexandra-protyanova, line 79: you can create separate function for throwing error, and pass needed phrase there

alexandra-protyanova (approved):
good job 👍

PR #931, opened 2023-01-19
solution

Anton-Zhytonbaiev:


OlhaYevstifieieva (approved):
GJ💫

src/figuresClasses.ts
```diff
@@ -1,19 +1,77 @@
-export interface Figure {
+type Color = 'red' | 'green' | 'blue';
 
+export interface Figure {
+  shape: string;
```
alexandra-protyanova, line 4: you can create `enum` for shape

src/figuresClasses.ts
```diff
@@ -1,19 +1,77 @@
...
+export class Rectangle implements Figure {
+  shape = 'rectangle';
+
+  constructor(
+    public color: Color,
+    public width: number,
+    public height: number,
+  ) {
+    if (width <= 0 || height <= 0) {
+      throw new Error('Wrong sides');
```
alexandra-protyanova, line 64: let us create separate function for `throw error` to prevent duplicating. Pass phrase you want to be thrown to this function

alexandra-protyanova (changes requested):

PR #933, opened 2023-01-19
solution

yaroslavna-horban:


src/figuresClasses.ts
```diff
@@ -1,19 +1,80 @@
...
+  }
 }
 
-export function getInfo(figure) {
+export class Rectangle implements Figure {
+  shape: Shape = 'rectangle';
+
+  constructor(
+    public color: Color,
+    public weight: number,
```
nzrii-kysel, line 65:
```suggestion
    public width: number,
```

nzrii-kysel (approved):

src/figuresClasses.ts
```diff
@@ -1,19 +1,80 @@
...
 export interface Figure {
+  shape: Shape;
+  color: Color;
 
+  getArea(): number;
 }
 
-export class Triangle {
+export class Triangle implements Figure {
+  shape: Shape = 'triangle';
```
Anna-Tkachenko, line 11: will be better to use enum for this values (in this way you'll have less chance to make an error)

src/figuresClasses.ts
```diff
@@ -1,19 +1,80 @@
...
+  constructor(
+    public color: Color,
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    if (this.a <= 0 && this.b <= 0 && this.c <= 0) {
+      throw new Error('Size must be greater than 0');
+    }
+
```
Anna-Tkachenko, line 22: will be better to create separate function that will accept input arguments and will check that each of them is bigger than 0
and call this function for each figure

src/figuresClasses.ts
```diff
@@ -1,19 +1,80 @@
...
+      || this.a + this.c <= this.b
+      || this.c + this.b <= this.a
+    ) {
+      throw new Error('Incorrect size!');
+    }
+  }
 
-export class Circle {
+  getArea(): number {
+    const half: number = (this.a + this.b + this.c) / 2;
```
Anna-Tkachenko, line 33: half what?

Anna-Tkachenko (approved):

PR #934, opened 2023-01-19
Solution

andrii-andriushchenko:


Serghii29 (commented):
Good job!

src/figuresClasses.ts
```diff
@@ -1,19 +1,95 @@
...
+}
 
+export interface Figure {
+  shape: GetShape;
+  color: GetColor;
+  getArea(): number;
 }
 
-export class Triangle {
+function getInvalidSides(...args: number[]): boolean {
```
Anna-Tkachenko, line 19: this function doesn't return invalid sides
will be better to name it somehow using `check` action

src/figuresClasses.ts
```diff
@@ -1,19 +1,95 @@
...
+  const sorted = args.sort((x, y) => x - y);
+
+  return (sorted[0] + sorted[1]) <= sorted[2];
+}
 
+function getZeroValues(...args: number[]): boolean {
```
Anna-Tkachenko, line 25: same here, with this function you can't get zero values

Anna-Tkachenko (approved):

src/figuresClasses.ts
andrii-andriushchenko: Thanks. Changed)

PR #937, opened 2023-01-19
Solution

antskpps:


svitjojo (approved):
Good job 👍 
My advice for code improvement is you can create a function for checking if the input value <= 0

src/figuresClasses.ts
```diff
@@ -1,19 +1,78 @@
...
 
+export interface Figure {
+  shape: Shape,
+  color: Color,
+  getArea(): number,
 }
 
-export class Triangle {
+export class Triangle implements Figure {
+  shape: Shape = 'triangle';
```
alexandra-protyanova, line 11: looks like it is better to use `enum` for shapes

src/figuresClasses.ts
```diff
@@ -1,19 +1,78 @@
...
-export function getInfo(figure) {
+export class Rectangle implements Figure {
+  shape: Shape = 'rectangle';
+
+  constructor(
+    public color: Color,
+    public width: number,
+    public heigth: number,
+  ) {
+    if (this.width <= 0 || this.heigth <= 0) {
```
alexandra-protyanova, line 66: U can create separate function for `throwing errors` or for checking for negative, I duplicate a lot of code

alexandra-protyanova (approved):

PR #938, opened 2023-01-19
Solution

Kravchenko71213:


src/figuresClasses.ts
```diff
@@ -1,19 +1,80 @@
...
+export class Circle implements Figure {
+  shape: Shape = 'circle';
+
+  constructor(
+    public color: Color,
+    public radius: number,
+  ) {
+    if (this.radius <= 0) {
+      throw new Error('radius size must be positive');
+    }
```
svitjojo, line 52:
```suggestion
    checkSizes(this.radius);
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,80 @@
...
 
-}
+  constructor(
+    public color: Color,
+    public width: number,
+    public height: number,
+  ) {
+    if (this.width <= 0 || this.height <= 0) {
+      throw new Error('sides must be positive');
+    }
```
svitjojo, line 70:
```suggestion
    checkSizes(this.width, this.height);
```

svitjojo (changes requested):
Use your function, mate :)

src/figuresClasses.ts
Kravchenko71213: thank you for your comments

src/figuresClasses.ts
```diff
@@ -1,19 +1,78 @@
+type Shape = 'triangle' | 'circle' | 'rectangle';
```
Denchepornyuk, line 1: it's better to use enum for `Shape`

Denchepornyuk (approved):

PR #940, opened 2023-01-20
solution

vladyslav78292:


src/figuresClasses.ts
```diff
@@ -1,19 +1,75 @@
...
+    }
 
-export class Circle {
+    if (a >= b + c || b >= a + c || c >= b + a) {
+      throw new Error('Invalid element length');
+    }
+  }
 
+  getArea(): number {
+    const p = (this.a + this.b + this.c) / 2;
```
nkondd, line 30: Improve naming

src/figuresClasses.ts
```diff
@@ -1,19 +1,75 @@
...
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    if (this.a <= 0 && this.b <= 0 && this.c <= 0) {
+      throw new Error('Invalid element length');
+    }
 
-export class Circle {
+    if (a >= b + c || b >= a + c || c >= b + a) {
```
nkondd, line 24: I recommend you to use more array methods and variables for validation instead of multiple conditions inside `if` statement.

nkondd (approved):

src/figuresClasses.ts
alexandra-protyanova: agree

src/figuresClasses.ts
```diff
@@ -1,19 +1,75 @@
...
-export function getInfo(figure) {
+export class Rectangle implements Figure {
+  shape: Shape = 'rectangle';
+
+  constructor(
+    public color: Color,
+    public width: number,
+    public height: number,
+  ) {
+    if (this.width <= 0 || this.height <= 0) {
```
alexandra-protyanova, line 63: you can create function which will check for negative numbers and will throw error

alexandra-protyanova (changes requested):

alexandra-protyanova (approved):
GJ!

PR #941, opened 2023-01-20
Initial commit

yaroslav-furhala:


src/figuresClasses.ts
```diff
@@ -1,19 +1,83 @@
-export interface Figure {
+// create interface 'Shape' & 'Color' with three classes implementing
+type Shape = 'triangle' | 'circle' | 'rectangle';
+type Color = 'red' | 'green' | 'blue';
 
+// create function error message
```
loralevitska, line 5: Remove comment. Try it write obvious code that explain all we need without comments)
Fix all cases

src/figuresClasses.ts
```diff
@@ -1,19 +1,83 @@
...
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    if (
+      a <= 0
+      || b <= 0
+      || c <= 0
+      || Math.max(a, b, c) >= (a + b + c - Math.max(a, b, c))
+    ) {
```
loralevitska, line 30: Move long expression to variable or two variables with meaningful naming

loralevitska (changes requested):
Almost perfect)
Let's improve

PR #943, opened 2023-01-20
solution_1

Borys-Andrew:


src/figuresClasses.ts
```diff
@@ -1,19 +1,73 @@
...
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    const sides = [a, b, c].sort((n1, n2) => n2 - n1);
+    const isPositiveSides = sides.some((el) => el <= 0);
+    const isInvalidSide = sides[0] >= sides[1] + sides[2];
+
+    if (isPositiveSides || isInvalidSide) {
+      throw new Error(`Fucking ${this.shape} Error`);
```
nzrii-kysel, line 24: 🔥

nzrii-kysel (approved):

src/figuresClasses.ts
Anna-Tkachenko: let's rewrite error text to users friendly one

src/figuresClasses.ts
```diff
@@ -1,19 +1,73 @@
...
-export class Rectangle {
+export class Circle implements Figure {
+  shape: Shape = 'circle';
 
+  constructor(
+    public color: Color,
+    public radius: number,
+  ) {
+    if (this.radius <= 0) {
+      throw new Error(`Fucking ${this.shape} Error`);
```
Anna-Tkachenko, line 44: same here

src/figuresClasses.ts
```diff
@@ -1,19 +1,73 @@
...
+  shape: Shape = 'rectangle';
+
+  constructor(
+    public color: Color,
+    public a: number,
+    public b: number,
+  ) {
+    if (this.a <= 0 || this.b <= 0) {
+      throw new Error(`Fucking ${this.shape} Error`);
+    }
```
Anna-Tkachenko, line 63: you checked is all input arguments are bigger than 0 for each figure
let's create several function for this purpose, that will accept the arguments and will check if each of them is bigger than 0
and will call one function for all figures

PR #944, opened 2023-01-20
Solution

Vasyliy22:


src/figuresClasses.ts
```diff
@@ -1,19 +1,73 @@
 export interface Figure {
+  color: string;
```
agniya-i, line 2: it would be better to add union type for color and shape

src/figuresClasses.ts
```diff
@@ -1,19 +1,73 @@
...
+export class Triangle implements Figure {
+  constructor(
+    public color: string,
+    public a: number,
+    public b: number,
+    public c: number,
+    public shape: string = 'triangle',
+  ) {
+    if (this.a <= 0 || this.b <= 0 || this.c <= 0) {
+      throw new Error('Error');
```
agniya-i, line 17: give more details in the error message

agniya-i (changes requested):

PR #947, opened 2023-01-21
Ts Figures Classes

partnersinbahamas:


partnersinbahamas:
This task not done yes, I posted it cause I don't understand what's problem with few tests, there is wrote `should return a correct square of a circle` and others, but I don't now what's problem. Hope you will give me answear. Thank you

src/figuresClasses.ts
```diff
@@ -1,19 +1,92 @@
...
+    getArea(): number;
 }
 
-export class Circle {
 
+export class Triangle implements Figure {
+    shape = Shapes.Triangle;
+
+    constructor(
+        public color = Colors,
```
TarasHoliuk, line 24:
```suggestion
        public color: Colors,
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,92 @@
...
+        if (width <= 0 || height <= 0) {
+            throw Error('This is not rectangle');
+        }
+    }
+
+    getArea(): number {
+        const {width, height} = this;
+       const area = width * height;
+       
+       return Math.floor(area * 100) / 100;
```
TarasHoliuk, line 86: you are doing the same 3 times - move it to a separate function

src/figuresClasses.ts
```diff
@@ -1,19 +1,92 @@
...
+        return Math.floor(area * 100) / 100;
+    }
 }
 
-export function getInfo(figure) {
+export class Rectangle implements Figure {
+    // public shape = Shapes.Rectangle;
+
+    constructor(
+        public shape = Shapes.Rectangle,
```
TarasHoliuk, line 72:
```suggestion
```
do it in the same way as in Circle and Triangle

src/figuresClasses.ts
```diff
@@ -1,19 +1,92 @@
...
+        if (sides.some((side) => side <= 0)) {
+            throw Error('One side of the sides less then 0')
+        }
+
+        if (sides[0] >= sides[1] + sides[2]) {
+            throw Error('This is not Triangle');
+        }
+    }
+
+    getArea(): number {
```
TarasHoliuk, line 42:
```suggestion
    getArea() {
```
you've already written it in the interface

TarasHoliuk (changes requested):

src/figuresClasses.ts
```diff
@@ -1,19 +1,97 @@
...
+
+    sides.sort((x, y) => x - y);
+
+    if (sides.some((side) => side <= 0)) {
+      throw Error('One side of the sides less then 0');
+    }
+
+    if (sides[0] >= sides[1] + sides[2]) {
+      throw Error('This is not Triangle');
+    }
```
tetiana-ivanchenko, line 42: you should fix this condition, you check only one side, but what if [1] or [2] side is more than sum of others?

tetiana-ivanchenko (changes requested):

src/figuresClasses.ts
partnersinbahamas: Так если [1], [2] больше чем [0] то это будет треугольник, разве нет?

vlad-demchuk: вірно, але потрібно перевіряти інші сторони теж, а зараз перевіряєш лише одну

vlad-demchuk (changes requested):

PR #948, opened 2023-01-22
Solution

SashaViun:


src/figuresClasses.ts
```diff
@@ -1,19 +1,95 @@
...
+    if (a >= b + c
+      || b >= a + c
+      || c >= a + b
+    ) {
+      throw Error('The longest side can\'t be longer than sum of two others');
+    }
+  }
+
+  getArea(): number {
+    const s = (this.a + this.b + this.c) / 2;
```
ihor-soloviov, line 43: It is a better way to name mathematical 's' as 'semiPerimetr'

src/figuresClasses.ts
```diff
@@ -1,19 +1,95 @@
...
+  ) {
+    this.color = color;
+
+    if (a <= 0 || b <= 0 || c <= 0) {
+      throw Error('Length can\'t be equal or less than 0');
+    }
+
+    if (a >= b + c
+      || b >= a + c
+      || c >= a + b
```
ihor-soloviov, line 36: ... or you can sort sides and check if the biggest one is '>=' sum of another ones. But this solution is good!

src/figuresClasses.ts
```diff
@@ -1,19 +1,95 @@
...
+    ) {
+      throw Error('The longest side can\'t be longer than sum of two others');
+    }
+  }
+
+  getArea(): number {
+    const s = (this.a + this.b + this.c) / 2;
+    const area = Math.sqrt(s * (s - this.a) * (s - this.b) * (s - this.c));
 
+    return Math.floor(area * 100) / 100;
```
ihor-soloviov, line 46: DRY with Math.floor(). Make a function for next turning the result

src/figuresClasses.ts
```diff
@@ -1,19 +1,95 @@
...
-export function getInfo(figure) {
+export class Rectangle implements Figure {
+  shape = Shape.Rectangle;
+
+  constructor(
+    public color: Color,
+    public width: number,
+    public height: number,
+  ) {
+    this.color = color;
```
ihor-soloviov, line 79: You are not using color in any of classes so you can skip this part in every class

ihor-soloviov (commented):
Good job man!

PR #949, opened 2023-01-22
Ts_figures_classes

SerhiiVovchenko:


src/figuresClasses.ts
```diff
@@ -1,19 +1,81 @@
...
+    const longestSide = Math.max(a, b, c);
+    const sumOtherSides = (a + b + c) - longestSide;
 
-export class Circle {
+    if (longestSide >= sumOtherSides) {
+      throw new Error('Incorrect values of the sides of the triangle');
+    }
 
+    if (a <= 0 || b <= 0 || c <= 0) {
+      throw new Error('Incorrect values of the sides of the triangle');
```
polosanya, line 27: Same error text for two cases? It's better to specify and make them different

polosanya (approved):
Good job! I'm approving it, but please pay attention to the comment

PR #950, opened 2023-01-23
first try

RudBeee:


src/figuresClasses.ts
```diff
@@ -1,19 +1,77 @@
...
 
+export interface Figure {
+  color: string;
+  shape: string;
+  getArea(): number;
 }
 
-export class Triangle {
+export class Triangle implements Figure {
+  shape = 'triangle';
```
TarasHoliuk, line 10: create unit type/enum (better enum) for Shape type

src/figuresClasses.ts
```diff
@@ -1,19 +1,77 @@
...
 
-}
+  constructor(
+    public color: string,
```
TarasHoliuk, line 13: create enum for colors

TarasHoliuk (changes requested):

src/figuresClasses.ts
RudBeee: but i get color as argument
how do i add it to him enum?

tetiana-ivanchenko: enum should be type of color

tetiana-ivanchenko (changes requested):

nazarmatsevych (approved):

PR #954, opened 2023-01-26
solution

Deniskakaka:


src/figuresClasses.ts
```diff
@@ -1,19 +1,86 @@
 export interface Figure {
-
+  shape: 'triangle' | 'circle' | 'rectangle';
+  color: 'red' | 'green' | 'blue';
+  getArea(): number;
```
IvanFesenko, line 4: Create seperate type Color and use it everywhere ypu need. No need to use union in all places like `'red' | 'green' | 'blue'`

src/figuresClasses.ts
```diff
@@ -1,19 +1,86 @@
...
+    if (radius <= 0) {
+      throw new Error('radius could more than 0');
+    }
+  }
 
+  getArea(): number {
+    // i know it not good but i don't know how rigth rounded down
+    return +(3.1415 * +((this.radius ** 2))
+      .toFixed(2))
+      .toFixed(2);
```
IvanFesenko, line 57:
```suggestion
     const area =  Math.PI * (this.radius ** 2);
     
     return Math.floor(area * 100) / 100;
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,86 @@
...
+    this.shape = 'rectangle';
+
+    if (this.width <= 0 || this.heigth <= 0) {
+      throw new Error('width and heigth could more than 0');
+    }
+  }
+
+  getArea(): number {
+    return +(this.heigth * this.width).toFixed(2);
+  }
```
IvanFesenko, line 81: Use same util to get rounded value

IvanFesenko (changes requested):
Good job. Small fix required

tetiana-ivanchenko (approved):

PR #957, opened 2023-01-27
Solution v0.1

07Spirit:


src/figuresClasses.ts
```diff
@@ -1,19 +1,93 @@
 export interface Figure {
-
+  shape: string;
+  color: string;
```
agniya-i, line 3: let's create `enum` for shapes and colors

src/figuresClasses.ts
```diff
@@ -1,19 +1,93 @@
...
+  getArea: () => number;
 }
 
-export class Triangle {
+export function rounded(area: number): number {
+  let areaFixed = area;
+
+  if (areaFixed !== +areaFixed.toFixed(3)) {
+    areaFixed = +areaFixed.toFixed(3).slice(0, -1);
```
agniya-i, line 11: try to use `Math.floor` here to make code more clear

agniya-i (changes requested):

src/figuresClasses.ts
07Spirit: It is not work enough for test. In test I must do 556.55, but if I use Math.floor, I will have 556.

07Spirit: And if I try 'round floor' the last number, I must use more code then this.

agniya-i: it can be done using something like this `Math.floor(number * 100) / 100` 🙂

agniya-i (approved):

src/figuresClasses.ts
07Spirit: Aaaa, Thanks)

PR #958, opened 2023-01-27
all tests Ok

diolna:


src/figuresClasses.ts
```diff
@@ -1,19 +1,85 @@
+type Shape = 'triangle' | 'circle' | 'rectangle';
```
AlisaVlasova, line 1: make it enum type

src/figuresClasses.ts
```diff
@@ -1,19 +1,85 @@
...
+  shape: Shape;
+  color: Color;
+
+  getArea(): number;
 
 }
 
-export class Triangle {
+export class Triangle implements Figure {
+  shape: Shape;
```
AlisaVlasova, line 13:
```suggestion
  shape: Shape.Triangle;
```
and use like this

AlisaVlasova (changes requested):

PR #962, opened 2023-02-01
Finish task

U-Ra-All:


src/figuresClasses.ts
```diff
@@ -1,19 +1,81 @@
...
+  public shape: string = 'rectangle';
+
+  constructor(
+    public color: string,
+    public width: number,
+    public height: number,
+  ) {
+    switch (true) {
+      case this.width <= 0:
+      case this.height <= 0:
```
witflash, line 68: And here `if` instead of `switch/case` too

U-Ra-All: > Pay attention to my comments, please.

Thanks, done

Viktor-Kost (approved):

PR #964, opened 2023-02-01
figures

Mykola-Busarhin:


src/figuresClasses.ts
```diff
@@ -1,19 +1,108 @@
 export interface Figure {
+  color: string;
+  shape: string;
 
+  getArea(): number;
 }
 
-export class Triangle {
+function isTrianglePossible(a: number, b: number, c:number): boolean {
+  const sum = [a, b, c];
```
ivbutsykin, line 9: `sum` is bad name because I guess to see sum of numbers instead (a + b + c). Suggest to rename in `sides` for example

src/figuresClasses.ts
```diff
@@ -1,19 +1,108 @@
...
+  const largest = Math.max(a, b, c);
```
ivbutsykin, line 10: If you create `sum` array before why not to use destructuring: `const largest = Math.max(...sum);`

src/figuresClasses.ts
```diff
@@ -1,19 +1,108 @@
...
 
+  for (let i = 0; i < 3; i += 1) {
```
ivbutsykin, line 12: Need rewrite more understandable: use `filter` method instead `for` loop and give descriptive name for received variable

src/figuresClasses.ts
```diff
@@ -1,19 +1,108 @@
...
+    if (sum[i] === largest) {
+      sum[i] = 0;
+    }
+  }
+
+  const whatIsLeft = sum.reduce((total, x) => total + x, 0);
```
ivbutsykin, line 18: Also, think about another name instead of `whatIsLeft` and I recommend not to use `x, y, ...` (use `cur`, `side` for ex), it's not a math  =)

src/figuresClasses.ts
```diff
@@ -1,19 +1,108 @@
...
+      throw new Error('Invalid data');
+    }
+
+    if (!isTrianglePossible(this.a, this.b, this.c)) {
+      throw new Error('Creating triangle is not possible');
+    }
+  }
+
+  getArea(): number {
+    const semiP = (this.a + this.b + this.c) / 2;
```
ivbutsykin, line 51: Suggest rename in `halfPerimeter`

ivbutsykin (changes requested):

src/figuresClasses.ts
```diff
@@ -1,19 +1,104 @@
...
 
+  const twoShorterSides = sides.filter((side) => side !== largest);
+  const sumOfShorterSides
+  = twoShorterSides.reduce((total, side) => total + side, 0);
+
+  if (largest >= sumOfShorterSides) {
+    return false;
+  }
+
+  return true;
```
SanyaBratashchuk, line 20:
```suggestion
  return largest < sumOfShorterSides
```

SanyaBratashchuk (approved):

PR #967, opened 2023-02-04
Develop

Vlad-Fedorishchev:


src/figuresClasses.ts
```diff
@@ -1,19 +1,78 @@
...
+    public color: Color,
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    this.shape = 'triangle';
 
-export class Circle {
+    if ((a + b <= c) || (a + c <= b) || (b + c <= a)) {
+      throw new Error('This is not a correct triangle');
```
polosanya, line 22: Why exactly it's not correct? You can make error text more meaningful

src/figuresClasses.ts
```diff
@@ -1,19 +1,78 @@
...
+  public shape: Shape;
+
+  constructor(
+    public color: Color,
+    public radius: number,
+  ) {
+    this.shape = 'circle';
 
+    if (radius <= 0) {
+      throw new Error('This is not a correct circle');
```
polosanya, line 47: Fix all cases

polosanya (changes requested):
image

agniya-i (approved):

PR #968, opened 2023-02-04
Figures solution

Sergey-Kovyazin:


src/figuresClasses.ts
```diff
@@ -1,19 +1,93 @@
...
+    const longestSide = Math.max(a, b, c);
+    const oppositeSides = (a + b + c) - longestSide;
+
+    if (longestSide >= oppositeSides) {
+      throw new Error('Not a triangle!');
+    }
+  }
+
+  getArea(): number {
+    const s = 0.5 * (this.a + this.b + this.c);
```
nazarmatsevych, line 40: what is 's' ?)

nazarmatsevych (approved):
Great!

src/figuresClasses.ts
Sergey-Kovyazin: This is the method of getting the area of a triangle. https://en.wikipedia.org/wiki/Heron's_formula

PR #969, opened 2023-02-06
solution

vlad-trundenko:


src/figuresClasses.ts
```diff
@@ -1,19 +1,78 @@
...
+      throw new Error('Values cannot be <= 0');
+    }
 
+    if (a + b <= c || a + c <= b || b + c <= a) {
+      throw new Error('The triangle does not exist');
+    }
+  }
+
+  getArea(): number {
+    const p: number = (this.a + this.b + this.c) / 2;
```
Anastasiia-Svintsova, line 34: Here will be the number in any case. So you don't need to specify the type here. And better to use more meaningful naming. What does the `p` mean? 
```suggestion
    const p = (this.a + this.b + this.c) / 2;
```

Anastasiia-Svintsova (approved):

PR #975, opened 2023-02-13
commit

skomskiy:


src/figuresClasses.ts
```diff
@@ -1,19 +1,87 @@
-export interface Figure {
+type Color ='red' | 'green' | 'blue';
 
+export interface Figure {
+  shape: 'triangle' | 'circle' | 'rectangle';
```
ViktorHavryliuk, line 4: Better move to separate type too

src/figuresClasses.ts
```diff
@@ -1,19 +1,87 @@
...
+    public color: Color,
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    this.color = color;
+    this.a = a;
+    this.b = b;
+    this.c = c;
+    this.shape = 'triangle';
```
ViktorHavryliuk, line 22: Do you need to set this in constructor if you have already done it above?

src/figuresClasses.ts
```diff
@@ -1,19 +1,87 @@
...
 
-export class Circle {
+    if (this.c >= this.a + this.b) {
```
ViktorHavryliuk, line 24: But what if a or b is bigger than sum of other sides?
Also add check for all sides being larger than 0

src/figuresClasses.ts
```diff
@@ -1,19 +1,87 @@
...
+    public heigth: number,
+    public width: number,
+  ) {
+    this.color = color;
+    this.heigth = heigth;
+    this.width = width;
+    this.shape = 'rectangle';
+
+    if (this.width <= 0 || this.heigth <= 0) {
+      throw new Error('width and heigth could more than 0');
```
ViktorHavryliuk, line 76:
```suggestion
      throw new Error('width and heigth must be more than 0');
```

ViktorHavryliuk (changes requested):

src/figuresClasses.ts
```diff
@@ -1,19 +1,93 @@
...
+    const p = (this.a + this.b + this.c) / 2;
+
+    return +(Math.sqrt(p * (p - this.a) * (p - this.b) * (p - this.c)))
+      .toFixed(2);
+  }
 }
 
-export class Rectangle {
+export class Circle implements Figure {
+  shape: 'circle';
```
etojeDenys, line 46: you can remove this line from each figure

etojeDenys (changes requested):

DarkMistyRoom (approved):

PR #976, opened 2023-02-13
solution

vladyslav-krainyk:


src/figuresClasses.ts
```diff
@@ -1,19 +1,80 @@
...
+export class Rectangle implements Figure {
+  shape: Shapes = 'rectangle';
+
+  constructor(
+    public color: Colors,
+    public width: number,
+    public height: number,
+  ) {
+    if (width <= 0 || height <= 0) {
+      throw new Error('Width or Height must be greater than zero');
```
ViktorHavryliuk, line 69:
```suggestion
      throw new Error('Width and Height must be greater than zero');
```

ViktorHavryliuk (approved):
Good work, only one tiny suggestion

PR #977, opened 2023-02-14
Solution

IhorChaikovskyi:


src/figuresClasses.ts
```diff
@@ -1,19 +1,79 @@
 export interface Figure {
-
+  color: string,
+  shape: string,
```
AlisaVlasova, line 3: create enum types for color and shape

src/figuresClasses.ts
```diff
@@ -1,19 +1,79 @@
...
+  getArea(): number;
 }
 
-export class Triangle {
+export class Triangle implements Figure {
+  shape: string = 'triangle';
```
AlisaVlasova, line 8:
```suggestion
  shape: Shape = Shape.Triangle;
```
and use this way

AlisaVlasova (changes requested):

nazarmatsevych (approved):

PR #979, opened 2023-02-15
solution

Yevhenii-Stanchenko:


src/figuresClasses.ts
```diff
@@ -1,19 +1,84 @@
...
+  constructor(
+    public color: Color,
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    const sides = [a, b, c].sort((d, e) => d - e);
+
+    if (sides[2] >= sides[0] + sides[1]
+        || sides.some((side) => side <= 0)
```
slavanosatskyi, line 26: try to move it to the separate variable with meaningful name, otherwise other developers will have hard times to understand what is this condition about

slavanosatskyi (changes requested):

src/figuresClasses.ts
```diff
@@ -1,19 +1,94 @@
-export interface Figure {
+type Shape = 'triangle' | 'circle' | 'rectangle';
```
AlisaVlasova, line 1: for the future
its better to use enum type here

src/figuresClasses.ts
```diff
@@ -1,19 +1,94 @@
...
 }
 
-export class Circle {
+export class Triangle implements Figure {
+  constructor(
+    public color: Figure['color'],
+    public a: number,
+    public b: number,
+    public c: number,
+    public shape: Figure['shape'] = 'triangle',
```
AlisaVlasova, line 20: so u can use it like this
```suggestion
    public shape: Shape = Shape.Triangle,
```

AlisaVlasova (approved):

PR #982, opened 2023-02-18
Initial solution

Dynamizx:


src/figuresClasses.ts
```diff
@@ -1,19 +1,87 @@
 export interface Figure {
+  shape: string;
+  color: string;
+  getArea(): number;
```
Volodymyr-Melnyk-Fron10der, line 4:
```suggestion
  getArea: () => number;
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,87 @@
...
+}
 
+export function getInfo(figure: Figure): string {
+  return `A ${figure.color} ${figure.shape} - ${figure.getArea()}`;
 }
 
-export class Triangle {
+export class Triangle implements Figure {
+  shape = 'triangle';
```
Volodymyr-Melnyk-Fron10der, line 12: I suggest to create an enum for all shape types

src/figuresClasses.ts
```diff
@@ -1,19 +1,87 @@
...
 
-export function getInfo(figure) {
+export class Rectangle implements Figure {
+  shape = 'rectangle';
+
+  color: string;
+
+  width: number;
+
+  height: number;
```
Volodymyr-Melnyk-Fron10der, line 70:
```suggestion
  shape = 'rectangle';
  color: string;
  width: number;
  height: number;
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,87 @@
...
+  }
 }
 
-export class Rectangle {
+export class Circle implements Figure {
+  shape = 'circle';
+
+  color: string;
+
+  radius: number;
```
Volodymyr-Melnyk-Fron10der, line 45:
```suggestion
  shape = 'circle';
  color: string;
  radius: number;
```

Volodymyr-Melnyk-Fron10der (changes requested):
Great! Just a few suggestions

qoa11a (approved):

PR #983, opened 2023-02-19
solution

kuznetsova-anastasiia:


src/figuresClasses.ts
```diff
@@ -1,19 +1,69 @@
-export interface Figure {
+type Shapes = 'triangle' | 'circle' | 'rectangle';
+type Colors = 'red' | 'green' | 'blue';
```
Bogdan-Mykhailov, line 2:
```suggestion
type ShapesType = 'triangle' | 'circle' | 'rectangle';
type ColorsType = 'red' | 'green' | 'blue';
```

Bogdan-Mykhailov: can help you understand what shapes and colors are

src/figuresClasses.ts
```diff
@@ -1,19 +1,69 @@
...
 
+export interface Figure {
+  shape : Shapes,
+  color: Colors,
```
Bogdan-Mykhailov, line 6:
```suggestion
  shape : ShapesType,
  color: ColorsType,
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,69 @@
...
+  getArea(): number,
 }
 
-export class Triangle {
+export class Triangle implements Figure {
+  shape : Shapes = 'triangle';
```
Bogdan-Mykhailov, line 11:
```suggestion
  shape : ShapesType = 'triangle';
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,69 @@
...
 
-}
+  constructor(
+    public color: Colors,
```
Bogdan-Mykhailov, line 14:
```suggestion
    public color: ColorsType,
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,69 @@
...
+    const s = (this.a + this.b + this.c) / 2;
+    const area = Math.sqrt(s * (s - this.a) * (s - this.b) * (s - this.c));
 
+    return Math.floor(area * 100) / 100;
+  }
 }
 
-export class Rectangle {
+export class Circle implements Figure {
+  shape: Shapes = 'circle';
```
Bogdan-Mykhailov, line 33:
```suggestion
  shape: ShapesType = 'circle';
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,69 @@
...
+
+  constructor(
+    public color: Colors,
```
Bogdan-Mykhailov, line 36:
```suggestion
    public color: ColorsType,
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,69 @@
...
+  }
 
+  getArea(): number {
+    return Math.floor((Math.PI * this.radius ** 2) * 100) / 100;
+  }
 }
 
-export function getInfo(figure) {
+export class Rectangle implements Figure {
+  shape: Shapes = 'rectangle';
```
Bogdan-Mykhailov, line 50:
```suggestion
  shape: ShapesType = 'rectangle';
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,69 @@
...
+
+  constructor(
+    public color: Colors,
```
Bogdan-Mykhailov, line 53:
```suggestion
    public color: ColorsType,
```

Bogdan-Mykhailov (approved):
Good Job ✌🏻

src/figuresClasses.ts
OleksiiBlezniuk: It's the same as to name constants like `${name}Const`, overkill
From TS syntax it's clear that it's a type

src/figuresClasses.ts
```diff
@@ -1,19 +1,69 @@
-export interface Figure {
+type ShapesType = 'triangle' | 'circle' | 'rectangle';
```
OleksiiBlezniuk, line 1:
```suggestion
type Shapes = 'triangle' | 'circle' | 'rectangle';
```

OleksiiBlezniuk: custom types in TS are called starting with capital letter - and from TS syntax it's clear that it's a type

OleksiiBlezniuk: Please fix

OleksiiBlezniuk: @Bogdan-Mykhailov FYI

src/figuresClasses.ts
```diff
@@ -1,19 +1,69 @@
...
-}
+  constructor(
+    public color: ColorsType,
+    public a: number = 0,
+    public b: number = 0,
+    public c: number = 0,
+  ) {
+    if (a <= 0 || b <= 0 || c <= 0 || a + b <= c) {
+      throw new Error('Enter valid values');
+    }
```
OleksiiBlezniuk, line 21: Btw, in all cases we have to check if passed value is greater than 0, let's create helper function for it and reuse in all cases

OleksiiBlezniuk: `a + b <= c`
But what if we'll pass `1 3 2` to this function?

src/figuresClasses.ts
```diff
@@ -1,19 +1,69 @@
...
+  }
 
-export class Circle {
+  getArea(): number {
+    const s = (this.a + this.b + this.c) / 2;
+    const area = Math.sqrt(s * (s - this.a) * (s - this.b) * (s - this.c));
 
+    return Math.floor(area * 100) / 100;
```
OleksiiBlezniuk, line 28: We using same rounding in all cases, so it will be nice to create helper function for it (not critical, but as for me - more logical)

src/figuresClasses.ts
```diff
@@ -1,19 +1,69 @@
...
+  }
 }
 
-export class Rectangle {
+export class Circle implements Figure {
+  shape: ShapesType = 'circle';
```
OleksiiBlezniuk, line 33: Looks like magic string, let's create enum instead of type and use it here

OleksiiBlezniuk (changes requested):
GJ, but there are few issues that will make this solution better

src/figuresClasses.ts
Bogdan-Mykhailov: @OleksiiBlezniuk 👍🏼 @kuznetsova-anastasiia sorry

src/figuresClasses.ts
```diff
@@ -1,19 +1,86 @@
-export interface Figure {
+enum Shapes {
+  triangle = 'triangle',
+  circle = 'circle',
+  rectangle = 'rectangle'
```
illiarerun, line 4: Better name your enums values starting with capital letter
```suggestion
  Triangle = 'triangle',
  Circle = 'circle',
  Rectangle = 'rectangle'
```

illiarerun (approved):

PR #984, opened 2023-02-19
Figure classes is solved.

sergey-vorobei:


src/figuresClasses.ts
```diff
@@ -1,19 +1,75 @@
...
 
-}
+  constructor(
+    public color: Color,
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    if (c >= a + b || b >= a + c || a >= b + c) {
+      throw new Error('Error');
```
anstsot, line 20: it would be great to add better error description
```suggestion
      throw new Error(`Error: sides ${a}, ${b} and ${c} can't form a triangle`);
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,75 @@
...
+    public width: number,
+    public height: number,
+  ) {
+    if (width <= 0 || height <= 0) {
+      throw new Error('Error. One of sids is less than 0');
+    }
+  }
+
+  getArea(): number {
+    return this.width * this.height;
```
anstsot, line 69: what if the width or height is in the following format: _41.345_ and has thousands?
method `getArea` have to return a number rounded down to hundredths. 
```suggestion
    return Math.floor(this.width * this.height) / 100;
```

anstsot (changes requested):
Looks great in general! Just small changes are needed

src/figuresClasses.ts
```diff
@@ -1,19 +1,77 @@
...
+    }
+
+    if (a <= 0 || b <= 0 || c <= 0) {
+      throw new Error('Error. One of sides less or equals 0');
+    }
+  }
 
-export class Circle {
+  getArea(): number {
+    const s = (this.a + this.b + this.c) / 2;
```
sTorba24, line 31:
```suggestion
  getArea(): number {
    const { a, b, c } = this
    const p = (this.a + this.b + this.c) / 2;
```

sTorba24: use destructuring

sTorba24: use meaningful names, `s` it's `p` - semiperimeter(in Math) actually

src/figuresClasses.ts
```diff
@@ -1,19 +1,77 @@
...
+export class Rectangle implements Figure {
+  public shape: Shape = 'rectangle';
+
+  constructor(
+    public color: Color,
+    public width: number,
+    public height: number,
+  ) {
+    if (width <= 0 || height <= 0) {
+      throw new Error('Error. One of sids is less than 0');
```
sTorba24, line 66: Use more meaningful description, you have width and height to describe about it
if width < 0 you automatically have 2 sides < 0, not 'one of sides' :) with height the same sure

sTorba24 (changes requested):
Good work! Let's make your code better :)

anstsot (approved):
Great job, approved!

PR #985, opened 2023-02-19
firstCommit

Roman-Malysh:


src/figuresClasses.ts
```diff
@@ -1,19 +1,73 @@
-export interface Figure {
+type Shape = 'triangle' | 'circle' | 'rectangle';
```
tetiana-ivanchenko, line 1: use enum

src/figuresClasses.ts
```diff
@@ -1,19 +1,73 @@
...
 
-}
+  constructor(
+    public color: Color,
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    if (c >= a + b || b >= a + c || a >= b + c) {
+      throw new Error('Error');
```
tetiana-ivanchenko, line 21: use more meanfull error text

tetiana-ivanchenko (changes requested):

src/figuresClasses.ts
```diff
@@ -1,19 +1,82 @@
...
+        shouldn't be less than a sum of two others`);
+    }
+
+    if (a <= 0 || b <= 0 || c <= 0) {
+      throw new Error('The figure`s values should be more than 0');
+    }
+  }
 
+  getArea(): number {
+    const s = (this.a + this.b + this.c) / 2;
```
sTorba24, line 39:
```suggestion
    const p = (this.a + this.b + this.c) / 2;
```

sTorba24: it's p in Math

src/figuresClasses.ts
```diff
@@ -1,19 +1,82 @@
...
+  circle = 'circle',
+  rectangle = 'rectangle',
 }
 
-export class Triangle {
+enum Color {
+  red = 'red',
+  green = 'green',
+  blue = 'blue'
+}
```
sTorba24, line 11: create Type for it 
`type Colors = 'red' | 'green' | 'blue';` for example

src/figuresClasses.ts
```diff
@@ -1,19 +1,82 @@
...
-export class Circle {
+export class Triangle implements Figure {
+  public shape: Shape = Shape.triangle;
+
+  constructor(
+    public color: Color,
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
```
sTorba24, line 27: set this.a = a ....

src/figuresClasses.ts
```diff
@@ -1,19 +1,82 @@
...
+    if (c >= a + b || b >= a + c || a >= b + c) {
+      throw new Error(`The longest side of a triangle 
+        shouldn't be less than a sum of two others`);
```
sTorba24, line 30: just 'not a triangle'

src/figuresClasses.ts
```diff
@@ -1,19 +1,82 @@
...
+export class Rectangle implements Figure {
+  public shape: Shape = Shape.rectangle;
+
+  constructor(
+    public color: Color,
+    public width: number,
+    public height: number,
+  ) {
+    if (width <= 0 || height <= 0) {
+      throw new Error('Error. One of sids is less than 0');
```
sTorba24, line 71: use width or height is less, because it's 2 sides

sTorba24 (changes requested):
Good work! let's improve it

src/figuresClasses.ts
Roman-Malysh: it is S in math

sTorba24: when you declare P as a perimeter => much logically to name half of perimeter like 'p' 
as for radiuses when we have circumcircle radius it's named R, incenter radius => r
But for this case it doesn't matter. I agree, half of people using `s`, so no worries about it:)

src/figuresClasses.ts
```diff
@@ -1,19 +1,77 @@
...
-export function getInfo(figure) {
+export class Rectangle implements Figure {
+  public shape: Shape = Shape.rectangle;
+
+  constructor(
+    public color: Color,
+    public width: number,
+    public height: number,
+  ) {
+    if (width < 0 || height < 0) {
```
sTorba24, line 65: it was ok before, you can't build rectangle with width or height = 0,

src/figuresClasses.ts
```diff
@@ -1,19 +1,77 @@
...
+      throw new Error('Error. One of sids is less than 0');
```
sTorba24, line 66: Fix error description, width or height set 2 sides. 
so use  'Width or height length cannot be zero or less'
for example

sTorba24 (changes requested):
Some comments

src/figuresClasses.ts
Viktor-Kost: isn't fixed

Viktor-Kost (approved):

PR #986, opened 2023-02-20
initial commit

ihorivna007:


src/figuresClasses.ts
```diff
@@ -1,19 +1,83 @@
...
+    public c: number,
+    public shape = Shape.Triangle,
+  ) {
+    negativeCheck(a);
+    negativeCheck(b);
+    negativeCheck(c);
+
+    if ((a > b && a > c && a >= b + c)
+      || (b > a && b > c && b >= a + c)
+      || (c > b && c > a && c >= b + a)) {
```
loralevitska, line 37: Long and difficult expressions move to variable with meaningful naming

src/figuresClasses.ts
```diff
@@ -1,19 +1,83 @@
...
+      throw new Error(`The longest side of a triangle 
+        shouldn't be less than a sum of two others`);
+    }
+  }
 
-export class Circle {
+  getArea(): number {
+    const p = (this.a + this.b + this.c) / 2;
```
loralevitska, line 44: What is `p`? Don't cut variable and use meaningful variable names

loralevitska (changes requested):
Keep up the good work! Some comments to fix:

src/figuresClasses.ts
```diff
@@ -1,19 +1,87 @@
...
+    throw new Error('The figure`s values should be more than 0');
+  }
 }
+export class Triangle implements Figure {
+  constructor(
+    public color: Color,
+    public firstSide: number,
+    public secondSide: number,
+    public thirdSide: number,
+    public shape = Shape.Triangle,
```
DarkMistyRoom, line 29: you shouldn't set `shape `in the constructor, because it will be always `triangle `for this class

DarkMistyRoom: fix it everywhere

DarkMistyRoom (changes requested):

mykhalenych (approved):

PR #987, opened 2023-02-21
figures_classes

Vadym-Prydatok:


src/figuresClasses.ts
```diff
@@ -1,19 +1,81 @@
...
 }
 
-export class Triangle {
+export class Triangle implements Figure {
+  public shape = Shape.triangle;
 
-}
+  getArea(): number {
+    const s = 0.5 * (this.a + this.b + this.c);
+    const A = Math.sqrt(s * (s - this.a) * (s - this.b) * (s - this.c));
```
DarkMistyRoom, line 23: what are `s `and `A`? use more descriptive names

src/figuresClasses.ts
```diff
@@ -1,19 +1,81 @@
...
 
-export class Circle {
+    return Math.floor(A * 100) / 100;
```
DarkMistyRoom, line 25: move rounding logic to a separate function

src/figuresClasses.ts
```diff
@@ -1,19 +1,81 @@
...
+    public b: number,
+    public c: number,
+  ) {
+    const arraySides = [a, b, c];
+    const positiveFilter = arraySides.filter((num) => num <= 0);
+    const sides = arraySides.sort((current, next) => next - current);
+
+    if (positiveFilter.length > 0 || sides[0] >= sides[1] + sides[2]) {
+      throw new Error('can`t formed triangle');
+    }
```
DarkMistyRoom, line 40: better specify each invalid value. show different messages depending on which value is invalid

src/figuresClasses.ts
```diff
@@ -1,19 +1,81 @@
...
+    return Math.floor(this.radius ** 2 * Math.PI * 100) / 100;
+  }
 
+  constructor(
+    public color: Color,
+    public radius: number,
+  ) {
+    if (radius <= 0) {
+      throw new Error('radius must be positive number');
+    }
```
DarkMistyRoom, line 57: you have to check if all arguments are positive in each constructor, so would be better to use a function for this and avoid duplicating

DarkMistyRoom (changes requested):

src/figuresClasses.ts
```diff
@@ -1,19 +1,94 @@
...
+  color: Color;
+  getArea(): number;
+}
 
+function roundedDown(number: number): number {
+  return Math.floor(number * 100) / 100;
 }
 
-export class Triangle {
+function negativeNumber(numbers: number[]):boolean {
```
TarasHoliuk, line 22:
```suggestion
function checkHasNegativeNumber(numbers: number[]): boolean {
```
use verbs in function names that describe what the function does
`has` because returns boolean, and you check the array, not a single number

Just FYI

TarasHoliuk (approved):

PR #988, opened 2023-02-21
result

novak-k:


src/figuresClasses.ts
```diff
@@ -1,19 +1,78 @@
-export interface Figure {
+type Shape = 'triangle' | 'circle' | 'rectangle';
```
AlisaVlasova, line 1: make it enum type

src/figuresClasses.ts
```diff
@@ -1,19 +1,78 @@
...
 
+export interface Figure {
+  shape: Shape;
+  color: Color;
+  getArea: () => number;
 }
 
-export class Triangle {
+export class Triangle implements Figure {
+  public shape: Shape = 'triangle';
```
AlisaVlasova, line 11:
```suggestion
  public shape: Shape = Shape.TRIANGLE;
```
and use like this

AlisaVlasova (changes requested):
Good job, just a small improvement

src/figuresClasses.ts
novak-k: ok, done

src/figuresClasses.ts
novak-k: ok

sTorba24 (approved):
Good work! Approved!

PR #989, opened 2023-02-21
solution

bohdan-gavrilyk:


src/figuresClasses.ts
```diff
@@ -1,19 +1,71 @@
 export interface Figure {
-
+  shape: string;
+  color: string;
+  getArea(): number;
 }
 
-export class Triangle {
+export class Triangle implements Figure {
+  public shape = 'triangle';
```
vlad-demchuk, line 8: create a literal type for shapes. it should be some exact value (`triangle' | 'circle' | 'rectangle'`), not any string

vlad-demchuk: the same with color

vlad-demchuk (changes requested):
Looks good, just small improvement

nazarmatsevych (approved):

PR #991, opened 2023-02-23
Solution

Vasyl-Pavlenko:


src/figuresClasses.ts
```diff
@@ -1,19 +1,81 @@
 export interface Figure {
-
+  shape: string;
+  color: string;
```
tetiana-ivanchenko, line 3: create enumsfor this properties

tetiana-ivanchenko (changes requested):

VitaliyBondarenko1982 (approved):

PR #994, opened 2023-02-24
commit

devWizHub:


src/figuresClasses.ts
```diff
@@ -1,19 +1,78 @@
-export interface Figure {
+// import { getTypeParameterOwner } from 'typescript';
 
+export interface Figure {
+  shape: string;
+  color: string;
```
Volodymyr-Melnyk-Fron10der, line 5: you have defined values for shape and color, so create an enums for them
image

src/figuresClasses.ts
```diff
@@ -1,19 +1,78 @@
...
+  getArea(): number;
 }
 
-export class Triangle {
+export class Triangle implements Figure {
+  shape = 'triangle';
```
Volodymyr-Melnyk-Fron10der, line 10: use your enum in all places like this

src/figuresClasses.ts
```diff
@@ -1,19 +1,78 @@
...
+
+    if (sides.some((side) => side <= 0)) {
+      throw new Error('All sides must have positive length');
+    }
+
+    if (sides[2] >= (sides[0] + sides[1])) {
+      throw new Error(
+        'The longest side must be less than the sum of the other two sides',
+      );
+    }
```
Volodymyr-Melnyk-Fron10der, line 28: I suggest to move validation commands into separated method

Volodymyr-Melnyk-Fron10der: check it throughout your code

src/figuresClasses.ts
```diff
@@ -1,19 +1,78 @@
...
 
-export class Triangle {
+export class Triangle implements Figure {
+  shape = 'triangle';
+
+  constructor(
+    public color: string,
+    public a: number,
+    public b: number,
+    public c: number,
```
Volodymyr-Melnyk-Fron10der, line 16: set values to `this` instead of making them public

Volodymyr-Melnyk-Fron10der (changes requested):
Keep it up👍

polosanya (approved):

PR #999, opened 2023-02-28
added

vovakov123:


src/figuresClasses.ts
```diff
@@ -1,19 +1,80 @@
...
+      throw new Error(`The longest side of a triangle
+        should be more than a sum of two others`);
+    }
+  }
 
-export class Circle {
+  getArea(): number {
+    const semiperimeter: number = (this.a + this.b + this.c) / 2;
 
+    return Math.floor(Math.sqrt(semiperimeter * (semiperimeter - this.a)
```
anstsot, line 37: maybe it would be better to use `Math.round`? but its not an issue here

anstsot (approved):
Great job, approved!

PR #1003, opened 2023-03-01
figuresClasses was done

pushkarskiyrodion:


src/figuresClasses.ts
```diff
@@ -1,19 +1,73 @@
...
+    const maxSide = Math.max(a, b, c);
+
+    if (maxSide >= (a + b + c) - maxSide) {
+      throw new Error('The max side should be less than a sum of two other');
+    }
+  }
 
-export class Circle {
+  getArea(): number {
+    const sum = 0.5 * (this.a + this.b + this.c);
```
SanyaBratashchuk, line 28:
```suggestion
    const semiPerimeter = 0.5 * (this.a + this.b + this.c);
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,73 @@
...
 
-}
+  constructor(
+    public color: string,
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    if (a <= 0 || b <= 0 || c <= 0) {
+      throw new Error('The digit cannot be less than 0 or equal to 0');
```
SanyaBratashchuk, line 17: You repeat text of this error in each class so move it to the variable

SanyaBratashchuk (changes requested):

src/figuresClasses.ts
pushkarskiyrodion: Ok, thanks for the advice :)

OleziO (approved):

PR #1004, opened 2023-03-02
Solution

lilia-obushenko:


src/figuresClasses.ts
```diff
@@ -1,19 +1,74 @@
...
 
+export interface Figure {
+  shape: Shape,
+  color: Color,
+  getArea(): number,
 }
 
-export class Triangle {
+export class Triangle implements Figure {
+  shape: Shape = 'triangle';
```
illiarerun, line 11: Looks like magic string, let's create enum instead of type and use it here

src/figuresClasses.ts
```diff
@@ -1,19 +1,74 @@
...
 
-}
+  constructor(
+    public color: Color,
+    public width: number,
+    public height: number,
+  ) {
+    if (this.height <= 0 || this.width <= 0) {
+      throw new Error('Side can not be less/equal to zero');
+    }
```
illiarerun, line 64: you have the same logic in Triangle class, create a helper function for it

src/figuresClasses.ts
```diff
@@ -1,19 +1,74 @@
...
+  }
 
-export function getInfo(figure) {
+  getArea(): number {
+    return Math.floor(this.width * this.height * 100) / 100;
```
illiarerun, line 68: you use rounding 3 times, create a helper function, DRY

src/figuresClasses.ts
```diff
@@ -1,19 +1,74 @@
...
+  }
+}
 
+export function getInfo(figure: Figure): string {
+  return `A ${figure.color} ${figure.shape} - ${figure.getArea()}`;
```
illiarerun, line 73: can we use destructuring here?

src/figuresClasses.ts
```diff
@@ -1,19 +1,74 @@
...
+      throw new Error('Invalid input values');
+    }
+  }
 
+  getArea(): number {
+    const p = (this.a + this.b + this.c) / 2;
+
+    return Math.floor(Math.sqrt(
+      p * (p - this.a) * (p - this.b) * (p - this.c),
+    ) * 100) / 100;
```
illiarerun, line 33: can we split it into variables so it'll be easier to read?

illiarerun (changes requested):

src/figuresClasses.ts
```diff
@@ -1,19 +1,93 @@
...
 
-export class Triangle {
+function checkSideLength(...sides: number[]): void {
+  if (sides.some((side) => side <= 0)) {
+    throw new Error('Side can not be less/equal to zero');
+  }
+}
+
+export class Triangle implements Figure {
+  shape = Shape.triangle;
```
sergii-nosachenko, line 26: Shape prop can be protected from changing
```suggestion
  readonly shape = Shape.triangle;
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,93 @@
...
+  ) {
+    checkSideLength(a, b, c);
 
+    if (a + b <= c || a + c <= b || b + c <= a) {
+      throw new Error('Invalid input values');
+    }
+  }
+
+  getArea(): number {
+    const p = (this.a + this.b + this.c) / 2;
```
sergii-nosachenko, line 42: What is `p`?

src/figuresClasses.ts
```diff
@@ -1,19 +1,93 @@
...
+  getArea(): number {
+    const area = this.width * this.height * 100;
 
+    return roundNumber(area);
+  }
 }
 
-export function getInfo(figure) {
+export function getInfo(figure: Figure): string {
+  const { color, shape } = figure;
```
sergii-nosachenko, line 90:
```suggestion
  const { color, shape, getArea } = figure;
```

sergii-nosachenko (changes requested):

src/figuresClasses.ts
lilia-obushenko: it doesn't work in that way

sergii-nosachenko: You are right 🤝

src/figuresClasses.ts
```diff
@@ -1,19 +1,95 @@
+enum Shape {
+  triangle = 'triangle',
+  circle = 'circle',
+  rectangle = 'rectangle',
+}
+
+type Color = 'red' | 'green' | 'blue';
```
sergii-nosachenko, line 7: Using an enum is better here.

sergii-nosachenko (changes requested):

sergii-nosachenko (approved):
Good work 👍

PR #1006, opened 2023-03-02
solution

anna-akhalmosulishvili:


src/figuresClasses.ts
```diff
@@ -1,19 +1,79 @@
...
+export class Triangle implements Figure {
+  constructor(
+    public shape: 'triangle',
+    public color: 'red' | 'green' | 'blue',
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    if (a <= 0 || b <= 0 || c <= 0 || checkTriangle([a, b, c])) {
+      throw new Error('Incorrect data');
```
DarkMistyRoom, line 29: better to specify the error. you can make a common function to check if all parameters are positive values(because it should be checked in all classes) and separate specific checks

src/figuresClasses.ts
```diff
@@ -1,19 +1,79 @@
...
+    }
+  });
 
+  return sum >= maxSide;
 }
 
-export class Circle {
+export class Triangle implements Figure {
+  constructor(
+    public shape: 'triangle',
```
DarkMistyRoom, line 22: the constructor doesn't receive the shape. it should be set as 'triangle' for all instances. check it everywhere

DarkMistyRoom (changes requested):
try to pass all tests. feel free to ask in the chat if you have problems with it
image

src/figuresClasses.ts
```diff
@@ -1,19 +1,88 @@
...
+    public color: Color,
+    public radius: number,
+  ) {
+    if (checkSides([radius])) {
+      throw new Error('Incorrect data: non-positive radius');
+    }
+  }
+
+  getArea(): number {
+    const res = Math.PI * (this.radius ** 2);
```
Volodymyr-Melnyk-Fron10der, line 62: to pass test
```suggestion
    const res = (Math.PI * (this.radius ** 2)).toFixed(1);
```

Volodymyr-Melnyk-Fron10der (approved):
good job👍

PR #1011, opened 2023-03-06
solution for figuresClasses

vladyslav-mosor:


Zelinskyi-Serhii (approved):
gj as for me.

src/figuresClasses.ts
```diff
@@ -1,19 +1,84 @@
...
 
+export interface Figure {
+  shape : Shape,
+  color: Color,
+  getArea(): number,
 }
 
-export class Triangle {
+export class Triangle implements Figure {
+  shape: Shape = 'triangle';
```
sergii-nosachenko, line 11:
```suggestion
  readonly shape: Shape = 'triangle';
```
I think the shape prop can be protected from changing.

src/figuresClasses.ts
```diff
@@ -1,19 +1,84 @@
...
+  getArea(): number {
+    const semiperimeter = (this.a + this.b + this.c) / 2;
+    const area = Math.sqrt(
+      semiperimeter
+      * (semiperimeter - this.a)
+      * (semiperimeter - this.b)
+      * (semiperimeter - this.c),
+    );
+
+    return Math.floor(area * 100) / 100;
```
sergii-nosachenko, line 39: It seems that you can create a helper function for this.

sergii-nosachenko (changes requested):

src/figuresClasses.ts
```diff
@@ -1,19 +1,84 @@
-export interface Figure {
+type Shape = 'triangle' | 'circle' | 'rectangle';
```
sergii-nosachenko, line 1: Can be converted to enum

src/figuresClasses.ts
```diff
@@ -1,19 +1,93 @@
...
+export class Triangle implements Figure {
+  readonly shape = Shape.Triangle;
+
+  constructor(
+    public color: Color,
+    public a: number = 0,
+    public b: number = 0,
+    public c: number = 0,
+  ) {
+    if (a <= 0 || b <= 0 || c <= 0) {
```
sergii-nosachenko, line 28: This and all similar checks can be done with a helper function and array method (some | every) on arguments list.

src/figuresClasses.ts
```diff
@@ -1,19 +1,93 @@
-export interface Figure {
-
+enum Shape {
+  Triangle = 'triangle',
+  Circle = 'circle',
+  Rectangle = 'rectangle'
 }
 
-export class Triangle {
+type Color = 'red' | 'green' | 'blue';
```
sergii-nosachenko, line 7: Using an enum is better here.

sergii-nosachenko (changes requested):

sergii-nosachenko (approved):
Good work 👍

PR #1016, opened 2023-03-07
solution

NikitaStambul:


src/figuresClasses.ts
```diff
@@ -1,19 +1,95 @@
...
+      || a <= 0
+      || b <= 0
+      || c <= 0
+    ) {
+      throw new Error('Impossible triangle');
+    }
+
+    this.a = a;
+    this.b = b;
+    this.c = c;
```
rashyd-hasratov, line 42: You don't need these lines of code,  the properties were added when you wrote them as constructor parameters with access modifiers.

src/figuresClasses.ts
```diff
@@ -1,19 +1,95 @@
...
+  }
+
+  getArea(): number {
+    const semi = (this.a + this.b + this.c) / 2;
+    const { a, b, c } = this;
+    const area = Math.sqrt(semi * (semi - a) * (semi - b) * (semi - c));
```
rashyd-hasratov, line 48:
```suggestion
    const { a, b, c } = this;
    const semi = (a + b + c) / 2;
    const area = Math.sqrt(semi * (semi - a) * (semi - b) * (semi - c));
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,95 @@
...
+    public width: number,
+    public height: number,
+  ) {
+    if (width <= 0 || height <= 0) {
+      throw new Error('Impossible rectangle');
+    }
+  }
+
+  getArea(): number {
+    return this.width * this.height;
```
rashyd-hasratov, line 89: It is better to return the same rounded square everywhere
```suggestion
    return Math.floor((this.width * this.height) * 100) / 100;
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,94 @@
...
+    public b: number,
+    public c: number,
+  ) {
+    if (
+      a >= b + c
+      || b >= a + c
+      || c >= a + b
+      || a <= 0
+      || b <= 0
+      || c <= 0
```
sergii-nosachenko, line 35: Create a helper function that accepts any number of args and checks them. Reuse it with all shapes.

src/figuresClasses.ts
```diff
@@ -1,19 +1,94 @@
...
+    ) {
+      throw new Error('Impossible triangle');
+    }
+  }
+
+  getArea(): number {
+    const { a, b, c } = this;
+    const semi = (a + b + c) / 2;
```
sergii-nosachenko, line 43: Semiperimeter?

src/figuresClasses.ts
```diff
@@ -1,19 +1,94 @@
...
+    const area = Math.sqrt(semi * (semi - a) * (semi - b) * (semi - c));
+    const roundedArea = Math.floor(area * 100) / 100;
```
sergii-nosachenko, line 45: Create a helper function for rounding numbers.

sergii-nosachenko (changes requested):

OleksiiBlezniuk (approved):

PR #1022, opened 2023-03-07
figuresClasses was implement

KuharchukDmytro:


Bogdan-Mykhailov (changes requested):
Nice work but i have some suggestions to improve your solution ✌🏻

Bogdan-Mykhailov (approved):
👍🏼

src/figuresClasses.ts
```diff
@@ -1,19 +1,92 @@
...
+}
 
+enum Color {
+  Red = 'red',
+  Green = 'green',
+  Blue = 'blue',
 }
 
-export class Triangle {
+function isValid(...args: number[]): void {
```
sergii-nosachenko, line 19: The name `isValid` refers to the boolean type. Rename function to describe the action it does.

src/figuresClasses.ts
```diff
@@ -1,19 +1,92 @@
...
+    isValid(a, b, c);
 
+    if (a >= b + c || b >= a + c || c >= a + b) {
+      throw new Error('Invalid input: hypotenuse >= legs sum');
+    }
+  }
+
+  getArea(): number {
+    const { a, b, c } = this;
+    const area: number = 0.25 * Math.sqrt(((a * a + b * b + c * c) ** 2)
```
sergii-nosachenko, line 45: `* 0.25` === ` / 4`
`a * a` === `a ** 2`
`(a * a + b * b + c * c) ** 2` === `a ** 4 + ...` (and it's repeated twice).

src/figuresClasses.ts
```diff
@@ -1,19 +1,92 @@
...
+  getArea(): number {
+    const { width, height } = this;
 
+    return width * height;
+  }
 }
 
-export function getInfo(figure) {
+export function getInfo(figure: Figure): string {
+  const { color, shape }: Figure = figure;
```
sergii-nosachenko, line 89:
```suggestion
  const { color, shape } = figure;
```
Useless when restructuring. It always knows the type of right-hand operator, so you couldn't extract non-existing props.

sergii-nosachenko (changes requested):

src/figuresClasses.ts
KuharchukDmytro: on my opninion, (a * a + b * b + c * c) ** 2 better leave how it was, because when we assign it, it looks much more complicated
image

PR #1024, opened 2023-03-08
Develop

Yekatery:


gretoq (approved):

src/figuresClasses.ts
```diff
@@ -1,19 +1,98 @@
...
+  shape: Shape,
+  color: Color,
+  getArea(): number,
+}
+
+function round(area: number): number {
+  return Math.floor((area) * 100) / 100;
+}
 
+function findSide(...sides: number[]): void {
```
OleksiiBlezniuk, line 23: Function name should represent action that is performing by it. From name `findSide` I expect that this function will return some `side` of figure, but instead - it returns nothing. Please use more clear names for your variables and functions

OleksiiBlezniuk, line 23:
```suggestion
function validateSides(...sides: number[]): void {
```

OleksiiBlezniuk (approved):
GJ, but use more clear names for functions

PR #1026, opened 2023-03-08
Solution

Ukrainiane-panda:


src/figuresClasses.ts
```diff
@@ -1,19 +1,85 @@
...
-
+  shape: string;
+  color: string;
+  getArea(): number;
 }
 
-export class Triangle {
+export class Triangle implements Figure {
+  shape: string = 'triangle';
+
```
IvanFesenko, line 9: Looks like useless type annotation
```suggestion
  shape = 'triangle';

```

IvanFesenko (approved):

PR #1028, opened 2023-03-08
Solution v1

yaroslav-vasilyev:


src/figuresClasses.ts
```diff
@@ -1,19 +1,78 @@
...
+export class Triangle implements Figure {
+  shape: Shape = 'triangle';
+
+  constructor(
+    public color: Colors,
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    if (a <= 0 || b <= 0 || c <= 0) {
```
milenadiamonds1, line 19: you can simplify it with Math.min

src/figuresClasses.ts
```diff
@@ -1,19 +1,78 @@
...
+      throw new Error('Length of triangle sides should be greater than 0');
+    }
+
+    if (a >= b + c || b >= a + c || c >= a + b) {
```
milenadiamonds1, line 23: you can simplify it too

milenadiamonds1 (approved):
GJ with some fixes!

src/figuresClasses.ts
```diff
@@ -1,19 +1,78 @@
...
 
+export interface Figure {
+  shape: Shape;
+  color: Colors;
+  getArea(): number;
 }
 
-export class Triangle {
+export class Triangle implements Figure {
+  shape: Shape = 'triangle';
```
OleksiiBlezniuk, line 11: Looks like magic string, let's use enum instead

src/figuresClasses.ts
OleksiiBlezniuk, line 19: Btw, in all figures you checking that every side should be greater than 0, let's create helper method for it (you can use rest operator for receiving different amount of params)

src/figuresClasses.ts
```diff
@@ -1,19 +1,78 @@
...
+    if (width <= 0 || height <= 0) {
+      throw new Error('Width and height of rectangle should be greater than 0');
+    }
+  }
 
-export function getInfo(figure) {
+  getArea(): number {
+    const area = this.width * this.height;
+
+    return Math.floor(area * 100) / 100;
```
OleksiiBlezniuk, line 72: You've written this logic in all classes, let's instead create helper function

src/figuresClasses.ts
```diff
@@ -1,19 +1,78 @@
...
+  }
+
+  getArea(): number {
+    const area = Math.PI * (this.radius ** 2);
+
+    return Math.floor(area * 100) / 100;
+  }
 }
 
 export class Rectangle {
```
OleksiiBlezniuk, line 56:
```suggestion
export class Rectangle implements Figure {
```

OleksiiBlezniuk (changes requested):

src/figuresClasses.ts
```diff
@@ -1,19 +1,99 @@
+
+type Colors = 'red' | 'green' | 'blue';
```
sergii-nosachenko, line 2: Better use enum

src/figuresClasses.ts
```diff
@@ -1,19 +1,99 @@
...
+  Rectangle = 'rectangle',
+}
+
 export interface Figure {
+  shape: Shape;
+  color: Colors;
+  getArea(): number;
+}
+
+function validateTriangleLength(...shapeSide: number[]): boolean {
```
sergii-nosachenko, line 16:
```suggestion
function isNotValidTriangle(...shapeSide: number[]): boolean {
```
As you're returning a boolean value it's better to set proper name.

src/figuresClasses.ts
```diff
@@ -1,19 +1,99 @@
...
+  const sumSides = shapeSide
+    .reduce((sideOne: number, sideTwo: number) => sideOne + sideTwo);
 
+  return shapeSide.some((side: number) => side >= (sumSides - side));
 }
 
-export class Triangle {
+function validateLength(...lengths : number[]): boolean {
```
sergii-nosachenko, line 23: Same here

src/figuresClasses.ts
```diff
@@ -1,19 +1,99 @@
...
+  return Math.min(...lengths) <= 0;
+}
 
+function helperArea(area: number): number {
```
sergii-nosachenko, line 27: This function return rounded value (not only area)

src/figuresClasses.ts
```diff
@@ -1,19 +1,99 @@
...
+  return Math.floor(area * 100) / 100;
 }
 
-export class Circle {
+export class Triangle implements Figure {
+  shape: Shape = Shape.Triangle;
```
sergii-nosachenko, line 32:
```suggestion
  readonly shape: Shape = Shape.Triangle;
```
The shape prop can be protected from changes.

src/figuresClasses.ts
```diff
@@ -1,19 +1,99 @@
...
+    }
 
+    if (validateTriangleLength(a, b, c)) {
+      throw new Error('The longest side of the triangle must be shorter than'
+       + ' the sum of the other two');
+    }
+  }
+
+  getArea(): number {
+    const p = (this.a + this.b + this.c) / 2;
```
sergii-nosachenko, line 51: What is `p`?

sergii-nosachenko (changes requested):

src/figuresClasses.ts
```diff
@@ -1,19 +1,103 @@
...
 }
 
-export class Triangle {
+function isValidLength(...lengths : number[]): boolean {
+  return Math.min(...lengths) <= 0;
+}
 
+function roundingArea(area: number): number {
+  return Math.floor(area * 100) / 100;
 }
```
Andr1yK, line 32:
```suggestion
function roundArea(area: number): number {
  return Math.floor(area * 100) / 100;
}
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,103 @@
...
+  const sumSides = shapeSide
+    .reduce((sideOne: number, sideTwo: number) => sideOne + sideTwo);
 
+  return shapeSide.some((side: number) => side >= (sumSides - side));
 }
 
-export class Triangle {
+function isValidLength(...lengths : number[]): boolean {
+  return Math.min(...lengths) <= 0;
+}
```
Andr1yK, line 28: what? Why -1 is valid?

image

src/figuresClasses.ts
```diff
@@ -1,19 +1,103 @@
...
+    public c: number,
+  ) {
+    if (isValidLength(a, b, c)) {
+      throw new Error('Length of triangle sides should be greater than 0');
+    }
 
+    if (isNotValidTriangle(a, b, c)) {
+      throw new Error('The longest side of the triangle must be shorter than'
+       + ' the sum of the other two');
+    }
```
Andr1yK, line 50: u can move it to some private method

Andr1yK (changes requested):

src/figuresClasses.ts
yaroslav-vasilyev: Why? We use it only in triangle

Andr1yK (approved):

src/figuresClasses.ts
Andr1yK: and I propose to make it a private method of the Triangle class 🙂

PR #1031, opened 2023-03-08
Figure classes

andriibezkorovainyi:


src/figuresClasses.ts
```diff
@@ -1,19 +1,77 @@
...
+      sideA >= sideB + sideC
+      || sideB >= sideA + sideC
+      || sideC >= sideA + sideB
+      || sideA <= 0
+      || sideB <= 0
+      || sideC <= 0
+    ) {
+      throw new Error('Invalid triangle');
+    }
+  }
```
valeriiayarzhemska, line 26: maybe it would be better to do it with sort method? too much code, check the solution which Sergiy published in Slack, if u need some help to understand

src/figuresClasses.ts
```diff
@@ -1,19 +1,77 @@
...
-export class Circle {
+  getArea(): number {
+    const semiPerimeter = (this.sideA + this.sideB + this.sideC) * 0.5;
 
+    return +Math.sqrt(
+      semiPerimeter
+      * (semiPerimeter - this.sideA)
+      * (semiPerimeter - this.sideB)
+      * (semiPerimeter - this.sideC),
+    ).toFixed(2);
```
valeriiayarzhemska, line 36: in every case u use different way to round squre, u can do like this Math.floor(value * 100) / 100 and make the function outside to write less code

src/figuresClasses.ts
```diff
@@ -1,19 +1,77 @@
 export interface Figure {
-
+  shape: 'triangle' | 'circle' | 'rectangle',
+  color: 'red' | 'green' | 'blue',
```
valeriiayarzhemska, line 3: here would be good to use enum, how do u think?

valeriiayarzhemska (changes requested):
some things to make better:)

valeriiayarzhemska (approved):

src/figuresClasses.ts
```diff
@@ -1,19 +1,102 @@
...
+    const [a, b, c] = sides.sort(
+      (prev, next) => prev - next,
+    );
+
+    if (c >= a + b) {
+      throw new Error(
+        `Can't build ${shape} with these sides: ${sides.join(', ')}`,
+      );
+    }
+  }
```
OleksiiBlezniuk, line 36: Helper functions should be as generic as it possible, so better instead move this logic to separate function

OleksiiBlezniuk (approved):

PR #1032, opened 2023-03-08
Solution

mstrilec:


arkushenko-anna (approved):
Good Job!

src/figuresClasses.ts
```diff
@@ -1,19 +1,100 @@
...
+export interface Figure {
+  shape: Shape,
+  color: Color,
+  getArea(): number,
 }
 
-export class Triangle {
+function checkLength(...args: number[]): void {
+  if (args.some((side) => side <= 0)) {
+    throw new Error('Please, enter valid values');
```
sergii-nosachenko, line 21: It's not an error, it's a user prompt.
```suggestion
    throw new Error('Incorrect figure parameters: all values need to be > 0!');
```

sergii-nosachenko (changes requested):

src/figuresClasses.ts
```diff
@@ -1,19 +1,100 @@
...
 }
 
-export class Triangle {
+function checkLength(...args: number[]): void {
+  if (args.some((side) => side <= 0)) {
+    throw new Error('Incorrect figure parameters: all values need to be > 0');
+  }
+}
 
+function floorArea(area: number): number {
```
OleksiiBlezniuk, line 25:
```suggestion
function roundDownToHundredth(number: number): number {
```

OleksiiBlezniuk: It's a generic helper, it will work same for any number, not only for areas

OleksiiBlezniuk (approved):

PR #1035, opened 2023-03-08
Solution

ksavelyeva:


Ambulence (approved):

src/figuresClasses.ts
```diff
@@ -1,19 +1,95 @@
...
+  constructor(
+
+    public color: Color,
+
+    public a: number,
+
+    public b: number,
+
+    public c: number,
+  ) {
```
OleksiiBlezniuk, line 31:
```suggestion
  constructor(
    public color: Color,
    public a: number,
    public b: number,
    public c: number,
  ) {
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,95 @@
...
+    if (a <= 0 || b <= 0 || c <= 0) {
+      throw new Error('All sides should have a length greater than 0');
+    }
```
OleksiiBlezniuk, line 34: You have similar check in all classes, would be better to move this to helper function

src/figuresClasses.ts
```diff
@@ -1,19 +1,95 @@
...
+    if ((a + b <= c) || (a + c <= b) || (b + c <= a)) {
+      throw new Error('Invalid triangle sides');
+    }
+  }
+
+  getArea(): number {
+    const p = (this.a + this.b + this.c) / 2;
+    const area = Math.sqrt(p * (p - this.a) * (p - this.b) * (p - this.c));
+
+    return Math.floor(area * 100) / 100;
```
OleksiiBlezniuk, line 45: Same here, would be better to create helper that will round numbers

OleksiiBlezniuk (approved):

PR #1036, opened 2023-03-08
created program

vadosik15:


src/figuresClasses.ts
```diff
@@ -1,19 +1,108 @@
...
+  }
 
+  getArea(): number {
+    const semiperimetr = (this.a + this.b + this.c) / 2;
+    const area = Math.sqrt(
+      semiperimetr
+      * (semiperimetr - this.a)
+      * (semiperimetr - this.b)
+      * (semiperimetr - this.c),
+    );
```
Yevhenii0536, line 57: you can use here destruction

src/figuresClasses.ts
```diff
@@ -1,19 +1,108 @@
...
+  readonly shape = Shape.Circle;
+
+  constructor(
+    public color:Color,
+    public radius: number,
+  ) {
+    super();
+
+    if (this.radius <= 0) {
+      throw new Error('wrong radius of circle');
```
Yevhenii0536, line 73: create helper function, which helps you to check this parameters

src/figuresClasses.ts
```diff
@@ -1,19 +1,108 @@
...
+
+    if (this.width <= 0 || this.height <= 0) {
+      throw new Error('some side is to small');
+    }
+  }
+
+  getArea(): number {
+    const area = this.width * this.height;
+
+    return +(area.toFixed(2));
```
Yevhenii0536, line 102: use constructor Number, instead - 'plus'

Yevhenii0536 (changes requested):

src/figuresClasses.ts
```diff
@@ -1,19 +1,108 @@
...
+  getArea(): number {
+    const semiperimetr = (this.a + this.b + this.c) / 2;
+    const area = Math.sqrt(
+      semiperimetr
+      * (semiperimetr - this.a)
+      * (semiperimetr - this.b)
+      * (semiperimetr - this.c),
+    );
+
+    return +(area.toFixed(2));
```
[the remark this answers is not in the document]
vadosik15: So for this i need use (this) in abstract class but i can't so we dont need this function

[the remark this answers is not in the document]
Yevhenii0536: i mean this:

instead this construction you can make helper
 return +(area.toFixed(2))

Yevhenii0536 (approved):
NJ!

PR #1042, opened 2023-03-08
solution

chernenko2:


src/figuresClasses.ts
```diff
@@ -1,19 +1,99 @@
...
+  getArea(): number {
+    const semiperimeter: number = (this.a + this.b + this.c) / 2;
+    const area: number = Math.sqrt(
+      semiperimeter
+      * (semiperimeter - this.a)
+      * (semiperimeter - this.b)
+      * (semiperimeter - this.c),
+    );
+
+    return +(area.toFixed(2));
```
OleksiiBlezniuk, line 56: IMO it's a strange approach, you've used different rounding logic to get same result (here is `toFixed`, it's division by 100 in next class)

Better to create helper function for that purpose and reuse it in all cases

src/figuresClasses.ts
```diff
@@ -1,19 +1,99 @@
...
+export class Circle implements Figure {
+  readonly shape = Shape.Circle;
+
+  constructor(
+    public color: Color,
+    public radius: number,
+  ) {
+    if (radius <= 0) {
+      throw new Error('Wrong length of radius');
+    }
```
OleksiiBlezniuk, line 69: But you already have method for it, reuse `checkSizes` method

OleksiiBlezniuk (approved):
Will approve, but please pay attention to comments

src/figuresClasses.ts
DenysDubas, line 69: here you can check the already created function checkSides.

create an abstract class that will implement Figure and add a function to it.

```

abstract class BasedForm implements Figure {
  abstract shape: Shape;

  abstract color: Color;

  abstract getArea(): number;

  protected checkAllSides(sides: number[]): void {
    if (sides.some((value: number) => value <= 0)) {
      throw new Error(
        `figure of shape ${this.shape} should have only positive params`,
      );
    }
  }
}
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,99 @@
...
+  readonly shape = Shape.Rectangle;
+
+  constructor(
+    public color: Color,
+    public width: number,
+    public height: number,
+  ) {
+    if (Math.min(width, height) <= 0) {
+      throw new Error('Wrond length of the width or height');
+    }
```
DenysDubas, line 89: here you can check the already created function checkSides.

src/figuresClasses.ts
```diff
@@ -1,19 +1,99 @@
...
+
+  getArea(): number {
+    const area = Math.PI * this.radius ** 2;
 
+    return Math.floor(area * 100) / 100;
+  }
 }
 
-export function getInfo(figure) {
+export class Rectangle implements Figure {
```
DenysDubas, line 79: ```export class Rectangle extends BasedForm```

DenysDubas (changes requested):

OleksiiBlezniuk (approved):
Approve, but pay attention to comments

PR #1048, opened 2023-03-09
figuresClasses done

Stepan-TS:


src/figuresClasses.ts
```diff
@@ -1,19 +1,112 @@
...
+  constructor(
+    public color: Color,
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    super();
+
+    const arrayOfSides = [a, b, c];
+    const isSomeSideNegative:boolean = arrayOfSides.some(
```
OleksiiBlezniuk, line 36: No need to specify type here, TS understand that there will be boolean value since it's primitive

OleksiiBlezniuk: And you've written similar logic in all classes, would be logical to create helper method for it and reuse it

src/figuresClasses.ts
```diff
@@ -1,19 +1,112 @@
...
+    const semiperimeter = (this.a + this.b + this.c) / 2;
+
+    const area = Math.sqrt(
+      semiperimeter
+      * (semiperimeter - this.a)
+      * (semiperimeter - this.b)
+      * (semiperimeter - this.c),
+    );
+
+    return Number(area.toFixed(2));
```
OleksiiBlezniuk, line 65: Why did you use `toFixed` here, but `math.floor(area * 100) / 100` in other classes? IMO it looks strange

OleksiiBlezniuk: Better to use one approach and to create helper function for it

src/figuresClasses.ts
```diff
@@ -1,19 +1,112 @@
...
+
+  constructor(
+    public color: Color,
+    public width: number,
+    public heigth: number,
+  ) {
+    super();
+
+    if (width <= 0 || heigth <= 0) {
+      throw new Error('Width adn Heigth should be only positive!');
```
OleksiiBlezniuk, line 101:
```suggestion
      throw new Error('Width and Heigth should be only positive!');
```

OleksiiBlezniuk (approved):
Approve, but pay attention to comments

PR #1049, opened 2023-03-09
Solution

BozhenaBodnar:


src/figuresClasses.ts
```diff
@@ -1,19 +1,118 @@
...
+
+  constructor(
+    public color: Color,
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    super();
+
+    this.checkAllSides([a, b, c]);
```
OleksiiBlezniuk, line 50: Or you can use rest operator in this function to be able pass any amount of params (up to you)

OleksiiBlezniuk (approved):

PR #1051, opened 2023-03-09
Solution for ts_figures_classes

petrovskyi-yevhen:


src/figuresClasses.ts
```diff
@@ -1,19 +1,75 @@
...
+    public c: number,
+  ) {
+    if (a <= 0 || b <= 0 || c <= 0) {
+      throw new Error(`Any length can't be <= 0`);
+    }
+
+    const longestSide = Math.max(a, b, c);
+
+    if (longestSide >= ((a + b + c) - longestSide)) {
+      throw new Error('The longest side of a triangle should be >= than a sum of two others');
```
tetiana-ivanchenko, line 26: fix linter errors

src/figuresClasses.ts
```diff
@@ -1,19 +1,75 @@
-export interface Figure {
+type Shape = 'triangle' | 'circle' | 'rectangle';
```
tetiana-ivanchenko, line 1: use enum

tetiana-ivanchenko (changes requested):

src/figuresClasses.ts
```diff
@@ -1,19 +1,81 @@
...
+      throw new Error('The longest side of a triangle should'
+      + 'be >= than a sum of two others');
+    }
+  }
+
+  getArea(): number {
+    const sum = 0.5 * (this.a + this.b + this.c);
+
+    return Math.floor(Math.sqrt(sum * (sum - this.a) * (sum - this.b)
+    * (sum - this.c)) * 100) / 100;
```
Anastasiia-Svintsova, line 40: Would be better to divide it into separate variables.
Like 
```
const square = ...
return Math.floor(Math.sqrt(square) * 100) / 100;

Anastasiia-Svintsova (approved):

PR #1055, opened 2023-03-10
Solution

DenysDubas:


src/figuresClasses.ts
```diff
@@ -1,19 +1,110 @@
...
+  getArea(): number {
+    const semiperimetr = (this.a + this.b + this.c) / 2;
+    const area = Math.sqrt(
+      semiperimetr
+    * (semiperimetr - this.a)
+    * (semiperimetr - this.b)
+    * (semiperimetr - this.c),
+    );
+
+    return Number(area.toFixed(2));
```
OleksiiBlezniuk, line 65: Why did you use different rounding logic for different classes? IMO - it's a strange approach, better to create helper function for that purpose

src/figuresClasses.ts
```diff
@@ -1,19 +1,110 @@
...
+  readonly shape = Shape.Rectangle;
+
+  constructor(
+    public color: Color,
+    public width: number,
+    public height: number,
+  ) {
+    super();
+
+    this.checkAllSides([this.height, this.width]);
```
OleksiiBlezniuk, line 98: In previous classes you didn't use `this` for reading these properties, let's use one approach instead

OleksiiBlezniuk (approved):
Approve, but pay attention to comments, let's use one approach for particular logic instead of writing something different every time

PR #1056, opened 2023-03-11
Figure classes completed

ArthurKrepko:


src/figuresClasses.ts
```diff
@@ -1,19 +1,114 @@
...
+    super();
+
+    this.checkAllSides([radius]);
+  }
+
+  getArea(): number {
+    const area = Math.PI * (this.radius ** 2);
 
+    return Math.floor(area * 100) / 100;
+  }
```
kuzmi491, line 89: maybe you need a separate function for this... but that's only my opinion=))

kuzmi491 (approved):

src/figuresClasses.ts
Andr1yK: agree

src/figuresClasses.ts
```diff
@@ -1,19 +1,114 @@
...
 
+  abstract getArea(): number;
+
+  protected checkAllSides(sides: number[]): void {
+    if (sides.some((side) => side <= 0)) {
+      throw new Error(
+        `Figure of Shape ${this.shape} should have only positive params`,
+      );
+    }
+  }
```
Andr1yK, line 32: 🔥

src/figuresClasses.ts
```diff
@@ -1,19 +1,114 @@
...
+
+    const [
+      sideA,
+      sideB,
+      sideC,
+    ]: number[] = [a, b, c].sort((prev, curr) => prev - curr);
+
+    if (sideC >= sideB + sideA) {
+      throw new Error('Invalid sides length!');
+    }
```
Andr1yK, line 56: let's move to private method

src/figuresClasses.ts
```diff
@@ -1,19 +1,114 @@
...
+    const semiPerimetr = (this.a + this.b + this.c) / 2;
+
+    const area = Math.sqrt(
+      semiPerimetr
+      * (semiPerimetr - this.a)
+      * (semiPerimetr - this.b)
+      * (semiPerimetr - this.c),
+    );
+
+    return +area.toFixed(2);
```
Andr1yK, line 69: better to use Math.round/floor

Andr1yK (approved):

PR #1059, opened 2023-03-12
the task was solved

batiukova-nataliia:


src/figuresClasses.ts
```diff
@@ -1,19 +1,81 @@
...
+    public color: Color,
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    if ((a <= 0 || b <= 0 || c <= 0)
+      || ((a + b) <= c || (a + c) <= b || (b + c) <= a)
+    ) {
+      throw new Error('invalid data');
+    }
```
etojeDenys, line 33: we can put some of this into a separate variable or break it into two if`s for better readability

etojeDenys (approved):
LGTM

PR #1062, opened 2023-03-14
Solution

OlhaKissa:


src/figuresClasses.ts
```diff
@@ -1,19 +1,75 @@
...
+
+  getArea(): number {
+    const square = (this.width * this.height);
+
+    return square;
+  }
+}
 
+export function getInfo(figure: string): string {
+  return `A ${this.Color} ${this.Shape} - ${this.getArea(figure)}`;
```
VitaliyBondarenko1982, line 74:
```suggestion
export function getInfo(figure: Figure): string {
  return `A ${figure.color} ${figure.shape} - ${figure.getArea()}`;
```

VitaliyBondarenko1982: some tests failed because incorrect getInfo function

src/figuresClasses.ts
```diff
@@ -1,19 +1,75 @@
...
+  }
 
-export class Circle {
+  getArea(): number {
+    const s: number = (this.a + this.b + this.c) / 2;
+    const area = Math.sqrt(
+      s * ((s - this.a) * (s - this.b) * (s - this.c))
+    );
 
+    return Math.floor((area * 100) / 100);
```
VitaliyBondarenko1982, line 30:
```suggestion
    return Math.floor(area * 100) / 100;
```

VitaliyBondarenko1982: check count getArea for triangle. looks like incorrect. should be to digits after comma

VitaliyBondarenko1982 (changes requested):

etojeDenys (approved):
looks good

PR #1065, opened 2023-03-15
Solution

MaksymHuriev:


package.json
```diff
@@ -25,8 +25,8 @@
     "typescript": "^4.3.5"
   },
   "devDependencies": {
-    "@mate-academy/scripts": "^0.9.2",
     "@mate-academy/eslint-config": "*",
+    "@mate-academy/scripts": "^1.2.8",
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,74 @@
...
+      throw new Error('The length side is zero or less');
+    }
+  }
 
+  getArea(): number {
+    const p: number = (this.a + this.b + this.c) / 2;
+
+    return (Math.round(Math.sqrt(p * (p - this.a) * (p - this.b)
+    * (p - this.c)) * 100) / 100);
+  }
```
SerhiiKirik, line 34: Make it more readable
Also, you can not give a type to variables if you put some value to this variable
Typescript knows about this type
But it's up to you
```suggestion
    const semiPerimeter  = (this.a + this.b + this.c) / 2;
    const area = Math.sqrt(p * (p - this.a) * (p - this.b) * (p - this.c);

    return Math.round(area) * 100) / 100);
  }
```

SerhiiKirik (approved):
Well done 🚀

PR #1066, opened 2023-03-15
Solution figure_classes

NazariiAlieksieiev:


src/figuresClasses.ts
```diff
@@ -1,19 +1,76 @@
 export interface Figure {
-
+  shape: 'triangle' | 'circle' | 'rectangle',
+  color: 'red' | 'green' | 'blue',
```
mykhalenych, line 3: the same with color

mykhalenych (changes requested):

mykhalenych (approved):

PR #1071, opened 2023-03-19
Solution-1

MykolaButylkov:


src/figuresClasses.ts
```diff
@@ -1,19 +1,83 @@
...
+    if (Math.max(a, b, c) >= (a + b + c) - Math.max(a, b, c)) {
+      throw new Error(
+        `throws an error: sides ${a}, ${b} and ${c} can't form a triangle`,
+      );
+    }
+  }
+
+  getArea(): number {
+    const { a, b, c } = this;
+    const s = (a + b + c) / 2; // semiperimeter
```
OleziO, line 32: Rename the variable to semiperimeter and remove the comment.

OleziO (changes requested):

src/figuresClasses.ts
MykolaButylkov: akey

VitaliyBondarenko1982 (approved):

PR #1072, opened 2023-03-20
Solution

yana-kotsulym:


package.json
```diff
@@ -25,8 +25,8 @@
     "typescript": "^4.3.5"
   },
   "devDependencies": {
-    "@mate-academy/scripts": "^0.9.2",
     "@mate-academy/eslint-config": "*",
+    "@mate-academy/scripts": "^1.2.8",
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,88 @@
...
+
+    if (hasTooBigSide) {
+      throw new Error('The shape is not a triangle');
+    }
+
+    this.shape = 'triangle';
+  }
 
+  getArea(): number {
+    const s = (this.a + this.b + this.c) / 2;
```
SerhiiKirik, line 39:
```suggestion
    const semiPerimeter = (this.a + this.b + this.c) / 2;
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,88 @@
...
+
```
SerhiiKirik, line 40:
```suggestion
  const square = Math.sqrt(s * (s - this.a) * (s - this.b) * (s - this.c));
```

PR #1073, opened 2023-03-20
Solution

yao157:


slavanosatskyi (approved):
Nice, though your error messages has be to descriptive - "Invalid data" is too abstract

PR #1076, opened 2023-03-22
Solution

IrynaK-a:


src/figuresClasses.js
```diff
@@ -0,0 +1,74 @@
+"use strict";
+Object.defineProperty(exports, "__esModule", { value: true });
+exports.getInfo = exports.Rectangle = exports.Circle = exports.Triangle = void 0;
+var Triangle = /** @class */ (function () {
```
polosanya, line 4: What's the purpose of this file?

src/figuresClasses.ts
```diff
@@ -1,19 +1,93 @@
...
+    } else if (b > a && b > c) {
+      longSide = b;
+      sumOfSides = a + c;
+    } else {
+      longSide = c;
+      sumOfSides = b + a;
+    }
+
+    if (longSide >= sumOfSides) {
+      throw new Error('Sides 1, 2 and 3 can not form a triangle.');
```
polosanya, line 40: It's better to specify why exactly they can't

src/figuresClasses.ts
```diff
@@ -1,19 +1,93 @@
...
+
+  constructor(
+    public color: Color,
+    public width:number,
+    public height:number,
+  ) {
+    this.shape = 'rectangle';
+
+    if (width <= 0 || height <= 0) {
+      throw new Error('Method not implemented.');
```
polosanya, line 82: Why?

polosanya (changes requested):
Please check all errors that you throw and try to make them informative for user, so they know what exactly to fix
image

IvanFesenko (approved):

PR #1080, opened 2023-03-23
solution

MSA1991:


src/figuresClasses.ts
```diff
@@ -1,19 +1,92 @@
...
+  ) {
+    const sideA = this.a;
+    const sideB = this.b;
+    const sideC = this.c;
+
+    const isTriangle = sideA >= sideB + sideC
+      || sideB >= sideA + sideC
+      || sideC >= sideA + sideB;
+
+    if (isValidSides(sideA, sideB, sideC)) {
```
loralevitska, line 35: Maybe `isInvalidSides`. The condition looks strange: if the parties are valid, issue an error

loralevitska (changes requested):
Good solution and optimized. Let's improve variable naming for whole solution

anstsot (approved):
Great job, approved!

PR #1082, opened 2023-03-24
develop

nazar-medushevskyi:


src/figuresClasses.ts
```diff
@@ -1,19 +1,89 @@
...
+  color: Color,
+  getArea(): number,
 }
 
-export class Triangle {
+function roundVolue(volue: number): number {
+  return Math.floor(volue * 100) / 100;
+}
 
+function isCorrectSides(...sides: number[]): void {
```
loralevitska, line 23: This naming is good for variable but for function name should means action. Ex. `check...` `get..` `calculate...` 
https://medium.com/swlh/how-to-better-name-your-functions-and-variables-e962a4ef335b

src/figuresClasses.ts
```diff
@@ -1,19 +1,89 @@
...
+}
 
+export interface Figure {
+  shape: Shape,
+  color: Color,
+  getArea(): number,
 }
 
-export class Triangle {
+function roundVolue(volue: number): number {
```
loralevitska, line 19:
```suggestion
function roundValue(value: number): number {
```

OR

```suggestion
function roundVolume(value: number): number {
```

loralevitska (changes requested):
Almost perfect) Let's improve
If you have any questions, please ask them in fe_chat and tag mentors. It will be faster, because your answers in GitHub will be visible to mentor only after you re-request review

sTorba24 (approved):
Good job!

PR #1083, opened 2023-03-26
Ready

sergik94:


src/figuresClasses.ts
```diff
@@ -1,19 +1,73 @@
...
+  }
 
+  getArea(): number | never {
+    return numberRound(Math.PI * this.radius ** 2);
+  }
 }
 
-export function getInfo(figure) {
+export class Rectangle implements Figure {
+  public shape: string = 'rectangle';
```
VitaliyBondarenko1982, line 54: Create enum or type `Shape` and replace types string in all places.

```
type Shape = 'circle' | 'rectangle' | 'triangle'
```
or
```
enum Shape {
  circle = 'circle',
  rectangle = 'rectangle',
  triangle = 'triangle',
}
```

VitaliyBondarenko1982 (changes requested):
Good

src/figuresClasses.ts
```diff
@@ -1,19 +1,75 @@
...
+  color: string;
+  getArea(): number | never;
+}
 
+function numberRound(x: number): number {
+  return Math.floor(x * 100) / 100;
 }
 
-export class Triangle {
+type Shape = 'triangle' | 'circle' | 'rectangle';
```
tetiana-ivanchenko, line 12: still didn't fix. use enum

tetiana-ivanchenko (changes requested):

src/figuresClasses.ts
```diff
@@ -1,19 +1,80 @@
...
+  return Math.floor(x * 100) / 100;
 }
 
-export class Triangle {
+export enum Shape {
+  triangle = 'triangle',
+  circle = 'circle',
+  rectangle = 'rectangle',
 
 }
```
DarkMistyRoom, line 17: image

src/figuresClasses.ts
```diff
@@ -1,19 +1,80 @@
...
+  public shape = Shape.triangle;
+
+  constructor(
+    public color: string,
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    if (a <= 0 || b <= 0 || c <= 0
+      || Math.max(a, b, c) >= (a + b + c - Math.max(a, b, c))) {
```
DarkMistyRoom, line 29: it isn't a mistake, but you can improve a bit your code. You should check if all arguments are positive in all cases. So you can create a function wich will check it and return the boolean value

src/figuresClasses.ts
```diff
@@ -1,19 +1,80 @@
...
+      throw new Error('the entered data is incorrect');
+    }
+  }
 
+  getArea(): number | never {
```
DarkMistyRoom, line 34:
```suggestion
  getArea(): number {
```

DarkMistyRoom: make it work. try don't use `never` or `undefined` in typescript

DarkMistyRoom (changes requested):

VitaliyBondarenko1982 (approved):

PR #1084, opened 2023-03-26
Solution

OlexandrIstomin:


src/figuresClasses.ts
```diff
@@ -1,19 +1,78 @@
 export interface Figure {
+  shape: 'triangle' | 'circle' | 'rectangle';
+  color: 'red' | 'green' | 'blue';
+  getArea: () => number;
+}
 
+function error(): never {
```
polosanya, line 7:
```suggestion
function showError(): void {
```
Get used to name functions according to these rules
image

polosanya (approved):

PR #1088, opened 2023-03-28
solution

YuriMoreau:


src/figuresClasses.ts
```diff
@@ -1,19 +1,80 @@
...
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    if (a <= 0 || b <= 0 || c <= 0) {
+      throw new Error('Negative value entered');
+    }
+
+    if (a >= b + c || b >= a + c || c >= a + b) {
+      throw new Error('Impossible triangle');
```
polosanya, line 25: It would be great to specify why exactly it's not possible
image

src/figuresClasses.ts
```diff
@@ -1,19 +1,80 @@
...
+    }
+  }
 
-export class Circle {
+  getArea(): number {
+    const p = (this.a + this.b + this.c) / 2;
+    const s = Math.sqrt(p * (p - this.a)
+    * (p - this.b)
+    * (p - this.c));
```
polosanya, line 33:
```suggestion
    const s = Math.sqrt(p * (p - this.a)
      * (p - this.b)
      * (p - this.c));
```

polosanya (changes requested):
Great! Let's just make error messages more descriptive and useful

polosanya (approved):

PR #1092, opened 2023-03-31
Solution

bojkovladislav:


src/figuresClasses.ts
```diff
@@ -1,19 +1,92 @@
...
+const isNotATriangle = (a: number, b: number, c: number): boolean => {
+  const temp = [a, b, c].sort((a, b) => a - b);
+
+  return temp[2] >= temp[0] + temp[1] ? true : false;
+};
+
+const lengthIsLessThanZero = (...params: colorOrNumber[]): boolean => {
+  return params.some((param) => {
+    if (typeof param === "string") {
+      return !param.length ? true : false;
```
SerhiiKirik, line 20: What is return ! operator??
Or > >= <= <
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Expressions_and_Operators
```suggestion
      return !param.length;
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,92 @@
...
+    }
+
+    return !param || param < 0 ? true : false;
```
SerhiiKirik, line 23:
```suggestion
    return !param || param < 0;
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,92 @@
...
+  shape: Shapes;
+  color: Colors;
+  getArea(): number;
 }
 
-export class Triangle {
+const isNotATriangle = (a: number, b: number, c: number): boolean => {
+  const temp = [a, b, c].sort((a, b) => a - b);
+
+  return temp[2] >= temp[0] + temp[1] ? true : false;
```
SerhiiKirik, line 14:
```suggestion
  return temp[2] >= temp[0] + temp[1];
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,92 @@
...
+  return params.some((param) => {
+    if (typeof param === "string") {
+      return !param.length ? true : false;
+    }
+
+    return !param || param < 0 ? true : false;
+  });
+};
+
+const round = (formula: number): number => {
```
SerhiiKirik, line 27:
```suggestion
const getRoundingDown = (formula: number): number => {
```

SerhiiKirik (changes requested):
Great 🔥 
Let's improve your code 🧑‍💻

OleziO (changes requested):

src/figuresClasses.ts
```diff
@@ -1,19 +1,101 @@
...
+type colorOrNumber = Colors | number;
+
 export interface Figure {
+  shape: Shapes;
+  color: Colors;
+  getArea(): number;
+}
+
+class Controller {
+  static lengthIsLessThanZero = (...params: colorOrNumber[]): boolean => {
```
anstsot, line 12: you can move these methods from `Controller` to `Figure` class and use them other classes because you're implementing the base `Figure` class
```suggestion
  shape: Shapes;
  color: Colors;
  getArea(): number;
  
  static lengthIsLessThanZero = (...params: colorOrNumber[]): boolean => {
```

anstsot (approved):
Great job, approved!

PR #1094, opened 2023-04-01
added implementation for classes of figures

Esteban4769:


src/figuresClasses.ts
```diff
@@ -1,19 +1,89 @@
...
+  public shape: Shape = Shape.Circle;
+
+  constructor(public color: Color, public radius: number) {
+    if (radius <= 0) {
+      throw new Error('Radius value cannot be lesser than 1!');
+    }
+  }
 
+  getArea(): number {
+    return Math.floor((Math.PI * (this.radius ** 2)) * 100) / 100;
```
PavloReutskiy, line 65: I rounded in this way all figures, not only Circle because of this condition, but I'm not sure if it's necessary. toFixed() it's like Math.round(), not Math.floor
image

PavloReutskiy (approved):
👍

src/figuresClasses.ts
```diff
@@ -1,19 +1,93 @@
...
+      .filter((side) => side !== longestSide)
+      .reduce((sum, side) => sum + side)) {
+      throw new Error('Side value cannot be lesser than 1!');
+    }
+
+    [this.a, this.b, this.c] = args;
+  }
+
+  getArea(): number {
+    const semiperimeter = 0.5 * (this.a + this.b + this.c);
```
liliia-livandovskaya, line 49:
```suggestion
      const { a, b, c } = this;
	    const semiPerimeter = 0.5 * (a + b + c);
```

PR #1099, opened 2023-04-04
Solution

igorkruz:


src/figuresClasses.ts
```diff
@@ -1,19 +1,79 @@
...
+    const longestSide = this.a >= b + c || this.b >= c + a || this.c >= a + b;
+
+    if (isNotValid || longestSide) {
+      // eslint-disable-next-line max-len
+      throw new Error('The incorrect value passed! The value should be bigger than 0, and the longes side cannot be bigger than the sum of the other 2 sides');
+    }
+  }
+
+  getArea():number {
+    const s = (this.a + this.b + this.c) / 2;
```
AnyaLyaLya, line 31: Maybe you can choose better values name

AnyaLyaLya (approved):

src/figuresClasses.ts
```diff
@@ -1,19 +1,79 @@
...
 
+export interface Figure {
+  shape: Shape;
+  color: Color;
+  getArea: () => number;
 }
 
-export class Triangle {
+export class Triangle implements Figure {
+  shape:Shape = 'triangle';
```
liliia-livandovskaya, line 12:
```suggestion
  shape: Shape = 'triangle';
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,79 @@
...
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    const isNotValid = a <= 0 || b <= 0 || c <= 0;
+    const longestSide = this.a >= b + c || this.b >= c + a || this.c >= a + b;
+
+    if (isNotValid || longestSide) {
+      // eslint-disable-next-line max-len
+      throw new Error('The incorrect value passed! The value should be bigger than 0, and the longes side cannot be bigger than the sum of the other 2 sides');
```
liliia-livandovskaya, line 25:
```suggestion
      throw new Error('The incorrect value passed! The value should be bigger than 0,`
       + `and the longes side cannot be bigger than the sum of the other 2 sides');
```

liliia-livandovskaya (approved):

src/figuresClasses.ts
igorkruz: I leave that as I had because the suggested variant also gives the max-length error
sugested

PR #1101, opened 2023-04-04
Created figures with calculated area

Evgenb000:


VenomMB (approved):
GJ!!!

src/figuresClasses.ts
```diff
@@ -1,19 +1,102 @@
...
+  readonly shape = 'triangle';
 
-}
+  readonly color: Color;
+
+  readonly a: number;
+
+  readonly b: number;
+
+  readonly c: number;
```
liliia-livandovskaya, line 19:
```suggestion
  readonly color: Color;
  readonly a: number;
  readonly b: number;
  readonly c: number;
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,102 @@
...
+    }
+
+    this.color = color;
+    this.a = a;
+    this.b = b;
+    this.c = c;
+  }
+
+  getArea(): number {
+    const s = (this.a + this.b + this.c) / 2;
```
liliia-livandovskaya, line 37:
```suggestion
    const semiPerimeter = (this.a + this.b + this.c) / 2;
```

src/figuresClasses.ts
liliia-livandovskaya: you can use `public` instead of `readonly`

liliia-livandovskaya (approved):

PR #1110, opened 2023-04-05
add solution

Pozdnya:


src/figuresClasses.ts
```diff
@@ -1,19 +1,91 @@
...
+    public b: number,
+    public c: number,
+  ) {
+    this.shape = Shape.Triangle;
+
+    if (this.a <= 0 || this.b <= 0 || this.c <= 0) {
+      throw new Error('One of the arguments are negative');
+    }
+
+    const longestSide = Math.max(this.a, this.b, this.c);
```
liliia-livandovskaya, line 29:
```suggestion
    const { a, b, c } = this;
    const longestSide = Math.max(a, b, c);
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,91 @@
...
+  Rectangle = 'rectangle',
+}
+export interface Figure {
+  shape: Shape,
+  color: Color,
+  getArea: () => number;
 }
 
 export class Triangle {
+  shape: Shape;
```
liliia-livandovskaya, line 15:
```suggestion
  public shape: Shape = Shape.Triangle;
```

liliia-livandovskaya (changes requested):
Great job! Lets work on improvement

liliia-livandovskaya (approved):

PR #1115, opened 2023-04-05
figures classes

ofenchuk:
for some reason it rounds up 0.9 to 1 in one test

### DIFF
--- a/src/figuresClasses.ts
+++ b/src/figuresClasses.ts
@@ -1,19 +1,76 @@
 export interface Figure {
-
+  shape: string,
+  color: string,
+  getArea: () => number,
 }
 
-export class Triangle {
+export class Triangle implements Figure {
+  shape: string = 'triangle';
 
+  constructor(
+    public color: string,
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    if (a <= 0 || c <= 0 || b <= 0) {
+      throw new Error('wrong side(s)');
+    }
+
+    const checkLongest = [this.a, this.b, this.c]
+      .sort((side1, side2) => side2 - side1);
+
+    if (checkLongest[0] >= checkLongest[1] + checkLongest[2]) {
+      throw new Error('unappropriated side');
+    }
+  }
+
+  getArea(): number {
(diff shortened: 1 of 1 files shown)


src/figuresClasses.ts
```diff
@@ -1,19 +1,76 @@
...
+    if (radius <= 0) {
+      throw new Error('wrong radius');
+    }
+  }
 
-export class Rectangle {
+  getArea(): number {
+    const result = Math.PI * (this.radius * this.radius);
 
+    return Math.round(result * 100) / 100;
```
KosolapovAnatolii, line 52: You can use Math.floor to fix bag whis wrong circle's square

KosolapovAnatolii (changes requested):
Very good!
Try to use Math.floor. It helped me pass that test

src/figuresClasses.ts
ofenchuk: yeah, it worked, thanks

src/figuresClasses.ts
```diff
@@ -1,19 +1,76 @@
...
+
+    if (checkLongest[0] >= checkLongest[1] + checkLongest[2]) {
+      throw new Error('unappropriated side');
+    }
+  }
+
+  getArea(): number {
+    const s = (this.a + this.b + this.c) / 2;
+    const result = Math.sqrt(s * (s - this.a)
+      * (s - this.b) * (s - this.c));
```
SerhiiKirik, line 31:
```suggestion
    const result = Math.sqrt(
      s * (s - this.a) * (s - this.b) * (s - this.c)
    );
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,76 @@
...
+    if (radius <= 0) {
+      throw new Error('wrong radius');
+    }
+  }
 
-export class Rectangle {
+  getArea(): number {
+    const result = Math.PI * (this.radius * this.radius);
 
+    return Math.floor(result * 100) / 100;
```
SerhiiKirik, line 52: move this to the separate functuion

SerhiiKirik (approved):
Well done 🚀

PR #1119, opened 2023-04-05
solution

glots-a:


src/figuresClasses.ts
```diff
@@ -1,19 +1,81 @@
...
+    const semiPerimetr = 0.5 * (this.a + this.b + this.c);
 
+    const area = Math.sqrt(
+      semiPerimetr
+      * (semiPerimetr - this.a)
+      * (semiPerimetr - this.b)
+      * (semiPerimetr - this.c),
+    );
+
+    return +area.toFixed(2);
```
Nazarii98, line 38: +area.toFixed(2)

when you want to change your type to a number, better to use object Number(); It is better for reading.

Nazarii98 (approved):
Good solution!

src/figuresClasses.ts
glots-a: 👍

liliia-livandovskaya, line 38:
```suggestion
    return Number(area.toFixed(2));
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,81 @@
...
+      throw new Error('Some of the lengths are lover or equal 0');
+    }
+
+    if (this.a >= b + c || this.b >= a + c || this.c >= a + b) {
+      throw new Error('Side should be less then sum of each other sides');
+    }
+  }
+
+  getArea(): number {
+    const semiPerimetr = 0.5 * (this.a + this.b + this.c);
```
liliia-livandovskaya, line 29:
```suggestion
    const { a, b, c } = this;
    const semiPerimeter = 0.5 * (a + b + c);
```

liliia-livandovskaya (approved):

PR #1121, opened 2023-04-05
add solution

Vlad-Lischenovskiy:


src/figuresClasses.ts
```diff
@@ -1,19 +1,82 @@
...
-}
+  constructor(
+    public color: Color,
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    if ((this.a + this.b) <= this.c
+    || (this.a + this.c) <= this.b
+    || (this.b + this.c) <= this.a) {
```
Anna-Kuzmenko, line 21: you can write a, b, c without this.

src/figuresClasses.ts
```diff
@@ -1,19 +1,82 @@
...
 
-export class Rectangle {
+export class Circle implements Figure {
+  shape: Shape = 'circle';
 
+  constructor(
+    public color: Color,
+    public radius: number,
+  ) {
+    if (this.radius <= 0) {
```
Anna-Kuzmenko, line 52: 'this' is not necessarily

Anna-Kuzmenko (approved):

src/figuresClasses.ts
```diff
@@ -1,19 +1,82 @@
...
+    if (this.a <= 0 || this.b <= 0 || this.c <= 0) {
+      throw new Error(
+        'we cannot calculate the area since the sides are less than 0',
+      );
+    }
+  }
 
-export class Circle {
+  getArea(): number {
+    const findArea = (this.a + this.b + this.c) / 2;
```
liliia-livandovskaya, line 35:
```suggestion
    const { a, b, c } = this;
    const semiPerimeter = (a + b + c) / 2;
```

liliia-livandovskaya (approved):

PR #1126, opened 2023-04-05
Solution

vksazonov:


iHateAntichrist (approved):
good job!

src/figuresClasses.ts
```diff
@@ -1,19 +1,76 @@
...
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    if (a <= 0
+      || b <= 0
+      || c <= 0
+      || a + b <= c
+      || a + c <= b
+      || c + b <= a) {
```
SerhiiKirik, line 23: you can handle it like that
```suggestion
    if (a >= (b + c) || b >= (a + c) || c >= (b + a)) {
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,76 @@
...
+      throw new Error('Please enter the correct value of the sides');
+    }
+  }
+
+  getArea(): number {
+    const halfP: number = (this.a + this.b + this.c) / 2;
 
+    return Number(Math.sqrt(
+      halfP * (halfP - this.a) * (halfP - this.b) * (halfP - this.c),
+    ).toFixed(2));
```
SerhiiKirik, line 33: Why do you need toFixed if you know how to do it with * 100 / 100
Remove it rounding to the separate function & reuse in all cases
also, split it with variable
```suggestion
    return getRounding(square);
```

SerhiiKirik (changes requested):
Well done 🚀 
Let's make some improvements 🧑‍💻

PR #1128, opened 2023-04-05
Solution

MylaCrystal:


asmd564 (approved):

src/figuresClasses.ts
```diff
@@ -1,19 +1,76 @@
...
+    }
+
+    if (a + b <= c || b + c <= a || c + a <= b) {
+      throw new Error('Sides do not form a triangle');
+    }
+  }
 
-export class Circle {
+  getArea(): number {
+    const s = (this.a + this.b + this.c) / 2;
```
liliia-livandovskaya, line 32:
```suggestion
    const { a, b, c } = this;
    const semiPerimeter = (a + b + c) / 2;
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,76 @@
-export interface Figure {
+enum Color {
+  Red = 'red',
+  Green = 'green',
+  Blue = 'blue'
+}
 
+export interface Figure {
+  shape: string;
```
liliia-livandovskaya, line 8: add type for `shape` as well

liliia-livandovskaya (changes requested):
Great job! Lets work on improvement

src/figuresClasses.ts
liliia-livandovskaya: create type for shape 
`type Shape = 'triangle' | 'circle' | 'rectangle';`
and use it in the code

liliia-livandovskaya (changes requested):
Pay attention to the comments

src/figuresClasses.ts
liliia-livandovskaya:
```suggestion
  shape: Shape;
```
not fixed

liliia-livandovskaya (changes requested):
Make sure you have fixed everything before re-apply for review

liliia-livandovskaya (approved):

PR #1132, opened 2023-04-05
solution

IvanTkachenko3301:


DimaDvm (approved):

src/figuresClasses.ts
```diff
@@ -1,19 +1,92 @@
...
+    }
 
+    this.color = color;
+    this.a = a;
+    this.b = b;
+    this.c = c;
+  }
+
+  getArea(): number {
+    const p = (this.a + this.b + this.c) / 2;
```
liliia-livandovskaya, line 34:
```suggestion
    const { a, b, c } = this;
    const semiPerimeter = (a + b + c) / 2;
```

liliia-livandovskaya (changes requested):
Pay attention to the comments

src/figuresClasses.ts
IvanTkachenko3301: Thanks for your comments

PR #1136, opened 2023-04-05
Definitely not the best solution

vinnykalisa:


AlexRog01 (approved):
Well done! The only thing i would add is a function for rounding, because you do it three times in different places. But it`s up to you.

package.json
```diff
@@ -25,8 +25,8 @@
     "typescript": "^4.3.5"
   },
   "devDependencies": {
-    "@mate-academy/scripts": "^0.9.2",
     "@mate-academy/eslint-config": "*",
+    "@mate-academy/scripts": "^1.2.8",
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,84 @@
...
+  getArea(): number {
+    const halfOfAPerimeter = (this.a + this.b + this.c) / 2;
+    const square = Math.floor(
+      Math.sqrt(halfOfAPerimeter
+        * (halfOfAPerimeter - this.a)
+        * (halfOfAPerimeter - this.b)
+        * (halfOfAPerimeter - this.c)) * 100,
+    ) / 100;
+
+    return square;
```
SerhiiKirik, line 40: DRY - don't repeat yourself
move this rounding to the separate function
```suggestion
    const square = Math.sqrt(halfOfAPerimeter
        * (halfOfAPerimeter - this.a)
        * (halfOfAPerimeter - this.b)
        * (halfOfAPerimeter - this.c));

    return getRounding(square);
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,84 @@
+type Shape = 'triangle' | 'circle' | 'rectangle';
+type Color = 'red' | 'green' |'blue';
+
 export interface Figure {
+  shape: Shape,
+  color: Color,
 
+  getArea: () => number;
```
SerhiiKirik, line 8:
```suggestion
  color: Color,
  getArea: () => number;
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,84 @@
...
+    public width: number,
+    public height: number,
+  ) {
+    if (width <= 0 || height <= 0) {
+      // eslint-disable-next-line
+      throw new Error(`Width (${width}) and heigth (${height}) must be greater than 0`);
+    }
+  }
 
+  getArea(): number {
```
SerhiiKirik, line 75: typeScript know about that
```suggestion
  getArea() {
```

SerhiiKirik (changes requested):
Well done 🚀 
Let's make some improvements 🧑‍💻

liliia-livandovskaya (approved):

PR #1138, opened 2023-04-06
solution

makandriival:


src/figuresClasses.ts
```diff
@@ -1,19 +1,99 @@
...
+
+  a: number;
+
+  b: number;
+
+  c: number;
 
+  constructor(color: Color, a: number, b: number, c: number) {
+    const isValidProps: boolean = a > 0 && b > 0 && c > 0;
+    const isValidSide: boolean = a + b > c && b + c > a && a + c > b;
```
SerhiiKirik, line 23: typeScript already know about this type
```suggestion
    const isValidProps = a > 0 && b > 0 && c > 0;
    const isValidSide = a + b > c && b + c > a && a + c > b;
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,99 @@
...
+    this.color = color;
+    this.a = a;
+    this.b = b;
+    this.c = c;
+  }
+
+  getArea(): number {
+    const s: number = (this.a + this.b + this.c) / 2;
+    const area: number
+      = Math.sqrt(s
```
SerhiiKirik, line 38:
```suggestion
    const area: number = Math.sqrt(s
```

SerhiiKirik (approved):
Great 🔥

PR #1139, opened 2023-04-06
Solution

TanyaS133:


package.json
```diff
@@ -25,8 +25,8 @@
     "typescript": "^4.3.5"
   },
   "devDependencies": {
-    "@mate-academy/scripts": "^0.9.2",
     "@mate-academy/eslint-config": "*",
+    "@mate-academy/scripts": "^1.2.8",
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,87 @@
...
+    public color: Color,
+    public radius: number,
+  ) {
+    if (radius <= 0) {
+      throw new Error('The radius is less than or equal to 0');
+    }
+  }
+
+  getArea(): number {
+    return Math.floor((Math.PI * this.radius ** 2) * 100) / 100;
```
SerhiiKirik, line 60: make a separate function & reuse in all cases
also, split this with variable
```suggestion
    return getRounding(...);
```

SerhiiKirik (approved):
Great 🔥

PR #1140, opened 2023-04-06
figuresClasses solution

TolikSerhiychuk:


src/figuresClasses.ts
```diff
@@ -1,19 +1,72 @@
...
+    }
+  }
 
+  getArea(): number {
+    const semiPer = (this.a + this.b + this.c) / 2;
+    const area = Math.sqrt(
+      semiPer * (semiPer - this.a) * (semiPer - this.b) * (semiPer - this.c),
+    );
+
+    return Math.floor(area * 100) / 100;
```
SerhiiKirik, line 31: DRY
You make it 3 times, so, move to a separate function & reuse it
```suggestion
    return getRounding(area);
```

SerhiiKirik (approved):
Well done 🚀

PR #1141, opened 2023-04-06
Solution

Zibi95:


src/figuresClasses.ts
```diff
@@ -1,19 +1,79 @@
...
 
+    return Math.floor(area * 100) / 100;
+  }
 }
 
-export function getInfo(figure) {
+export function getInfo(figure:Figure):string {
+  const { color, shape } = figure;
 
+  return `A ${color} ${shape} - ${figure.getArea()}`;
```
Radoslaw-Czerniawski, line 78: A question from me, if you would destructure getArea on line 76 and would call this here alone not as method on figure object, would it work? If not what you would need to do to make it work?

Radoslaw-Czerniawski (approved):
Good job :D

src/figuresClasses.ts
Zibi95: No it wouldn't because 'this' keyword would refer to undefined or window object (depends from strict mode). To make it work i can declare getArea function with arrow function or use call() method with proper this.

PR #1143, opened 2023-04-06
solution

Goto-Jesus:


src/figuresClasses.ts
```diff
@@ -1,19 +1,108 @@
...
+      throw new Error('Incorrect proportion of the sides of a triangle!');
+    }
+  }
+
+  shape: Shape = 'triangle';
+
+  getArea(): number {
+    const p = (this.a + this.b + this.c) / 2;
+    const S = Math.sqrt(p * (p - this.a) * (p - this.b) * (p - this.c));
+    const normalizeArea: string = S.toFixed(3);
```
SerhiiKirik, line 44: you make this normalize 3 times
So, make a separate function for this & reuse in all cases

SerhiiKirik (changes requested):
Well done 🚀 
Let's improve your solution 🧑‍💻

src/figuresClasses.ts
Goto-Jesus: done, thx for advice)))

src/figuresClasses.ts
```diff
@@ -1,19 +1,104 @@
...
+
+    if ((a + b) <= c || (b + c) <= a || (c + a) <= b) {
+      throw new Error('Incorrect proportion of the sides of a triangle!');
+    }
+  }
+
+  shape: Shape = 'triangle';
+
+  getArea(): number {
+    const p = (this.a + this.b + this.c) / 2;
```
SerhiiKirik, line 44:
```suggestion
    const perimeter = (this.a + this.b + this.c) / 2;
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,104 @@
...
+    const S = Math.sqrt(p * (p - this.a) * (p - this.b) * (p - this.c));
 
+    return normalize(S);
```
SerhiiKirik, line 47:
```suggestion
    const square = Math.sqrt(p * (p - this.a) * (p - this.b) * (p - this.c));

    return normalize(square);
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,104 @@
...
+
+    if (width <= 0 || height <= 0) {
+      throw new Error(`Wrong size - width: ${width}, height: ${height}!!!`);
+    }
+  }
+
+  shape: Shape = 'rectangle';
+
+  getArea(): number {
+    const S = this.height * this.width;
```
SerhiiKirik, line 94: why s?
Try to give more clearly name

PR #1144, opened 2023-04-06
figures classes solution

AP-Solution:


src/figuresClasses.ts
```diff
@@ -1,19 +1,71 @@
...
+      throw new Error('These sides cannot form a triangle.');
+    }
+  }
 
-export class Circle {
+  getArea(): number {
+    const p = (this.a + this.b + this.c) / 2;
+    const area = Math.sqrt(p * (p - this.a) * (p - this.b) * (p - this.c));
 
+    return Math.floor(area * 100) / 100;
```
SerhiiKirik, line 29: DRY
You make it 3 times, so, move to the separate function & reuse it

SerhiiKirik (approved):
Well done 🚀

PR #1151, opened 2023-04-11
solution

DanielSorokowski:


src/figuresClasses.ts
```diff
@@ -1,19 +1,84 @@
...
+      throw new Error('all values should be positive');
+    }
+
+    if (
+      a + b <= c
+      || a + c <= b
+      || b + c <= a
+    ) {
+      throw new Error('Impossible to create triangle from that values');
+    }
```
solaryasha, line 30: All three classes have similar checks. Maybe, let's define a private method that does this check for us ?

```typescript
private checkSides() {
  // write your code here
}
```

solaryasha (changes requested):

src/figuresClasses.ts
```diff
@@ -1,19 +1,98 @@
 export interface Figure {
+  color: string;
+  shape: string;
```
solaryasha, line 3:
```suggestion
  shape: Shape;
```

solaryasha: Let's restrict the value of the `shape` property

solaryasha (changes requested):

PR #1156, opened 2023-04-18
solution

Nikolaymelni:


polosanya (changes requested):
Almost done! Let's try to optimize your solution)

src/figuresClasses.ts
```diff
@@ -1,19 +1,73 @@
+type Shape = 'triangle' | 'circle' | 'rectangle';
+type Color = 'red' | 'green' | 'blue';
+
 export interface Figure {
+  shape: Shape;
+  color: Color;
+  getArea: () => number;
+}
 
+function checkSides(a: number, b = 1, c = 1): void {
```
polosanya, line 10: Better make all the parameters 0 by default to certainly throw the error if nothing was passed.
Also, you can get the array of arguments using rest syntax `(...args)`
https://javascript.info/rest-parameters-spread

polosanya (approved):
Well done!

PR #1158, opened 2023-04-19
solution v1

bjeziorek:


ghostargg (approved):

kasiakubisiak (changes requested):

src/figuresClasses.ts
```diff
@@ -1,19 +1,110 @@
...
-export class Rectangle {
+export class Circle implements Figure {
+  shape: Shape;
+
+  color: Color;
+
+  radius: number;
 
+  getArea(): number {
+    return Math.floor(((2 * Math.PI * this.radius * this.radius) / 2)
```
d-wasilewski, line 69: No need to divide the area by 2, directly calculate the area of the circle

src/figuresClasses.ts
```diff
@@ -1,19 +1,110 @@
...
+  shape: Shape;
+
+  color: Color;
+
+  width: number;
+
+  height: number;
+
+  getArea(): number {
+    return +((this.width * this.height)
```
d-wasilewski, line 93: Use Math.floor instead of toFixed

src/figuresClasses.ts
```diff
@@ -1,19 +1,110 @@
...
+  }
+
+  constructor(color: Color, width: number, height: number) {
+    if (width <= 0 || height <= 0) {
+      throw new Error('all values have to be greater than 0');
+    }
+    this.shape = Shape.rectangle;
+    this.color = Color[color];
+    this.width = width;
+    this.height = height;
```
d-wasilewski, line 104: checklist 5

src/figuresClasses.ts
```diff
@@ -1,19 +1,110 @@
...
 
-export function getInfo(figure) {
+export class Rectangle implements Figure {
+  shape: Shape;
+
+  color: Color;
+
+  width: number;
+
+  height: number;
```
d-wasilewski, line 90: do you need these here?

src/figuresClasses.ts
```diff
@@ -1,19 +1,110 @@
...
+
+  c: number;
+
+  getArea(): number {
+    const half = (this.a
+      + this.b
+      + this.c)
+      / 2;
+
+    return Math.floor(+((Math.sqrt(
```
d-wasilewski, line 35: is this `+` necessary here?

src/figuresClasses.ts
```diff
@@ -1,19 +1,103 @@
...
+
+    if (a + b <= c || a + c <= b || b + c <= a) {
+      throw new Error('not triangle');
+    }
+
+    this.shape = Shape.triangle;
+    this.color = color;
+    this.a = a;
+    this.b = b;
+    this.c = c;
```
d-wasilewski, line 52: this is also not necessary, check checklist

d-wasilewski (changes requested):

src/figuresClasses.ts
d-wasilewski: except for the shape

src/figuresClasses.ts
```diff
@@ -1,19 +1,92 @@
...
+      throw new Error('not triangle');
+    }
 
+    this.shape = Shape.triangle;
+  }
 }
 
-export class Rectangle {
+export class Circle implements Figure {
+  shape: Shape;
```
d-wasilewski, line 51: You didn't need to remove it completety but now you have it repeated twice, here and in the constructor. If you want to keep it here you can remove it from the constructor. Or the other way around, just don't repeat it

d-wasilewski (changes requested):

d-wasilewski (approved):

PR #1159, opened 2023-04-19
solution

ghostargg:


src/figuresClasses.ts
```diff
@@ -1,19 +1,78 @@
+type Shape = 'triangle' | 'circle' | 'rectangle';
+
 export interface Figure {
+  color: string;
```
d-wasilewski, line 4: you can also create a type for the color

src/figuresClasses.ts
```diff
@@ -1,19 +1,78 @@
+type Shape = 'triangle' | 'circle' | 'rectangle';
```
d-wasilewski, line 1: you can create an enum for that. This will allow you to also assign variables with this enum instead of individual fields `(shape = Shape.circle;)`

src/figuresClasses.ts
```diff
@@ -1,19 +1,78 @@
...
+
 export interface Figure {
+  color: string;
+  shape: Shape;
+  r? : number;
+  a? : number;
+  b? : number;
+  c? : number;
+  width? : number;
+  height?: number;
```
d-wasilewski, line 11: those fields shouldnt be here, just shape color and getArea

src/figuresClasses.ts
```diff
@@ -1,19 +1,78 @@
...
-}
+  constructor(public color: string, public a: number,
+    public b: number, public c: number) {
+    if (a <= 0 || b <= 0 || c <= 0) {
+      throw new Error('All sides have to be higher than 0');
+    }
+
+    if (a + b <= c) {
+      throw new Error(`Sides ${a}, ${b} and ${c} can't form a traingle`);
+    }
```
d-wasilewski, line 27: Check for all the possible cases in this if statement

src/figuresClasses.ts
```diff
@@ -1,19 +1,78 @@
...
+    ) / 100;
+  }
 }
 
-export class Rectangle {
+export class Circle implements Figure {
+  shape: Shape = 'circle';
+
+  constructor(public color: string, public r: number) {
+    if (r <= 0) {
```
d-wasilewski, line 47: rename r => radius

src/figuresClasses.ts
```diff
@@ -1,19 +1,78 @@
...
+  shape: Shape = 'rectangle';
+
+  constructor(public color: string, public width: number,
+    public height: number) {
+    if (width <= 0 || height <= 0) {
+      throw new Error('All sides have to be positive');
+    }
+
+    this.width = width;
+    this.height = height;
```
d-wasilewski, line 68: do you need this?

d-wasilewski (changes requested):

src/figuresClasses.ts
```diff
@@ -1,19 +1,81 @@
...
+
+    if (a + b <= c
+      || b + c <= a
+      || a + c <= b) {
+      throw new Error(`Sides ${a}, ${b} and ${c} can't form a traingle`);
+    }
+
+    this.a = a;
+    this.b = b;
+    this.c = c;
```
d-wasilewski, line 37: In TypeScript, when you declare a constructor parameter with an access modifier TypeScript implicitly creates a property on the class with the same name as the parameter and assigns the parameter value to that property.

src/figuresClasses.ts
```diff
@@ -1,19 +1,81 @@
...
+  }
 }
 
-export class Rectangle {
+export class Circle implements Figure {
+  shape = Shape.Circle;
+
+  constructor(public color: Color, public radius: number) {
+    if (radius <= 0) {
+      throw new Error('Radius have to be positive');
```
d-wasilewski, line 54:
```suggestion
      throw new Error('Radius has to be positive');
```

d-wasilewski (changes requested):

src/figuresClasses.ts
```diff
@@ -1,19 +1,73 @@
...
 
-export class Rectangle {
+export class Circle implements Figure {
+  shape = Shape.Circle;
 
+  constructor(public color: Color, public radius: number) {
+    if (radius <= 0) {
+      throw new Error('Radius has to be positive');
+    }
+    this.radius = radius;
```
d-wasilewski, line 48: same thing

d-wasilewski (changes requested):

PR #1160, opened 2023-04-19
solution

MikiOOOO:


ghostargg (approved):

src/figuresClasses.ts
```diff
@@ -1,19 +1,69 @@
...
 
+export interface Figure {
+  shape: Shape,
+  color: Color,
+  getArea(): number,
 }
 
-export class Triangle {
+export class Triangle implements Figure {
+  shape: Shape= 'triangle';
```
kasiakubisiak, line 11: Let's create the enum `Shape` with the shape values, so we could restrict those values that are being passed to the shape property

src/figuresClasses.ts
```diff
@@ -1,19 +1,69 @@
...
 
-export class Circle {
+    return Math.floor(area * 100) / 100;
+  }
 
+  private isValidTriangle(): boolean {
+    const [a, b, c] = [this.a, this.b, this.c].sort((x, y) => x - y);
+
+    return a + b > c;
+  }
```
kasiakubisiak, line 36: Maybe lets create a method called for example "getSortedSides" and make it return the sides sorted in ascending order?

```suggestion
  private isValidTriangle(): boolean {
    const [a, b, c] = this.getSortedSides();
    return a + b > c;
  } 
```

kasiakubisiak (changes requested):

d-wasilewski (approved):

PR #1161, opened 2023-04-20
solution

Dka888:


danielpopek94 (approved):

src/figuresClasses.ts
```diff
@@ -1,19 +1,106 @@
...
+  color: string;
+
+  a: number;
+
+  b: number;
+
+  c: number;
 
-export class Circle {
+  private s: number;
```
kasiakubisiak, line 19: what is 's'? I suggest using more descriptive variable name

src/figuresClasses.ts
```diff
@@ -1,19 +1,106 @@
...
 
+  constructor(color: string, a: number, b: number, c: number) {
+    this.shape = 'triangle';
```
kasiakubisiak, line 22: Let's create the enum `Shape` with the values so that it could restrict the values that passed to the shape property

src/figuresClasses.ts
```diff
@@ -1,19 +1,106 @@
-export interface Figure {
 
+export interface Figure {
+  shape: string;
```
kasiakubisiak, line 3:
```suggestion
  shape: Shape;
```

kasiakubisiak (changes requested):

src/figuresClasses.ts
```diff
@@ -1,19 +1,107 @@
-export interface Figure {
+type Shape ='triangle' | 'circle' | 'rectangle';
 
+export interface Figure {
+  shape: Shape;
+  color: string;
```
d-wasilewski, line 5: create a type for this

src/figuresClasses.ts
```diff
@@ -1,19 +1,107 @@
-export interface Figure {
+type Shape ='triangle' | 'circle' | 'rectangle';
 
+export interface Figure {
+  shape: Shape;
```
d-wasilewski, line 4: create an enum for shape and use it later like shape = Shape.rectangle; (instead of `type`)

src/figuresClasses.ts
```diff
@@ -1,19 +1,107 @@
...
+
+  c: number;
 
-export class Circle {
+  private halfSides: number;
 
+  constructor(color: string, a: number, b: number, c: number) {
+    this.shape = 'triangle';
+    this.color = color;
+    this.a = a;
```
d-wasilewski, line 25: In TypeScript, when you declare a constructor parameter with an access modifier TypeScript implicitly creates a property on the class with the same name as the parameter and assigns the parameter value to that property.

src/figuresClasses.ts
```diff
@@ -1,19 +1,107 @@
...
+    this.b = b;
+    this.c = c;
+    this.halfSides = (this.a + this.b + this.c) / 2;
```
d-wasilewski, line 28: just create a variable for this, this doesnt need to be a class field

src/figuresClasses.ts
```diff
@@ -1,19 +1,107 @@
...
+  }
+
+  private checkSides(): void {
+    if (this.a + this.b <= this.c
+      || this.b + this.c <= this.a
+      || this.c + this.a <= this.b) {
+      throw new Error('That is not a triagnle!');
+    }
+
+    if (this.a <= 0 || this.b <= 0 || this.c <= 0) {
```
d-wasilewski, line 39: just run this code inside the constructor, no need to create a function as you won't reuse it anyway.
Also, you don't have to use this.a, you can use just `a` as you want to only check the coming variables

d-wasilewski (changes requested):

src/figuresClasses.ts
```diff
@@ -1,19 +1,72 @@
...
+    }
+
+    if (a <= 0 || b <= 0 || c <= 0) {
+      throw new Error('Side value should be more than 0');
+    }
+  }
+
+  private halfSides = (this.a + this.b + this.c) / 2;
 
+  getArea(): number {
```
d-wasilewski, line 33: It's a method property used to calculate something inside of it, the class doesnt need to have this itself
```suggestion
  getArea(): number {
    const halfSides = (this.a + this.b + this.c) / 2;
```

d-wasilewski (changes requested):

d-wasilewski (approved):

PR #1165, opened 2023-04-20
Solution

Vaskovics:


src/figuresClasses.ts
```diff
@@ -1,19 +1,87 @@
...
+    }
+  });
+
+  if (params.length === 3) {
+    if (params[0] + params[1] <= params[2]
+      || params[0] + params[2] <= params[1]
+      || params[1] + params[2] <= params[0]) {
+      throw new Error(`The longest side of a triangle has to be'
+      + '<= than a sum of two others`);
+    }
```
SanyaBratashchuk, line 20: create separate validation util for specific rules for each Shape

SanyaBratashchuk (changes requested):

src/figuresClasses.ts
```diff
@@ -1,19 +1,89 @@
...
-export class Triangle {
+const lengthValidator: (...args: number[]) => void = (...params: number[]) => {
+  params.forEach((parm) => {
+    if (parm <= 0) {
+      throw new Error('Length should be > 0! Thank you!');
+    }
+  });
+};
+
+const shapeValidator: (...args: number []) => void = (...params: number[]) => {
```
SanyaBratashchuk, line 15:
```suggestion
const trianglesSdesValidator: (...args: number []) => void = (...params: number[]) => {
```

SanyaBratashchuk (approved):

PR #1166, opened 2023-04-21
first solution

Ozwiena:


danielpopek94 (approved):

src/figuresClasses.ts
```diff
@@ -1,19 +1,76 @@
 export interface Figure {
-
+  shape: 'triangle' | 'circle' |'rectangle',
+  color: 'red' | 'green' | 'blue',
```
d-wasilewski, line 3: create a type for this

src/figuresClasses.ts
```diff
@@ -1,19 +1,76 @@
 export interface Figure {
-
+  shape: 'triangle' | 'circle' |'rectangle',
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,76 @@
...
+      throw new Error('This is not a triangle!');
+    }
+
+    if (this.a <= 0 || this.b <= 0 || this.c <= 0) {
+      throw new Error('This is not a triangle!');
+    }
+
+    this.color = color;
+    this.a = a;
+    this.b = b;
```
d-wasilewski, line 24: In TypeScript, when you declare a constructor parameter with an access modifier TypeScript implicitly creates a property on the class with the same name as the parameter and assigns the parameter value to that property.

src/figuresClasses.ts
```diff
@@ -1,19 +1,76 @@
...
-export class Triangle {
+export class Triangle implements Figure {
+  constructor(public color: 'red' | 'green' | 'blue', public a: number,
+    public b: number, public c: number, public shape: 'triangle' = 'triangle') {
+    const theLongest = Math.max(a, b, c);
+    const sides = [a, b, c];
+    const shorterSides = sides.filter((x) => x !== theLongest);
 
-}
+    if (theLongest >= shorterSides[0] + shorterSides[1]) {
```
d-wasilewski, line 14: it looks more complicated and complex than just checking these 3 cases manually

d-wasilewski: maybe you can try to sort it first and then you will know which one is longed, and you will also have shorter sides

src/figuresClasses.ts
```diff
@@ -1,19 +1,76 @@
...
+      Math.sqrt(p * (p - this.a) * (p - this.b) * (p - this.c)) * 100,
+    ) / 100;
+  }
 }
 
-export class Rectangle {
+export class Circle implements Figure {
+  constructor(public color: 'red' | 'green' | 'blue',
+    public r: number, public shape: 'circle' = 'circle') {
+    if (this.r <= 0) {
```
d-wasilewski, line 41: rename r to radius

src/figuresClasses.ts
```diff
@@ -1,19 +1,76 @@
...
+    this.a = a;
+    this.b = b;
+  }
+
+  getArea(): number {
+    return Math.floor(this.a * this.b * 100) / 100;
+  }
+}
 
+export function getInfo(figure: Triangle | Circle | Rectangle): string {
```
d-wasilewski, line 74: figure: Figure

d-wasilewski (approved):

PR #1167, opened 2023-04-22
develop

nata-prog123:


src/figuresClasses.ts
```diff
@@ -1,19 +1,70 @@
 export interface Figure {
-
+  shape: string;
+  color: string;
+  getArea(): number;
 }
 
-export class Triangle {
+export class Triangle implements Figure {
+  shape: string = 'triangle';
```
TarasHoliuk, line 8: shouldn't be a `string` type. you have to create the corresponding enum for the shape

src/figuresClasses.ts
```diff
@@ -1,19 +1,70 @@
...
 
-}
+  constructor(
+    public color: string,
```
TarasHoliuk, line 11: shouldn't be a string type. you have to create the corresponding enum or union type for the color

src/figuresClasses.ts
```diff
@@ -1,19 +1,70 @@
...
+    public radius: number,
+  ) {
+    if (radius <= 0) {
+      throw new Error('Radius must be greater than zero!');
+    }
+  }
 
+  getArea(): number {
+    return Math.floor(Math.PI * this.radius * this.radius * 100) / 100;
```
TarasHoliuk, line 46: you are rounding the number to hundredths 3 times - it's better to move it to a separate function

TarasHoliuk (changes requested):

tetiana-ivanchenko (approved):

PR #1168, opened 2023-04-23
solution

sylcym:


DanielSorokowski (approved):

src/figuresClasses.ts
```diff
@@ -1,19 +1,89 @@
-export interface Figure {
+enum Color {
```
d-wasilewski, line 1: this can be just a type, not an enum, because that's how you use it

src/figuresClasses.ts
```diff
@@ -1,19 +1,89 @@
...
+  'red', 'green', 'blue'
+}
 
+export interface Figure {
+  shape: 'triangle' | 'circle' |'rectangle';
```
d-wasilewski, line 6: create an enum for shape and use it later like   `shape = Shape.rectangle;`

src/figuresClasses.ts
```diff
@@ -1,19 +1,89 @@
...
+    public color: Color,
+    public a: number,
+    public b: number,
+    public c: number,
+    public shape: 'triangle',
+  ) {
+    if ((this.a <= 0 || this.b <= 0 || this.c <= 0)
+        || this.c >= (this.a + this.b) || this.b >= (this.a + this.c)
+        || this.a >= (this.b + this.c)) {
+      throw new Error('ERROR');
```
d-wasilewski, line 22: checklist

src/figuresClasses.ts
```diff
@@ -1,19 +1,89 @@
...
+    }
 
-}
+    this.shape = 'triangle';
+    this.color = color;
+    this.a = a;
```
d-wasilewski, line 27: In TypeScript, when you declare a constructor parameter with an access modifier TypeScript implicitly creates a property on the class with the same name as the parameter and assigns the parameter value to that property.

src/figuresClasses.ts
```diff
@@ -1,19 +1,89 @@
...
-export class Triangle {
+export class Triangle implements Figure {
+  constructor(
+    public color: Color,
+    public a: number,
+    public b: number,
+    public c: number,
+    public shape: 'triangle',
+  ) {
+    if ((this.a <= 0 || this.b <= 0 || this.c <= 0)
```
d-wasilewski, line 19:
```suggestion
    if ((a <= 0 || b <= 0 || c <= 0)
```
Do these checks in the constructor this way, as you want check the coming variables anyway

src/figuresClasses.ts
```diff
@@ -1,19 +1,89 @@
...
+    this.b = b;
+    this.c = c;
+  }
 
-export class Circle {
+  getArea(): number {
+    const s = (this.a + this.b + this.c) / 2;
+    const a = +(
+      Math.sqrt(s * (s - this.a) * (s - this.b) * (s - this.c))
+    ).toFixed(2);
```
d-wasilewski, line 36: use Math.floor instead of toFixed

src/figuresClasses.ts
```diff
@@ -1,19 +1,89 @@
...
+
+    this.shape = 'circle';
+    this.color = color;
+    this.radius = radius;
+  }
 
+  getArea(): number {
+    const a = Math.floor(Math.PI * (this.radius ** 2) * 100) / 100;
+
+    return a;
```
d-wasilewski, line 60:
```suggestion
    return Math.floor(Math.PI * (this.radius ** 2) * 100) / 100;
```

d-wasilewski (changes requested):

src/figuresClasses.ts
sylcym: done

src/figuresClasses.ts
sylcym: done

src/figuresClasses.ts
sylcym: done

src/figuresClasses.ts
sylcym: done

src/figuresClasses.ts
sylcym: done

src/figuresClasses.ts
```diff
@@ -1,19 +1,85 @@
...
+    }
+
+    if (c >= (a + b) || b >= (a + c)
+        || a >= (b + c)) {
+      throw new Error(
+        'The longest side must be greater or equal than a sum of two others',
+      );
+    }
+
+    this.shape = 'triangle';
```
d-wasilewski, line 34: its unnecessary

src/figuresClasses.ts
```diff
@@ -1,19 +1,85 @@
...
+    if (width <= 0 || height <= 0) {
+      throw new Error('Width and height must be positive');
+    }
+    this.shape = 'rectangle';
+  }
+
+  getArea(): number {
+    const a = this.width * this.height;
+
+    return a;
```
d-wasilewski, line 79:
```suggestion
    return this.width * this.height;
```

d-wasilewski (changes requested):

src/figuresClasses.ts
```diff
@@ -1,19 +1,85 @@
...
-export class Circle {
+export class Triangle implements Figure {
+  constructor(
+    public color: Color,
+    public a: number,
+    public b: number,
+    public c: number,
+    public shape: Shape.triangle,
+  ) {
+    if ((a <= 0 || b <= 0 || c <= 0)) {
```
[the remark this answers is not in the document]
sylcym: done

src/figuresClasses.ts
sylcym: done

src/figuresClasses.ts
sylcym: done

d-wasilewski (approved):

PR #1169, opened 2023-04-23
solution

YBM3:


src/figuresClasses.ts
```diff
@@ -1,19 +1,81 @@
 export interface Figure {
-
+  shape : string;
```
d-wasilewski, line 2: Let's create the `Shape` type here with the appropriate values so that it could restrict the values that passed to the shape property

`shape: Shape`

d-wasilewski (changes requested):

src/figuresClasses.ts
```diff
@@ -1,19 +1,84 @@
+enum Shape {
+  Triangle = 'triangle',
+  Circle = 'circle',
+  Rectangle = 'rectangle',
+}
 export interface Figure {
-
+  shape: Shape;
+  color: string;
```
d-wasilewski, line 8: You can also create a `type` for the color with the predefined values

d-wasilewski (changes requested):
Great job!

PR #1170, opened 2023-04-23
solution

AnastassiaH:


src/figuresClasses.ts
```diff
@@ -1,19 +1,76 @@
...
-
+export interface Figure {
+  shape: string,
+  color: string,
+  getArea: () => number,
 }
 
-export class Circle {
+export class Triangle implements Figure {
+  shape: string = 'triangle';
```
VitaliyBondarenko1982, line 20: make type or enum for shape and use in all places instead string

VitaliyBondarenko1982 (changes requested):

src/figuresClasses.ts
```diff
@@ -1,19 +1,79 @@
-export interface Figure {
+function checkError(...sides: number[]): void {
+  const isError = [...sides].some((el) => el <= 0);
+  const sidesSum = [...sides].reduce((a, i) => a + i, 0);
+  const maxSide = Math.max(...sides);
+  const isTriangle = sides.length === 3
+    ? (maxSide >= (sidesSum - maxSide)) : false;
 
+  if (isError || isTriangle) {
+    throw new Error('error message');
```
vlad-demchuk, line 9: Please provide a more informative error message, like 'Invalid side length or triangle inequality violated 📐'

vlad-demchuk (approved):

PR #1173, opened 2023-04-23
Solution

KielpinskiJ:


YBM3 (approved):

src/figuresClasses.ts
```diff
@@ -1,19 +1,65 @@
 export interface Figure {
+  shape: 'triangle' | 'circle' | 'rectangle'
+  color: 'red' | 'green' | 'blue'
```
d-wasilewski, line 3: here you should use the types you created

src/figuresClasses.ts
```diff
@@ -1,19 +1,65 @@
...
 
+  getArea(): number;
 }
 
-export class Triangle {
+type Color = 'red' | 'green' | 'blue';
+type Shape = 'triangle' | 'circle' | 'rectangle';
```
d-wasilewski, line 9: instead of `type` you can create an `enum` here and use it later like `shape = Shape.rectangle`;

src/figuresClasses.ts
```diff
@@ -1,19 +1,65 @@
...
+  }
 }
 
-export function getInfo(figure) {
+export class Rectangle implements Figure {
+  shape: Shape = 'rectangle';
+
+  constructor(public color: Color, public a: number, public b: number) {
+    if (a <= 0 || b <= 0) {
+      throw new Error('All values must be higher than 0');
```
d-wasilewski, line 54:
```suggestion
      throw new Error('All values must be greater than 0');
```

d-wasilewski (changes requested):

PR #1176, opened 2023-04-24
First solution

marshalek37:


Amsko6 (approved):

src/figuresClasses.ts
```diff
@@ -1,19 +1,75 @@
 export interface Figure {
-
+  shape: string;
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,75 @@
...
+  color: string;
```
d-wasilewski, line 3: create a type for available color values

src/figuresClasses.ts
```diff
@@ -1,19 +1,75 @@
...
+    public color: string,
+    public radius: number,
+  ) {
+    if (radius <= 0) {
+      throw new Error('Radius must be greater than zero!');
+    }
+  }
 
+  getArea(): number {
+    return Math.floor(Math.PI * this.radius * this.radius * 100) / 100;
```
d-wasilewski, line 51:
```suggestion
    return Math.floor(Math.PI * this.radius ** 2 * 100) / 100;
```

d-wasilewski (approved):

PR #1179, opened 2023-04-24
solution

natalia-klonowska:


src/figuresClasses.ts
```diff
@@ -1,19 +1,65 @@
 export interface Figure {
-
+  shape: string;
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,65 @@
...
+  color: string;
```
d-wasilewski, line 3: create a type for available color values

src/figuresClasses.ts
```diff
@@ -1,19 +1,65 @@
...
+    public b: number,
+    public c: number,
+  ) {
+    if (a <= 0 || b <= 0 || c <= 0) {
+      throw new Error('some variable is less or equal to zero');
+    }
+
+    const longestSide = Math.max(a, b, c);
+
+    if (longestSide >= (a + b + c - longestSide)) {
```
d-wasilewski, line 22: that's a smart solution! 👏

d-wasilewski (changes requested):

d-wasilewski (approved):

PR #1180, opened 2023-04-25
Develop

SuaBy514:


src/figuresClasses.ts
```diff
@@ -1,19 +1,82 @@
...
 }
 
-export class Rectangle {
+export class Circle implements Figure {
+  public shape;
+
+  constructor(public color: 'red' | 'green' | 'blue', public radius: number) {
+    this.shape = 'circle';
+    this.color = color;
+    this.radius = radius;
```
d-wasilewski, line 46: checklist 5

src/figuresClasses.ts
```diff
@@ -1,19 +1,82 @@
 export interface Figure {
-
+  shape: string,
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,82 @@
...
+  color: string,
```
d-wasilewski, line 3: create a type for the color with predefined values

src/figuresClasses.ts
```diff
@@ -1,19 +1,82 @@
...
 }
 
-export class Triangle {
+export class Triangle implements Figure {
+  public shape;
 
-}
+  public sides;
+
+  constructor(public color: 'red' | 'green' | 'blue', ...args: number[]) {
```
d-wasilewski, line 12: use expected values instead of args, you know you want to only get 3 sides of the triangle, no more or no less, you shouldn't use ...args in this case

src/figuresClasses.ts
```diff
@@ -1,19 +1,82 @@
...
+    }
+  }
+
+  getArea(): number {
+    return this.sides[0] * this.sides[1];
+  }
+}
 
+export function getInfo(
+  figure: {['color']: string, ['shape']: string, ['getArea']: Function},
```
d-wasilewski, line 79:
```suggestion
  figure: Figure,
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,82 @@
...
+    this.color = color;
+    this.sides = [...args].sort();
+
+    if (this.sides[0] <= 0) {
+      throw new Error('Rectangle not possible with given sides!');
+    }
+  }
+
+  getArea(): number {
+    return this.sides[0] * this.sides[1];
```
d-wasilewski, line 74: here aswell, get width and height in the constructor and use it as variables, not an array

d-wasilewski (changes requested):

src/figuresClasses.ts
SuaBy514: Wait... you can do that? I spent like a good 20 minutes getting that part to work... it was this simple all along...?

src/figuresClasses.ts
SuaBy514: I just find arrays to be easier to work with, but I understand

src/figuresClasses.ts
SuaBy514: But... the checklist only has 3 points in it

SuaBy514: 1. [CODE STYLE] - ```getArea``` function doesn't need to accept any parameters. All necessary values are already available as object fields
2. [CODE STYLE] - Enum constant names usually start with capital letter. Values can have any case that's necessary.

BAD EXAMPLE:
```
export enum Directions {
  up = 'UP',
  down = 'DOWN',
  left = 'LEFT',
  right = 'RIGHT',
}
```

GOOD EXAMPLE:
```
export enum Directions {
  Up = 'UP',
  Down = 'DOWN',
  Left = 'LEFT',
  Right = 'RIGHT',
}
```

3. [CODE STYLE] - If you throw an error try to provide its detailed description in error message.

BAD EXAMPLE:
```
if (x % 2 !== 0) {
  throw new Error('Something went wrong')
}
```

GOOD EXAMPLE:
```
if (x % 2 !== 0) {
  throw new Error('Resulting value should be even number')
}
```



That's.. the whole file

d-wasilewski: you are right, I added this to the wrong file...
it's point 4 right now

src/figuresClasses.ts
d-wasilewski: assign this enum as a shape type
shape: Shape

The same with the color

src/figuresClasses.ts
```diff
@@ -1,19 +1,109 @@
...
+
+  public c: number;
 
-export class Circle {
+  constructor(color: Color, a: number, b: number, c: number) {
+    this.shape = Shape.Triangle;
+    this.color = color;
+    this.a = a;
+    this.b = b;
+    this.c = c;
```
d-wasilewski, line 31: check checklist again (I updated it) and remove it

src/figuresClasses.ts
```diff
@@ -1,19 +1,109 @@
...
 
+    const sides: number[] = [a, b, c].sort(); /* Sort helps with detecting 0s
+                                      if they were not in the first position */
+
+    if ((sides[0] + sides[1] <= sides[2])
+    || (sides[2] + sides[0] <= sides[1])
+    || (sides[2] + sides[1] <= sides[0])
```
d-wasilewski, line 38: change arrays to regular variables in all places
Even if it's easier for you to manage it's its harder to read for somebody else, you know what they contain but it might not be so obvious for someone checking it, it's better to use named variables

d-wasilewski (changes requested):

src/figuresClasses.ts
SuaBy514: image
Linter disagrees

sylcym (approved):

src/figuresClasses.ts
```diff
@@ -1,19 +1,106 @@
 export interface Figure {
+  shape: string,
```
d-wasilewski, line 2:
```suggestion
  shape: Shape,
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,106 @@
...
+  color: string,
```
d-wasilewski, line 3:
```suggestion
  color: Color,
```

src/figuresClasses.ts
d-wasilewski: This is because you should also delete it's declaration above the constructor

src/figuresClasses.ts
```diff
@@ -1,19 +1,106 @@
...
+    this.b = b;
+    this.c = c;
+  }
+
+  getArea(): number {
+    /* I hate math and therefore hate you for this,
+    but as requested, Heron's formula */
+    const semiPer: number = (this.a + this.b + this.c) / 2;
+    const bracket1: number = semiPer - this.a;
+    const bracket2: number = semiPer - this.b;
```
d-wasilewski, line 48: can you think of better naming?

src/figuresClasses.ts
```diff
@@ -1,19 +1,106 @@
...
+  }
 }
 
-export class Rectangle {
+export class Circle implements Figure {
+  public color: Color;
+
+  public shape: Shape;
+
+  public radius: number;
```
d-wasilewski, line 62:
```suggestion
 shape = Shape.Rectangle
```

d-wasilewski (changes requested):
remove the unnecessary things, if it won't work come to QnA

src/figuresClasses.ts
SuaBy514: image

SuaBy514: The linter does not let me do it in any way that does not have:

- A public declaration above the constructor
- A parameter reference
- A this.color

HOWEVER
It seems to be working fine with a, b, and c

SuaBy514: HOWEVER HOWEVER
Doing that does not allow the getArea function to read a, b, or c

SuaBy514: image
as seen here

kasiakubisiak (approved):

PR #1182, opened 2023-04-25
ts_figures_classes

JulyaPetrovskaya:


src/figuresClasses.ts
```diff
@@ -1,19 +1,100 @@
...
-export class Circle {
+    if (a + b <= c || b + c <= a || c + a <= b) {
+      throw new Error(`Triangle with sides ${a}, ${b}, and ${c} can't exist`);
+    }
+  }
 
+  getArea(): number {
+    const [a, b, c] = [...this.sides];
+    const perimeter = (a + b + c);
+
```
etojeDenys, line 36: Instead of calculating `perimeter / 2` multiple times, it's better to store it in a variable.

src/figuresClasses.ts
```diff
@@ -1,19 +1,100 @@
...
 }
 
-export class Triangle {
+export class Triangle implements Figure {
+  shape: 'triangle';
 
-}
+  sides: number[];
+
+  color: 'red' | 'green' | 'blue';
```
etojeDenys, line 12: seems redundant

etojeDenys (changes requested):

src/figuresClasses.ts
```diff
@@ -1,19 +1,98 @@
 export interface Figure {
-
+  shape: 'triangle' | 'circle' | 'rectangle';
+  color: 'red' | 'green' | 'blue';
```
anstsot, line 3: it will be great to create `enum` typings for shape and color and use it everywhere

anstsot (approved):
Great job, approved!

PR #1183, opened 2023-04-25
add task solution

MaksymKos:


src/figuresClasses.ts
```diff
@@ -1,19 +1,74 @@
...
 
-}
+  constructor(
+    public color: Color,
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    if (a <= 0 || b <= 0 || c <= 0 || c >= a + b) {
+      throw new Error(`Sides ${a}, ${b} and ${c} can't form a triangle`);
```
etojeDenys, line 20: The triangle inequality check should also include `a >= b + c` and `b >= a + c`. 📐

etojeDenys (changes requested):

src/figuresClasses.ts
OleziO: Not resolved. Just replace condition with `a >= b + c` and `b >= a + c`

OleziO (changes requested):

DarkMistyRoom (approved):

PR #1186, opened 2023-04-28
figures classes task was done

2pasha:


src/figuresClasses.ts
```diff
@@ -1,19 +1,95 @@
...
+  color: string;
+  getArea(): number;
 }
 
-export class Triangle {
+export class Triangle implements Figure {
+  shape: string;
 
-}
+  color: string;
```
IvanFesenko, line 10: Redudant. Its already was declared in Figure

src/figuresClasses.ts
```diff
@@ -1,19 +1,95 @@
 export interface Figure {
-
+  shape: string;
+  color: string;
+  getArea(): number;
 }
 
-export class Triangle {
+export class Triangle implements Figure {
+  shape: string;
```
IvanFesenko, line 8: because this field not be changed
```suggestion
  shape = 'triangle'
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,95 @@
...
+
+    if (this.a >= this.b + this.c
+      || this.b >= this.a + this.c
+      || this.c >= this.b + this.a
+    ) {
+      throw new Error('Invalid triangle sides');
+    }
 
-export class Circle {
+    this.shape = 'triangle';
```
IvanFesenko, line 29:
```suggestion
```

IvanFesenko (changes requested):
Check my comment. Fix all similar cases

src/figuresClasses.ts
```diff
@@ -1,19 +1,86 @@
...
+
+  constructor(
+    public color: string,
+    private radius: number,
+  ) {
+    if (radius <= 0) {
+      throw new Error('Circle radius must be greater than 0');
+    }
+
+    this.color = color;
```
etojeDenys, line 49: It's not necessary to assign 'color' and 'radius' to 'this.color' and 'this.radius' respectively, as they are already assigned using 'public' and 'private' keywords in the constructor.

src/figuresClasses.ts
```diff
@@ -1,19 +1,86 @@
...
+  constructor(
+    public color: string,
+    private width: number,
+    private height: number,
+  ) {
+    if (this.width <= 0 || this.height <= 0) {
+      throw new Error('Rectangle width and height must be greater than 0');
+    }
+
+    this.color = color;
```
etojeDenys, line 72: It's not necessary to assign 'color', 'width', and 'height' to 'this.color', 'this.width', and 'this.height' respectively, as they are already assigned using 'public' and 'private' keywords in the constructor.

etojeDenys (changes requested):

nazarmatsevych (approved):

PR #1189, opened 2023-04-30
Solution

Ssviatt:


src/figuresClasses.ts
```diff
@@ -1,19 +1,85 @@
...
+export class Triangle implements Figure {
+  shape: string = 'triangle';
+
+  constructor(
+    public color: string,
+    public sideA: number,
+    public sideB: number,
+    public sideC: number,
+  ) {
+    if (sideA <= 0 || sideB <= 0 || sideC <= 0) {
```
tetiana-ivanchenko, line 56: Instead of using 'sideA', 'sideB', and 'sideC', use 'this.sideA', 'this.sideB', and 'this.sideC' to refer to the class properties.

src/figuresClasses.ts
```diff
@@ -1,19 +1,85 @@
...
+
+  constructor(
+    public color: string,
+    public radius: number,
+  ) {
+    if (radius <= 0) {
+      throw new Error('Invalid radius');
+    }
+    this.color = color;
+    this.radius = radius;
```
tetiana-ivanchenko, line 39: You don't need to set 'color' and 'radius' again, because they are already assigned in the constructor using 'public'. Remove these lines.

src/figuresClasses.ts
```diff
@@ -1,19 +1,85 @@
...
+
+    if (sideA + sideB <= sideC
+        || sideA + sideC <= sideB
+        || sideB + sideC <= sideA) {
+      throw new Error('Invalid sides for a triangle');
+    }
+    this.color = color;
+    this.sideA = sideA;
+    this.sideB = sideB;
+    this.sideC = sideC;
```
tetiana-ivanchenko, line 68: You don't need to set 'color', 'sideA', 'sideB', and 'sideC' again, because they are already assigned in the constructor using 'public'. Remove these lines.

src/figuresClasses.ts
```diff
@@ -1,19 +1,85 @@
...
+    public color: string,
+    public width: number,
+    public height: number,
+  ) {
+    if (width <= 0 || height <= 0) {
+      throw new Error('Invalid width or height');
+    }
+    this.color = color;
+    this.width = width;
+    this.height = height;
```
tetiana-ivanchenko, line 20: You don't need to set 'color', 'width', and 'height' again, because they are already assigned in the constructor using 'public'. Remove these lines.

tetiana-ivanchenko (changes requested):

src/figuresClasses.ts
```diff
@@ -1,19 +1,76 @@
 export interface Figure {
-
+  shape: string;
+  color: string;
+  getArea(): number;
 }
 
-export class Triangle {
+export class Rectangle implements Figure {
+  shape: string = 'rectangle';
```
ivbutsykin, line 8: TS will know that `shape` is `string` by first assign:
```suggestion
  shape = 'rectangle';
```

ivbutsykin (approved):
Good job!

PR #1190, opened 2023-05-02
Solution

NadieinOleh:


src/figuresClasses.ts
```diff
@@ -1,19 +1,78 @@
+type Shape = 'triangle' | 'circle' | 'rectangle';
+type Color = 'red' | 'green' | 'blue';
+
 export interface Figure {
+  shape: Color;
+  color : Shape;
```
polosanya, line 6:
```suggestion
  shape: Shape;
  color: Color;
```
It seems you have mixed up the types)

src/figuresClasses.ts
```diff
@@ -1,19 +1,78 @@
...
-export class Triangle {
+export class Triangle implements Figure {
+  shape: Shape = 'triangle';
 
-}
+  constructor(
+    public color : Color,
+    public a:number,
+    public b:number,
+    public c:number,
```
polosanya, line 18:
```suggestion
    public color: Color,
    public a: number,
    public b: number,
    public c: number,
```
Fix formatting everywhere

src/figuresClasses.ts
```diff
@@ -1,19 +1,78 @@
...
+    }
+  }
 
+  getArea(): number {
+    const semiperimeter = (this.a + this.b + this.c) / 2;
+    const underSqrt = semiperimeter
+      * (semiperimeter - this.a)
+      * (semiperimeter - this.b) * (semiperimeter - this.c);
+    const AreaLong = Math.sqrt(underSqrt);
+    const Area = +((AreaLong).toFixed(2));
```
polosanya, line 34:
```suggestion
    const areaLong = Math.sqrt(underSqrt);
    const area = +((AreaLong).toFixed(2));
```
There is no need to capitalize the first letter

polosanya (changes requested):

PR #1198, opened 2023-05-11
classes for figures were added

Olena-Shevchenko:


src/figuresClasses.ts
```diff
@@ -1,19 +1,79 @@
...
-}
+  constructor(
+    public color: Color,
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    if (a <= 0 || b <= 0 || c <= 0) {
+      throw new Error('Wrong sides');
+    } else if ((a + b <= c) || (a + c <= b) || (b + c <= a)) {
```
polosanya, line 21: Use more descriptive text for errors.
Also, there is no need to use else...if as code will not go further after throwing an error anyway
image

polosanya (changes requested):
Good job! Let's improve your solution a bit 🙂

witflash (approved):
Well done 👍🏻

PR #1201, opened 2023-05-12
Add solution

ruslanpashkov:


src/figuresClasses.ts
```diff
@@ -1,19 +1,104 @@
...
+    if (!this.isValidRectangle()) {
+      throw new Error(
+        `Width ${this.width} and height ${this.height} cannot form a rectangle`,
+      );
+    }
+  }
+
+  isValidRectangle(): boolean {
+    const { width, height } = this;
+
```
dinaminahi, line 89: Add a check for the width and height to be <= 0, and if any of them is <= 0, return false in isValidRectangle method.

src/figuresClasses.ts
```diff
@@ -1,19 +1,104 @@
...
+    const area = width * height;
+
+    return Math.floor(area * 100) / 100;
+  }
+}
 
+export function getInfo(figure: Triangle | Circle | Rectangle): string {
+  return `A ${figure.color} ${figure.shape} - ${figure.getArea()}`;
```
dinaminahi, line 103: Use the Figure interface as the type of the parameter in the getInfo function instead of using an explicit union of Triangle, Circle, and Rectangle.

dinaminahi (approved):
🥇

src/figuresClasses.ts
DidenkoDenis306, line 103:
```suggestion
  return `The ${figure.color} ${figure.shape} - ${figure.getArea()}`;
```

DidenkoDenis306 (changes requested):

PR #1206, opened 2023-05-16
add figuresClasses solution

DidenkoDenis306:
One test fails because it asks for a circle area of 113.09 and gets 113.1. toFixed(2) does not work.

### DIFF
--- a/src/figuresClasses.ts
+++ b/src/figuresClasses.ts
@@ -1,19 +1,100 @@
-export interface Figure {
+type Shape = 'triangle' | 'circle' | 'rectangle';
+type Color = 'red' | 'green' | 'blue';
 
+export interface Figure {
+  shape: Shape;
+  color: Color;
+  getArea(): number;
 }
 
-export class Triangle {
+export class Triangle implements Figure {
+  shape: Shape = 'triangle';
 
+  color: Color;
+
+  sides: number[];
+
+  constructor(
+    color: Color,
+    a: number,
+    b: number,
+    c: number,
+  ) {
+    if (a <= 0
+      || b <= 0
+      || c <= 0) {
+      throw new Error('Length of each side should be greater than 0');
+    }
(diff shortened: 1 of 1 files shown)


src/figuresClasses.ts
```diff
@@ -1,19 +1,100 @@
...
+
+    this.color = color;
+    this.sides = [a, b, c];
+  }
+
+  getArea(): number {
+    const [a, b, c] = this.sides;
+    const s = (a + b + c) / 2;
+
+    return Number((Math.sqrt(s * (s - a) * (s - b) * (s - c))).toFixed(2));
```
33Orange, line 43: Try to use `Math.floor`

src/figuresClasses.ts
```diff
@@ -1,19 +1,100 @@
...
+    if (radius <= 0) {
+      throw new Error('Radius should be greater than 0');
+    }
+
+    this.color = color;
+    this.radius = radius;
+  }
+
+  getArea(): number {
+    return Number((Math.PI * (this.radius ** 2)).toFixed(2));
```
33Orange, line 64: Try to use `Math.floor`

src/figuresClasses.ts
```diff
@@ -1,19 +1,100 @@
...
+      throw new Error('Width and height should be greater than 0');
+    }
+
+    this.color = color;
+    this.width = width;
+    this.height = height;
+  }
 
+  getArea(): number {
+    return Number((this.width * this.height).toFixed(2));
```
33Orange, line 92: Try to use `Math.floor`

33Orange (changes requested):
Well done !

A few moments to fix

33Orange (changes requested):
Well done !

A few moments to fix

dinaminahi (approved):

PR #1211, opened 2023-05-18
Solution

G37ua:


src/figuresClasses.ts
```diff
@@ -1,19 +1,95 @@
...
+export class Triangle implements Figure {
+  shape = 'triangle';
+
+  color: string;
+
+  a: number;
+
+  b: number;
+
+  c: number;
```
polosanya, line 16:
```suggestion
  shape = 'triangle';
  color: string;
  a: number;
  b: number;
  c: number;
```
There is actually no need to add empty lines between each property

src/figuresClasses.ts
```diff
@@ -1,19 +1,95 @@
 export interface Figure {
-
+  shape: string;
+  color: string;
```
polosanya, line 3: It would be great to specify what exactly shape or color it could have
image

src/figuresClasses.ts
```diff
@@ -1,19 +1,95 @@
...
+    }
+
+    if (a + b <= c || a + c <= b || b + c <= a) {
+      throw new Error(`Sides ${a}, ${b}, and ${c} can't form a triangle`);
+    }
+
+    this.color = color;
+    this.a = a;
+    this.b = b;
+    this.c = c;
```
polosanya, line 35: image

polosanya (changes requested):
Almost done!

src/figuresClasses.ts
G37ua: Linter swears, wants to add empty lines between each property.

G37ua:
Corrected. Linter has nothing more to swear at, so there are no more properties between which he wanted empty lines :)

polosanya (approved):
Well done!

PR #1213, opened 2023-05-22
add task solution

maksym-nemera:


src/figuresClasses.ts
```diff
@@ -1,19 +1,79 @@
...
+    public height: number,
+  ) {
+    if (this.width < 0 || this.height < 0) {
+      throw new Error('Width and height of the rectangle must be non-negative');
+    }
+  }
+
+  getArea(): number {
+    return this.width * this.height;
+  }
```
33Orange, line 74: The area calculation should be rounded down to hundredths using Math.floor(area * 100) / 100.
```suggestion
  getArea(): number {
    const area = this.width * this.height;
    return Math.round(area * 100) / 100;
  }
```

33Orange (changes requested):
Almost done.

33Orange (approved):
Well done !

PR #1214, opened 2023-05-23
done first try

VladyslavKucher:


src/figuresClasses.ts
```diff
@@ -1,19 +1,105 @@
...
+    * (semiperimeter - this.B)
+      * (semiperimeter - this.C));
+
+    return Math.round(area * 100) / 100;
+  }
 }
 
-export class Rectangle {
+export class Circle implements Figure {
+  shape: 'triangle' | 'circle' | 'rectangle';
```
polosanya, line 56: It should be circle

src/figuresClasses.ts
```diff
@@ -1,19 +1,105 @@
...
+  getArea(): number {
+    const area = Math.PI * this.radius * this.radius;
+
+    return Math.floor(area * 100) / 100;
+  }
 }
 
-export function getInfo(figure) {
+export class Rectangle implements Figure {
+  shape: 'triangle' | 'circle' | 'rectangle';
```
polosanya, line 78: Same problem here

polosanya (changes requested):

anstsot (approved):
Good job, approved!

PR #1215, opened 2023-05-23
solution

yuron-maker:


src/figuresClasses.ts
```diff
@@ -1,19 +1,77 @@
...
+
+  constructor(
+    public color: Color,
+    public width : number,
+    public height : number,
+  ) {
+    this.shape = 'rectangle';
+
+    if (width <= 0 || height <= 0) {
+      throw new Error('Not a rectangle');
```
sTorba24, line 64: use more meaningful text to explain an error

PR #1217, opened 2023-05-24
Develop

mariuszgit:


src/figuresClasses.ts
```diff
@@ -1,19 +1,63 @@
+/* eslint-disable max-len */
 export interface Figure {
-
+  shape: 'triangle' | 'circle' | 'rectangle';
+  color: 'red' | 'green' | 'blue';
+  getArea(): number;
 }
 
 export class Triangle {
```
darokrk, line 8: We should reuse our interface Figure to type other classes.

Please use `class Triangle implements Figure`

Also for other classes

src/figuresClasses.ts
```diff
@@ -1,19 +1,63 @@
...
+  public shape = 'triangle';
+
+  constructor(public color: string, public a: number, public b: number, public c: number) {
```
darokrk, line 11: You should change this constructor to base on types from Figure interface.

Please try `constructor(
    public a: number,
    public b: number,
    public c: number,
    public shape: 'triangle' | 'circle' | 'rectangle',
  )`
  
  Then we are strictly sure that our values be always the same as we declared in types :)
  Try to fix other classes.

darokrk (changes requested):
Looks good, but needs some small improvements. Check my suggestion and we are ready to go 👍🏽

Radoslaw-Czerniawski (approved):
Looks good :)

PR #1222, opened 2023-05-27
added figures classes: triangle, circle, rectangle

AlexKherson:


AlexKherson:
add task solution

src/figuresClasses.ts
```diff
@@ -1,19 +1,95 @@
 export interface Figure {
+  shape:string,
+  color:string,
+  getArea: () =>{}
```
naziamoff, line 4: the type for getArea looks a little weird 💀 
Why doesn't it return a number?

src/figuresClasses.ts
```diff
@@ -1,19 +1,95 @@
...
+}
 
+function roundedDownToHundredths(area:number):number {
```
naziamoff, line 7:
```suggestion
function roundDownToHundredths(area:number):number {
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,95 @@
...
+  }
+
+  getArea = ():number => {
+    const p = (this.a + this.b + this.c) / 2;
+    const area = Math.sqrt(p * (p - this.a) * (p - this.b) * (p - this.c));
+
+    return roundedDownToHundredths(area);
+  };
+
+  private isValidSides():boolean {
```
naziamoff, line 38:
```suggestion
  private areSidesValid():boolean {
```

naziamoff, line 38: great job creating a private method for that 🔥

src/figuresClasses.ts
```diff
@@ -1,19 +1,95 @@
...
+    public b:number,
+    public c:number,
+  ) {
+    if (!this.isValidSides()) {
+      throw new Error('incorrect sides of triangle');
+    }
+
+    if (!this.isTriangle()) {
+      throw new Error('this is not triangle');
+    }
```
naziamoff, line 26: you can just call `this.isTriangle()` without the if statement - and move the error message to the private method

src/figuresClasses.ts
```diff
@@ -1,19 +1,95 @@
...
+    return this.a > 0 && this.b > 0 && this.c > 0;
+  }
+
+  private isTriangle():boolean {
+    const sides = [this.a, this.b, this.c];
 
-export class Circle {
+    sides.sort((a, b) => a - b);
 
+    return sides[0] + sides[1] > sides[2];
```
naziamoff, line 47: I meant something like that:
```suggestion
    if (sides[0] + sides[1] > sides[2]) {
      throw Error...
    }
```

naziamoff (changes requested):
great job actually 🔥 
Still a couple of things to polish

src/figuresClasses.ts
```diff
@@ -1,19 +1,94 @@
...
+  public shape = 'triangle';
 
-}
+  constructor(
+    public color:string,
+    public a:number,
+    public b:number,
+    public c:number,
+  ) {
+    if (!this.AreValidSides()) {
```
naziamoff, line 20: should not be capitalized
```suggestion
    if (!this.areValidSides()) {
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,94 @@
...
+      throw new Error('this is not triangle');
+    }
+  }
 }
 
-export class Rectangle {
+export class Circle implements Figure {
+  public shape = 'circle';
+
+  constructor(public color:string, public radius:number) {
```
naziamoff, line 53:
```suggestion
  constructor(
    public color:string, 
    public radius:number
  ) {
```

naziamoff (approved):
👍

PR #1229, opened 2023-05-31
task solution

hekaq:


src/figuresClasses.ts
```diff
@@ -1,19 +1,82 @@
...
+    public b: number,
+    public c: number,
+  ) {
+    if (a <= 0 || b <= 0 || c <= 0) {
+      throw new Error('Length of each side must be greater than 0');
+    }
+
+    if (a + b <= c || a + c <= b || b + c <= a) {
+      throw new Error('The given sides cannot form a triangle');
+    }
```
naziamoff, line 27: I would move this logic to a separate method, like `validateTriangle` or sth

naziamoff (approved):
clear and concise. Well done 👍

PR #1230, opened 2023-05-31
solution

skinnyjazz:


src/figuresClasses.ts
```diff
@@ -1,19 +1,82 @@
...
+  public shape: Shape = 'triangle';
 
-}
+  constructor(
+    public color: Color,
+    public sideA: number,
+    public sideB: number,
+    public sideC: number,
+  ) {
+    if (sideA === 0 || sideB === 0 || sideC === 0) {
```
naziamoff, line 20: what if I pass `-1`?

src/figuresClasses.ts
```diff
@@ -1,19 +1,82 @@
...
+      throw Error('Lengths of the sides must be greater than 0.');
+    }
+
+    if (
+      Math.max(sideA, sideB, sideC)
+      >= sideA + sideB + sideC - Math.max(sideA, sideB, sideC)
+    ) {
```
naziamoff, line 27: too complex logic to have it contained inside the if condition

src/figuresClasses.ts
```diff
@@ -1,19 +1,82 @@
...
+    public color: Color,
+    public sideA: number,
+    public sideB: number,
+    public sideC: number,
+  ) {
+    if (sideA === 0 || sideB === 0 || sideC === 0) {
+      throw Error('Lengths of the sides must be greater than 0.');
+    }
+
+    if (
```
naziamoff, line 24: also, let's integrate both and create a separate method like `validateTriangle`  for that purpose

src/figuresClasses.ts
```diff
@@ -1,19 +1,82 @@
...
 
+    return Math.floor(area * 100) / 100;
+  }
 }
 
-export class Rectangle {
+export class Circle implements Figure {
+  public shape: Shape = 'circle';
+
+  constructor(public color: Color, public radius: number) {
```
naziamoff, line 47:
```suggestion
  constructor(
    public color: Color, 
    public radius: number
  ) {
```

naziamoff (changes requested):
not bad, but there are a couple of important suggestions

naziamoff (approved):

PR #1232, opened 2023-05-31
Solution

amashutikov:


src/figuresClasses.ts
```diff
@@ -1,19 +1,79 @@
...
+    if (a <= 0 || b <= 0 || c <= 0) {
+      throw new Error('Size cant be <= 0');
+    } else if (
+      a >= b + c
+      || b >= a + c
+      || c >= a + b
+    ) {
+      throw new Error('Wrong size of triangle sides');
+    }
+  }
```
naziamoff, line 28: looks really huge. Let's create a method `validateTriangle` which will handle such logic

naziamoff (approved):
good job 🔥

PR #1234, opened 2023-05-31
add solution

growingananas:


src/figuresClasses.ts
```diff
@@ -1,19 +1,70 @@
...
-}
+  constructor(
+    public color: 'red' | 'green' | 'blue',
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    if (this.a <= 0 || this.b <= 0 || this.c <= 0
+      || this.a + this.b <= this.c || this.a + this.c <= this.b
+      || this.b + this.c <= this.a) {
```
naziamoff, line 18: looks messy. Let's move all this logic to a separate method which would validate the input

src/figuresClasses.ts
```diff
@@ -1,19 +1,70 @@
...
+      throw new Error('This figure is not a triangle');
+    }
+  }
 
-export class Circle {
+  getArea(): number {
+    const perimeter = (this.a + this.b + this.c) / 2;
 
+    const squareOfTriangle = Math.sqrt(perimeter * ((perimeter - this.a)
+    * (perimeter - this.b) * (perimeter - this.c)));
```
naziamoff, line 27: I would come up with something like that
```suggestion
    const squareOfTriangle = Math.sqrt(perimeter * (
        (perimeter - this.a) * (perimeter - this.b) * (perimeter - this.c)
      ));
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,70 @@
...
+  constructor(
+    public color: 'red' | 'green' | 'blue',
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    if (this.a <= 0 || this.b <= 0 || this.c <= 0
+      || this.a + this.b <= this.c || this.a + this.c <= this.b
+      || this.b + this.c <= this.a) {
+      throw new Error('This figure is not a triangle');
```
naziamoff, line 19: better to have it more descriptive, as it would be difficult to understand what exactly is wrong with my input in that case

naziamoff (changes requested):
not bad, but let's polish a little

src/figuresClasses.ts
```diff
@@ -1,19 +1,80 @@
...
+    if (
+      this.a <= 0
+      || this.b <= 0
+      || this.c <= 0
+      || this.a + this.b <= this.c
+      || this.a + this.c <= this.b
+      || this.b + this.c <= this.a
+    ) {
+      throw new Error('Error: the sides do not form a triangle');
+    }
```
naziamoff, line 29: let's split into two errors - it should be clear what is wrong with my input

naziamoff (approved):
great job 🔥 
See one suggestion

PR #1235, opened 2023-05-31
Solution

julshvets:


src/figuresClasses.ts
```diff
@@ -1,19 +1,91 @@
...
 
+    if (a >= b + c || b >= a + c || c >= a + b) {
+      throw new Error('Sides cannot form a triangle');
+    }
+  }
+
+  getArea():number {
+    const p:number = (this.a + this.b + this.c) / 2;
+    const triangleArea = Math.sqrt(p
+      * (p - this.a) * (p - this.b) * (p - this.c));
```
naziamoff, line 40: quite difficult to read. Consider creating some extra variables or moving some logic to a separate helper

src/figuresClasses.ts
```diff
@@ -1,19 +1,91 @@
...
+    public b: number,
+    public c: number,
+  ) {
+    if (a <= 0 || b <= 0 || c <= 0) {
+      throw new Error('Invalid side value of triangle');
+    }
 
+    if (a >= b + c || b >= a + c || c >= a + b) {
+      throw new Error('Sides cannot form a triangle');
+    }
```
naziamoff, line 34: would be nice to create a method like `validateTriangle` or sth like that where you handle such cases

naziamoff (approved):
good job 🔥 

See minor suggestions

src/figuresClasses.ts
julshvets: Done.

src/figuresClasses.ts
julshvets: Done.
Thank you!

PR #1236, opened 2023-05-31
Add task solution

IrynaBalandiukh:


src/figuresClasses.ts
```diff
@@ -1,19 +1,74 @@
...
+  shape: Shape = 'circle';
+
+  constructor(public color: Color, public radius: number) {
+    if (radius <= 0) {
+      throw new Error('Radius can not be less than 1');
+    }
+  }
 
+  getArea(): number {
+    return Math.floor((Math.PI * this.radius * this.radius) * 100) / 100;
```
33Orange, line 50:
```suggestion
    return Math.floor((Math.PI * (this.radius ** 2) * 100) / 100;
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,74 @@
...
+    public height: number,
+  ) {
+    if (width <= 0 || height <= 0) {
+      throw new Error('Rectangle side can not be less than 1');
+    }
+  }
 
-export function getInfo(figure) {
+  getArea(): number {
+    return this.width * this.height;
```
33Orange, line 68:
```suggestion
    return Math.floor(this.width * this.height * 100) / 100;
```

33Orange (changes requested):
Good job.

A few moments

naziamoff (approved):
🔥

PR #1239, opened 2023-05-31
add functionality to figureClasses

cephy77:


src/figuresClasses.ts
```diff
@@ -1,19 +1,75 @@
...
+  public shape: Figure['shape'] = 'triangle';
 
-}
+  constructor(
+    public color: Figure['color'],
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    if (this.isInvalid()) {
```
naziamoff, line 16: usually it is done this way. It is much better to start booleans with positive - as it is much easier to process in human mind
```suggestion
    if (!this.isValid()) {
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,75 @@
...
+    const sides = [this.a, this.b, this.c].sort((a, b) => b - a);
+    const isSideTooSmall = sides.some((side) => side <= 0);
+    const isSideTooLarge = sides[0] >= sides[1] + sides[2];
 
-export class Circle {
+    return isSideTooSmall || isSideTooLarge;
+  }
 
+  getArea(): number {
+    const s = [this.a, this.b, this.c].reduce((acc, curr) => acc + curr, 0) / 2;
```
naziamoff, line 30: what's s?

naziamoff (approved):
well done 🔥 
see a couple of suggestions

src/figuresClasses.ts
cephy77: @naziamoff Always break a leg in such things

src/figuresClasses.ts
cephy77: @naziamoff  semiperimeter, and yes I wrote this as 's' because of the line length issues

PR #1240, opened 2023-05-31
add solution

allagutsul27:


src/figuresClasses.ts
```diff
@@ -1,19 +1,80 @@
...
 
-export class Circle {
+export class Circle implements Figure {
+  shape: Shape = 'circle';
+
+  constructor(
+    public color: Color,
+    public radius: number,
+  ) {
+    this.radius = radius;
```
etojeDenys, line 46: This line is unnecessary as the radius is already assigned in the constructor's parameter

src/figuresClasses.ts
```diff
@@ -1,19 +1,80 @@
...
+
+  constructor(
+    public color: Color,
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    const longestSide = Math.max(a, b, c);
+
+    if (longestSide >= this.b + this.c || longestSide >= this.a + this.c
```
etojeDenys, line 21: You should change the condition in the if statement to check if the sum of the two smallest sides is less than or equal to the longest side

src/figuresClasses.ts
```diff
@@ -1,19 +1,80 @@
...
+  }
 
+  getArea(): number {
+    return Math.floor(this.width * this.height * 100) / 100;
+  }
 }
 
-export function getInfo(figure) {
+export function getInfo(figure: Figure): string {
+  const figureArea = (figure.getArea() * 100) / 100;
```
etojeDenys, line 77: You don't need to recalculate the area here, you can directly use figure.getArea()

etojeDenys (approved):
lgtm

PR #1245, opened 2023-05-31
solution

snuber-w:


src/figuresClasses.ts
```diff
@@ -1,19 +1,99 @@
-export interface Figure {
+type Shape = 'triangle' | 'circle' | 'rectangle';
+type Color = 'red' | 'green' | 'blue';
+
+const error = new Error('somethig went wrong');
```
naziamoff, line 4: no, you want your errors to be maximally descriptive. I as a user comes here - passed irrelevant input. I don't know what I did wrong. But I should know

src/figuresClasses.ts
```diff
@@ -1,19 +1,99 @@
...
-export class Circle {
+  isValidTriangle(): boolean {
+    const { a, b, c } = this;
 
+    return a > 0
+      && b > 0
+      && c > 0
+      && a < b + c
+      && b < a + c
+      && c < a + b;
```
naziamoff, line 34: should be separated into two different error messages

naziamoff (changes requested):
good job 🔥 
Let's fix the errors issue and we're good

33Orange (approved):
Well done !

PR #1247, opened 2023-05-31
solution

kirilshorin:


src/figuresClasses.ts
```diff
@@ -1,19 +1,84 @@
...
+
+  isValid(): boolean {
+    const { a, b, c } = this;
+
+    return a > 0
+      && b > 0
+      && c > 0
+      && a < b + c
+      && b < a + c
+      && c < a + b;
```
naziamoff, line 33: let's separate this logic to two different errors, and make the error text more descriptive

src/figuresClasses.ts
```diff
@@ -1,19 +1,84 @@
...
+
+  constructor(
+    public color: Colors,
+    public a:number,
+    public b:number,
+    public c:number,
+  ) {
+    if (!this.isValid()) {
+      throw new Error('Not a valid triangle');
+    }
```
naziamoff, line 22: you can hanlde error throwing right inside the isValid method.
Just have it in the constructor like `this.isValid()` - and everything will be done there

naziamoff (approved):
good job 👍 
See some minor suggestions

PR #1248, opened 2023-05-31
add solution

niaYaro:


src/figuresClasses.ts
```diff
@@ -1,19 +1,94 @@
...
-}
+  constructor(
+    public color: string,
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    this.shape = 'triangle';
+
+    function properTriangle(
```
naziamoff, line 18: functions always start with a verb, the naming is not clear here

src/figuresClasses.ts
```diff
@@ -1,19 +1,94 @@
...
+      aSide: number,
+      bSide: number,
+      cSide: number,
+    ): boolean {
+      const allSides = [aSide, bSide, cSide];
+      const longgestSide = Math.max(aSide, bSide, cSide);
```
naziamoff, line 24:
```suggestion
      const longestSide = Math.max(...allSides);
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,94 @@
...
+    public c: number,
+  ) {
+    this.shape = 'triangle';
+
+    function properTriangle(
+      aSide: number,
+      bSide: number,
+      cSide: number,
+    ): boolean {
+      const allSides = [aSide, bSide, cSide];
```
naziamoff, line 23: you can also sort them right away, and the max side would be either the first element of the array or the last one, depending on the asc/desc sorting

src/figuresClasses.ts
```diff
@@ -1,19 +1,94 @@
...
+        .reduce((total, side) => total + side);
 
-export class Circle {
+      return longgestSide < sumOfshorterSides;
+    }
 
+    if (this.a <= 0
+      || this.b <= 0
+      || this.c <= 0
+      || !properTriangle(this.a, this.b, this.c)
```
naziamoff, line 35: you should move all the conditions to the new line if you move at least one
```suggestion
    if (
      this.a <= 0
      || this.b <= 0
      || this.c <= 0
      || !properTriangle(this.a, this.b, this.c)
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,94 @@
...
+    ) {
+      throw new Error('Error: unacceptable parameter provided.');
+    }
```
naziamoff, line 38: I would handle this logic inside the validation function

src/figuresClasses.ts
```diff
@@ -1,19 +1,94 @@
...
+  }
+
+  getArea(): number {
+    const halfP = (this.a + this.b + this.c) / 2;
```
naziamoff, line 42: Is halfP 3.14 / 2?

naziamoff (changes requested):

src/figuresClasses.ts
```diff
@@ -1,19 +1,96 @@
...
-}
+  constructor(
+    public color: string,
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    this.shape = 'triangle';
+
+    function isItProperTriangle(
```
naziamoff, line 24:
```suggestion
    function isValidTriangle(
```

naziamoff (approved):
👍

PR #1249, opened 2023-05-31
made ts task figures classes

vladyslav-lapshyn:


src/figuresClasses.ts
```diff
@@ -1,19 +1,83 @@
...
+
 export interface Figure {
+  shape: Shapes;
+  color: Colors;
+  getArea(): number;
+}
 
+function getRoundNumber(area: number): number {
+  return Math.floor(area * 100) / 100;
 }
```
naziamoff, line 12: you don't just round the number here, let's be more specific
```suggestion
function getRoundedDownArea(area: number): number {
  return Math.floor(area * 100) / 100;
}
```

naziamoff (approved):
well done 🔥

PR #1250, opened 2023-06-01
Solution

Yuliia-kucheriavenko:


src/figuresClasses.ts
```diff
@@ -1,19 +1,82 @@
...
-}
+  constructor(
+    public color: Color,
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    if (a <= 0
+      || b <= 0
+      || c <= 0) {
```
naziamoff, line 20: each of the conditions should take a separate line if you decided to move at least one of them

naziamoff (changes requested):
not bad, but a couple of things to polish

src/figuresClasses.ts
```diff
@@ -1,19 +1,89 @@
...
+  color: Color;
+  getArea(): number;
+}
 
+function validateTriangle(a: number, b: number, c: number): void {
+  if (
+    a <= 0
+    || b <= 0
+    || c <= 0
+  ) {
```
naziamoff, line 15: pretty clear enough this way, I believe. Or you can create a variable for that - would be even better
```suggestion
  if (a <= 0 || b <= 0 || c <= 0) {
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,89 @@
...
+    throw new Error('Each side should have length greater than 0');
+  } else if (
```
naziamoff, line 17: I don't think you need else here, just if would be enough

src/figuresClasses.ts
```diff
@@ -1,19 +1,89 @@
+type Shape = 'triangle' | 'circle' | 'rectangle';
+type Color = 'red' | 'green' | 'blue';
+
 export interface Figure {
+  shape: Shape;
+  color: Color;
+  getArea(): number;
+}
 
+function validateTriangle(a: number, b: number, c: number): void {
```
naziamoff, line 10: 🔥

naziamoff (approved):
awesome!
Clean and clear 👍

PR #1251, opened 2023-06-01
add task solution

EugeneSherepa:


src/figuresClasses.ts
```diff
@@ -1,19 +1,90 @@
...
-export class Rectangle {
+    if (
+      this.sideA + this.sideB <= this.sideC
+      || this.sideA + this.sideC <= this.sideB
+      || this.sideB + this.sideC <= this.sideA
+    ) {
+      throw new Error(
+        'Any of the triangle`s side must be smaller than the sum of two other',
+      );
+    }
```
naziamoff, line 55: would be nice to move this logic to some `validateTriangle` method

naziamoff (approved):
Great job! 🔥

PR #1253, opened 2023-06-01
figuresClasses solution

PANCH1811:


src/figuresClasses.ts
```diff
@@ -1,19 +1,79 @@
...
+    }
+  }
 
-export class Circle {
+  getArea(): number {
+    const p = (this.a + this.b + this.c) / 2;
 
+    return Math.floor(Math.sqrt(p * (p - this.a)
+      * (p - this.b) * (p - this.c)) * 100) / 100;
+  }
```
naziamoff, line 37: I would come up with something like that - to make it clear what happens on each step
```suggestion
    return Math.floor(
      Math.sqrt(
        p * (p - this.a) * (p - this.b) * (p - this.c)
      ) * 100) / 100
    };
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,79 @@
...
+  public shape: Shape = 'triangle';
 
-}
+  constructor(
+    public color: Color,
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    if (this.a <= 0 || this.b <= 0 || this.c <= 0) {
```
naziamoff, line 19: also, would be nice to move some piece of logic to a separate method inside the class and just call it in the constructor

naziamoff (changes requested):
not bad, but some changes required to get it polished 💅

src/figuresClasses.ts
PANCH1811: image

silchyn: @PANCH1811 so what happened when you followed these instructions?

src/figuresClasses.ts
```diff
@@ -1,19 +1,80 @@
...
+      throw new Error('a, b and c must be > 0');
+    }
+
+    const sides = [this.a, this.b, this.c];
+    const longestSide = Math.max(...sides);
+    const isValidTriangle = sides
+      .reduce((sum, side) => sum + side, 0) - longestSide <= longestSide;
```
silchyn, line 26:
```suggestion
    const longestSide = Math.max(this.a, this.b, this.c);
    const perimeter = this.a + this.b + this.c;
    const isValidTriangle = perimeter - longestSide <= longestSide;
```

src/figuresClasses.ts
PANCH1811: @silchyn a comma will destroy a mathematical expression, and adding braces- the linter asks for more braces, why?

PANCH1811: nvm, it somehow works with that comma...

silchyn (changes requested):

silchyn (changes requested):

src/figuresClasses.ts
```diff
@@ -1,19 +1,82 @@
...
+    }
+
+    const sides = [this.side1, this.side2, this.side3];
+    const longestSide = Math.max(...sides);
+    const perimeter = this.side1 + this.side2 + this.side3;
+    const isValidTriangle = perimeter - longestSide <= longestSide;
+
+    if (isValidTriangle) {
+      throw new Error('Side1 + Side2 should be greater than Side3');
+    }
```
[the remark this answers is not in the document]
PANCH1811: sry, now it is.

src/figuresClasses.ts
```diff
@@ -1,19 +1,89 @@
...
+export class Circle implements Figure {
+  public shape: Shape = 'circle';
 
+  constructor(
+    public color: Color,
+    public radius: number,
+  ) {
+    if (this.radius <= 0) {
+      throw new Error('Radius should be > 0');
+    }
```
silchyn, line 59: > also, would be nice to move some piece of logic to a separate method inside the class and just call it in the constructor

src/figuresClasses.ts
```diff
@@ -1,19 +1,89 @@
...
+  public shape: Shape = 'rectangle';
+
+  constructor(
+    public color: Color,
+    public width: number,
+    public height: number,
+  ) {
+    if (this.width < 0 || this.height < 0) {
+      throw new Error('width and height should be > 0');
+    }
```
silchyn, line 77: > also, would be nice to move some piece of logic to a separate method inside the class and just call it in the constructor

silchyn (changes requested):

silchyn (approved):

PR #1254, opened 2023-06-01
add solution

ulositska:


src/figuresClasses.ts
```diff
@@ -1,19 +1,81 @@
...
+    public b: number,
+    public c: number,
+  ) {
+    if (a <= 0 || b <= 0 || c <= 0) {
+      throw new Error('Side length is not valid');
+    }
+
+    if (a >= b + c || b >= a + c || c >= a + b) {
+      throw new Error('sum of two sides must always be greater than the third');
+    }
```
naziamoff, line 29: would be nice to have a separate method like `validateTriangle` for that purpose

naziamoff (approved):
good job 🔥

PR #1255, opened 2023-06-01
add FiguresClasses solution

vchervinko:


src/figuresClasses.ts
```diff
@@ -1,19 +1,79 @@
...
+  ) {
+    if (a <= 0 || b <= 0 || c <= 0) {
+      throw new Error('triangle cannot exist');
+    }
 
-export class Circle {
+    if (a >= b + c || b >= a + c || c >= a + b) {
+      throw new Error('triangle cannot exist');
+    }
+  }
```
naziamoff, line 26: - the error messages must be more descriptive
- as you have two errors to throw - it is definitely better to create some separate method like `validateTriangle` or sth

src/figuresClasses.ts
```diff
@@ -1,19 +1,79 @@
...
 
+  getArea(): number {
+    const s = 0.5 * (this.a + this.b + this.c);
+    const Trianglearea = Math.sqrt(
```
naziamoff, line 30: should this be in upper case?

src/figuresClasses.ts
```diff
@@ -1,19 +1,79 @@
...
-export class Rectangle {
+export class Circle implements Figure {
+  shape: Shape = 'circle';
+
+  constructor(
+    public color: Color,
+    public radius: number,
+  ) {
+    if (radius <= 0) {
+      throw new Error('circle cannot exist');
```
naziamoff, line 46: more descriptive pls. Like `Radius can not be less than 0` or so

naziamoff (changes requested):
not bad, but let's polish a little

naziamoff (approved):
🔥

PR #1256, opened 2023-06-01
 ts_figures_classes_solution

YelyzavetaMolochynska:


src/figuresClasses.ts
```diff
@@ -1,19 +1,77 @@
+type Shape = 'triangle' | 'circle' | 'rectangle';
 export interface Figure {
-
+  shape: Shape;
+  color: 'red' | 'green' | 'blue';
```
Volodymyr-Melnyk-Fron10der, line 4: move it to the type and reuse throughout the code

Volodymyr-Melnyk-Fron10der (changes requested):

polosanya (approved):
Nice!

PR #1258, opened 2023-06-01
figure classes

YeMyr:


src/figuresClasses.ts
```diff
@@ -1,19 +1,80 @@
...
 
-export class Circle {
+  getArea(): number {
+    const p: number = (this.sideA + this.sideB + this.sideC) / 2;
 
+    return Math.floor(
+      Math.sqrt(p * (p - this.sideA) * (p - this.sideB) * (p - this.sideC))
+      * 100,
+    ) / 100;
+  }
```
naziamoff, line 40: the formatting looks wrong a little. Think of sth to make it more readable

src/figuresClasses.ts
```diff
@@ -1,19 +1,80 @@
...
+    public sideC: number,
+  ) {
+    if (sideA <= 0 || sideB <= 0 || sideC <= 0) {
+      throw new Error('No one side cannot be less than or equal to 0');
+    }
+
+    const sides: number[] = [sideA, sideB, sideC].sort(
+      (current, next) => next - current,
+    );
+
```
naziamoff, line 26: would be nice to move this logic to a separate method in the constructor

naziamoff (approved):
great job 👍

src/figuresClasses.ts
YeMyr: packed up everything into validation function (method), please, check if it is ok

src/figuresClasses.ts
YeMyr: I've made changes acceptable for eslint, please, take a look.

PR #1259, opened 2023-06-01
Solution

PanyaPrimal:


src/figuresClasses.ts
```diff
@@ -1,19 +1,81 @@
...
+    public c: number,
+  ) {
+    if (a <= 0 || b <= 0 || c <= 0) {
+      throw new Error('Length of sides must be greater than 0');
+    }
+
+    if (a + b <= c || b + c <= a || c + a <= b) {
+      throw new Error(
+        `Sides ${this.a}, ${this.b} and ${this.c} cannot form a triangle`,
+      );
```
naziamoff, line 26: good job - clear and concise 🔥 
To make it perfect - move this logic to a separate method and just call this method inside the constructor

naziamoff (approved):
awesome job 🔥 
see one suggestion

PR #1263, opened 2023-06-01
Solution

Mykola-Maik:


naziamoff (changes requested):
not bad, but let's be more specific in our errors. It must be maximally clear what is wrong

33Orange (approved):
Well done !

PR #1264, opened 2023-06-01
Solution

maryanmats:


src/figuresClasses.ts
```diff
@@ -1,19 +1,78 @@
-export interface Figure {
+type FigureShape = 'triangle' | 'circle' | 'rectangle';
+
+type FigureColor = 'red' | 'green' | 'blue';
```
naziamoff, line 3: as it is not stored in some constants file and you have this file only internally in the file - makes no sense to be that specific in the type naming
```suggestion
type Shape = 'triangle' | 'circle' | 'rectangle';

type Color = 'red' | 'green' | 'blue';
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,78 @@
...
+    }
+
+    if (a >= b + c || b >= a + c || c >= a + b) {
+      throw new Error('sum of two sides must always be greater than the third');
+    }
+  }
 
-export class Circle {
+  getArea(): number {
+    const s = 0.5 * (this.a + this.b + this.c);
```
naziamoff, line 30: use division if you want to divide, not multiplication)

naziamoff (approved):
good job 👍 

See some minor suggestions

PR #1267, opened 2023-06-02
solution

SofiaLiubchenko:


src/figuresClasses.ts
```diff
@@ -1,19 +1,77 @@
...
+  ) {
+    const sides = [this.a, this.b, this.c];
+    const longestSide = Math.max(this.a, this.b, this.c);
+
+    const twoOther = sides.filter((side) => side !== longestSide)
+      .reduce((side, sum) => side + sum);
+
+    if (twoOther <= longestSide || sides.some((num) => num <= 0)) {
+      throw new Error('Not valid data');
+    }
```
33Orange, line 27: In your code, 2 different errors display the same message. These are different errors.
Use more descriptive messages for errors
```suggestion
    const sumOfTwoLowerSides = sides
      .filter((side) => side !== longestSide)
      .reduce((side, sum) => side + sum);

    if (sides.some((num) => num <= 0)) {
      throw new Error('All sides have to be greater than zero');
    }
    if (sumOfTwoLowerSides <= longestSide) {
      throw new Error('The longest side of a triangle is >= than a sum of two others',);
    }
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,77 @@
...
-export class Rectangle {
+export class Circle implements Figure {
+  public shape: Figure['shape'] = 'circle';
 
+  constructor(
+    public color: Color,
+    public radius: number,
+  ) {
+    if (this.radius <= 0) {
+      throw new Error('Not valid data');
```
33Orange, line 46: Same here. Use more descriptive messages for errors

src/figuresClasses.ts
```diff
@@ -1,19 +1,77 @@
...
+export class Rectangle implements Figure {
+  public shape: Figure['shape'] = 'rectangle';
+
+  constructor(
+    public color: Color,
+    public width: number,
+    public height: number,
+  ) {
+    if (this.width <= 0 || this.height <= 0) {
+      throw new Error('Not valid data');
```
33Orange, line 66: Same here

33Orange (changes requested):
Almost done

PR #1269, opened 2023-06-02
solution

thienlaoo:


src/figuresClasses.ts
```diff
@@ -1,19 +1,93 @@
...
 export class Triangle {
+  color: string;
+
+  shape: string;
+
+  a: number;
+
+  b: number;
+
+  c: number;
```
naziamoff, line 16: no need for those empty lines in between
```suggestion
  color: string;
  shape: string;
  a: number;
  b: number;
  c: number;
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,93 @@
 export interface Figure {
-
+  shape: 'triange' | 'circle' | 'rectangle';
+  color: 'red' | 'green' | 'blue';
```
naziamoff, line 3: would be better to create enums for that - it is much easier to operate with enums than with string values

src/figuresClasses.ts
```diff
@@ -1,19 +1,93 @@
...
+    this.b = b;
+    this.c = c;
+
+    if (a <= 0 || b <= 0 || c <= 0) {
+      throw new Error('length is < than 0');
+    }
+
+    if (a + b <= c || a + c <= b || b + c <= a) {
+      throw new Error('sides cant form a triangle');
+    }
```
naziamoff, line 31: would be great to create a separate method for that and just call it inside the constructor

src/figuresClasses.ts
```diff
@@ -1,19 +1,93 @@
...
+
+    return parseFloat((Math.sqrt(s * (s - this.a)
+      * (s - this.b) * (s - this.c))).toFixed(2));
+  }
 }
 
 export class Circle {
+  color: string;
+
+  shape: string;
```
naziamoff, line 45: redundant empty lines

src/figuresClasses.ts
```diff
@@ -1,19 +1,93 @@
...
+    }
+  }
+
+  getArea(): number {
+    return Math.floor((this.radius ** 2 * Math.PI) * 100) / 100;
+  }
 }
 
 export class Rectangle {
+  color: string;
```
naziamoff, line 67: redundant empty lines

naziamoff (approved):
well done 👍 
See some suggestions on how it can get even better

PR #1277, opened 2023-06-11
Figure

Tetiana-Hishchak:


src/figuresClasses.ts
```diff
@@ -1,19 +1,76 @@
...
 
+export interface Figure {
+  shape:Shape;
+  color:Color;
+  getArea():number;
 }
 
-export class Triangle {
+export class Triangle implements Figure {
+  public shape:Shape ='triangle';
```
VitaliyBondarenko1982, line 11:
```suggestion
  public shape: Shape = 'triangle';
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,76 @@
-export interface Figure {
+type Shape = 'triangle' | 'circle' | 'rectangle';
+type Color = 'red' | 'green' | 'blue';
 
+export interface Figure {
+  shape:Shape;
+  color:Color;
+  getArea():number;
```
VitaliyBondarenko1982, line 7:
```suggestion
  shape: Shape;
  color: Color;
  getArea(): number;
```
please check code style in all cases

src/figuresClasses.ts
```diff
@@ -1,19 +1,76 @@
...
+        than the sum of other two`);
+    }
+  }
 
-export class Circle {
+  getArea():number {
+    const p :number = (this.a + this.b + this.c) / 2;
+    const s :number = p * (p - this.a) * (p - this.b) * (p - this.c);
 
+    return Math.floor(Math.sqrt(s) * 100) / 100;
```
VitaliyBondarenko1982, line 33:
```suggestion
    const { a, b, c } = this;
    const halfPerimeter = (a + b + c) / 2;
    const area = p * (p - a) * (p - b) * (p - c);

    return Math.floor(Math.sqrt(s) * 100) / 100;
```
consider to destructure this.
better to make more meaningful naming.
redundant to set directly type for halfPerimeter and area.
typescript knows that a, b, c are numbers and that result will be a number.

VitaliyBondarenko1982 (changes requested):
Good job.
Please check a few comments.

src/figuresClasses.ts
```diff
@@ -1,19 +1,79 @@
...
+      throw new Error(`The longest side of a triangle cannot be greater
+        than the sum of other two`);
+    }
+  }
 
-export class Circle {
+  getArea(): number {
+    const { a, b, c } = this;
+    const halfPerimeter: number = (a + b + c) / 2;
+    const area: number = Math.sqrt(halfPerimeter * (halfPerimeter - a
```
VitaliyBondarenko1982, line 32:
```suggestion
    const halfPerimeter = (a + b + c) / 2;
    const area = Math.sqrt(halfPerimeter * (halfPerimeter - a
```
typescript already knows that it is numbers

VitaliyBondarenko1982 (approved):
Well done

PR #1278, opened 2023-06-11
solution

Serhii-Shtokalo:


src/figuresClasses.ts
```diff
@@ -1,19 +1,80 @@
...
 export class Rectangle {
+  shape:Shape = 'rectangle';
 
-}
+  constructor(
+    public color:Color,
+    public width:number,
+    public heigth: number,
+  ) {
+    if (this.width <= 0 || this.heigth <= 0) {
```
DarkMistyRoom, line 68: In each case, you check if all arguments are positive. So it would be good to create a function for this

DarkMistyRoom (approved):
Good job!

PR #1279, opened 2023-06-12
solution

LiliiaDanylenko:


src/figuresClasses.ts
```diff
@@ -1,19 +1,79 @@
...
+type Color = 'red' | 'green' | 'blue';
 
+export interface Figure {
+  shape: Shape;
+  color: Color;
+  getArea(): number;
 }
 
-export class Triangle {
+function isPositiveNumber(figure: object): never | void {
```
OleziO, line 10: Remove `never` type

OleziO (changes requested):

volodymyr-soltys97 (approved):
Nice work 🔥

PR #1280, opened 2023-06-12
add task solution

RomanOstrous:


src/figuresClasses.ts
```diff
@@ -1,19 +1,71 @@
...
+    return Math.floor(area * 100) / 100;
+  }
 }
 
 export class Circle {
+  shape: Shape = 'circle';
 
+  constructor(
+    public color: Color,
+    public radius: number,
```
polosanya, line 39: It is better to make it private

src/figuresClasses.ts
```diff
@@ -1,19 +1,71 @@
...
 }
 
 export class Rectangle {
+  shape: Shape = 'rectangle';
 
-}
+  constructor(
+    public color: Color,
+    public width: number,
+    public height: number,
```
polosanya, line 57: Same here

src/figuresClasses.ts
```diff
@@ -1,19 +1,71 @@
...
 
 export class Circle {
+  shape: Shape = 'circle';
 
+  constructor(
+    public color: Color,
+    public radius: number,
+  ) {
+    if (radius <= 0) {
+      throw new Error('the radius cannot be less than ab or equal to 0');
```
polosanya, line 42:
```suggestion
      throw new Error('the radius cannot be less or equal to 0');
```

polosanya (changes requested):

agniya-i (approved):

PR #1284, opened 2023-06-13
done

Alessandranitti15:


src/figuresClasses.ts
```diff
@@ -1,19 +1,106 @@
 export interface Figure {
-
+  shape: 'triangle' | 'circle' | 'rectangle';
+  color: 'red' | 'green' | 'blue';
```
Anastasiia-Svintsova, line 3: Create separate types for `shape` and `color`. You'll need to use them then. Seems that you can even use enum here

src/figuresClasses.ts
```diff
@@ -1,19 +1,106 @@
...
+  getArea(): number;
 }
 
-export class Triangle {
+export class Triangle implements Figure {
+  shape: 'triangle' | 'circle' | 'rectangle';
```
Anastasiia-Svintsova, line 8: The shape should be 'triangle' for the Triangle class

src/figuresClasses.ts
```diff
@@ -1,19 +1,106 @@
...
 
-}
+  color: 'red' | 'green' | 'blue';
```
Anastasiia-Svintsova, line 10: As I said above, it's better to create a type for the Color
```suggestion
  color: Color;
```

Anastasiia-Svintsova (changes requested):
Good start! Check my comments to avoid repeatable code

lerastarynets (approved):
All comments fixed. You're doing amazing, keep it up!)

PR #1288, opened 2023-06-14
Implement_figure_classes

rialleons:


src/figuresClasses.ts
```diff
@@ -1,19 +1,88 @@
-export interface Figure {
+type Shape = 'triangle' | 'circle' | 'rectangle';
+type Color = 'red' | 'green' | 'blue';
 
+export interface Figure {
+  shape: Shape;
+  color: Color;
+  area: number;
 }
```
mykhalenych, line 8: The Figure interface should have a getArea method instead of the area property.

src/figuresClasses.ts
```diff
@@ -1,19 +1,88 @@
...
+    if (a <= 0 || b <= 0 || c <= 0) {
+      throw new Error('One of the sides is lower than zero');
+    }
+
+    if (a + b <= c || a + c <= b || b + c <= a) {
+      throw new Error('Longest side is bigger than the sum of two others');
+    }
+
+    this.shape = 'triangle';
+    this.area = this.getArea();
```
mykhalenych, line 30: Don't store the area as a property in the constructor, as it's not required.

src/figuresClasses.ts
```diff
@@ -1,19 +1,88 @@
...
+export interface Figure {
+  shape: Shape;
+  color: Color;
+  area: number;
 }
 
 export class Triangle {
+  shape: Shape;
+
+  area: number;
```
mykhalenych, line 13: Remove the area property from the Triangle class, since it should be defined in the Figure interface.

src/figuresClasses.ts
```diff
@@ -1,19 +1,88 @@
...
+    const s = 0.5 * (a + b + c);
 
+    return Math.floor(Math.sqrt(s * (s - a) * (s - b) * (s - c)) * 100) / 100;
+  }
 }
 
 export class Circle {
+  shape: Shape;
 
+  area: number;
```
mykhalenych, line 44: Remove the area property from the Circle class, since it should be defined in the Figure interface.

src/figuresClasses.ts
```diff
@@ -1,19 +1,88 @@
...
+    public color: Color,
+    public width: number,
+    public height: number,
+  ) {
+    if (width <= 0 || height <= 0) {
+      throw new Error('One of the sides is lower than zero');
+    }
+
+    this.shape = 'rectangle';
+    this.area = this.getArea();
```
mykhalenych, line 78: Don't store the area as a property in the constructor, as it's not required.

etojeDenys (approved):
lgtm

PR #1290, opened 2023-06-14
Figures Classes

ilona-romaniuk:


src/figuresClasses.ts
```diff
@@ -1,19 +1,88 @@
...
+        .filter((value) => typeof value === 'number')
+        .sort((x, y) => x - y);
 
-export class Circle {
+    if (
+      Object.values(this).some((value) => typeof value === 'number'
+      && value <= 0)
+      || sides[0] + sides[1] <= sides[2]
+    ) {
+      throw new Error('Invalid sides values');
```
polosanya, line 29: It would be great to specify why exactly the sides are invalid
image

polosanya (changes requested):
Almost done! Let's try to make your code cleaner and easier to read

witflash (approved):
Well done 👍

PR #1292, opened 2023-06-15
solution

IzzetYusufov:


src/figuresClasses.ts
```diff
@@ -1,19 +1,79 @@
 export interface Figure {
-
+  color: string,
+  getArea(): number,
+  a?: number,
```
mykhalenych, line 4: Remove optional properties a, b, c, radius, width, and height from the Figure interface, as these properties should be specific to the respective classes.

src/figuresClasses.ts
```diff
@@ -1,19 +1,79 @@
...
+      if (this.b + this.c > this.a) {
+        if (this.a + this.c > this.b) {
+          const perimetr = (this.a + this.b + this.c) / 2;
 
+          return Math.floor(Math.sqrt(perimetr * (perimetr - this.a)
+          * (perimetr - this.b) * (perimetr - this.c)) * 100) / 100;
+        }
+      }
+    }
+
```
mykhalenych, line 36: Remove this line, as the error will be thrown in the constructor after suggested changes.

src/figuresClasses.ts
```diff
@@ -1,19 +1,79 @@
...
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    this.getArea();
+  }
 
-export class Circle {
+  getArea():number {
+    if (this.a + this.b > this.c) {
```
mykhalenych, line 26: Move the triangle validation logic to the constructor. Throw an error if the triangle is invalid, and remove the checks in the getArea method.

src/figuresClasses.ts
```diff
@@ -1,19 +1,79 @@
...
+  shape: string = 'triangle';
 
-}
+  constructor(
+    public color: string,
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    this.getArea();
```
mykhalenych, line 22: Remove the call to this.getArea() in the constructor, as it is not needed here.

src/figuresClasses.ts
```diff
@@ -1,19 +1,79 @@
...
+    public width: number,
+    public height: number,
+  ) {
+    if (this.width <= 0 || this.height <= 0) {
+      throw new Error(`It's not a ${this.shape}`);
+    }
+  }
+
+  getArea():number {
+    return this.width * this.height;
```
mykhalenych, line 72: Round down the area to hundredths as required by the task. Change the return statement to: `return Math.floor(this.width * this.height * 100) / 100;`

mykhalenych (changes requested):

src/figuresClasses.ts
```diff
@@ -1,19 +1,73 @@
...
-export class Rectangle {
+export class Circle implements Figure {
+  shape: string = 'circle';
+
+  constructor(
+    public color: string,
+    public radius: number,
+  ) {
+    if (this.radius <= 0) {
+      throw new Error(`It's not a ${this.shape}`);
```
volodymyr-soltys97, line 43: Change the error message to 'The circle's radius can't be less than or equal to 0'

src/figuresClasses.ts
```diff
@@ -1,19 +1,73 @@
...
+export class Rectangle implements Figure {
+  shape: string = 'rectangle';
+
+  constructor(
+    public color: string,
+    public width: number,
+    public height: number,
+  ) {
+    if (this.width <= 0 || this.height <= 0) {
+      throw new Error(`It's not a ${this.shape}`);
```
volodymyr-soltys97, line 61: Change the error message to 'The rectangle's width and height can't be less than or equal to 0'

volodymyr-soltys97 (approved):
Great job 👍

PR #1293, opened 2023-06-15
add task solution

sasdaiv:


src/figuresClasses.ts
```diff
@@ -1,19 +1,75 @@
...
+    public width: number,
+    public height: number,
+  ) {
+    if (width <= 0 || height <= 0) {
+      throw new Error('width or height cannot be less or equal to 0');
+    }
+  }
+
+  getArea(): number {
+    return this.width * this.height;
```
mykhalenych, line 69: better will be round down the area to hundredths before returning it

PR #1304, opened 2023-06-20
'done'

Hanna-Balabukha:


src/figuresClasses.ts
```diff
@@ -1,19 +1,79 @@
...
 
-}
+  constructor(
+    public color: Color,
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    if (a <= 0 || b <= 0 || c <= 0) {
+      throw new Error('Error');
```
Anastasiia-Svintsova, line 20: The error text must contain some meaning, not just `Error`.  Fix it everywhere 
For example:
```suggestion
      throw new Error('Invalid side length. Side length must be greater than 0.' );
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,79 @@
...
+    }
+  }
 
+  getArea(): number {
+    const square = 0.5 * (this.a + this.b + this.c);
+    const area = Math.sqrt(square * (square - this.a)
+    * (square - this.b) * (square - this.c));
+
+    return +area.toFixed(2);
```
Anastasiia-Svintsova, line 33: use Math.floor here as well
image

Anastasiia-Svintsova (changes requested):
Good start! Check my comments

sTorba24 (approved):
Well done!

PR #1311, opened 2023-06-24
add task solution

MelnychukVol:


src/figuresClasses.ts
```diff
@@ -1,19 +1,77 @@
...
 
-export class Circle {
+  getArea(): number {
+    const perimeter = (this.a + this.b + this.c) / 2;
+    const triangleArea = Math.sqrt(perimeter
+      * (perimeter - this.a)
+      * (perimeter - this.b)
+      * (perimeter - this.c));
 
+    return Math.floor(triangleArea * 100) / 100;
```
volodymyr-soltys97, line 36: The area should be rounded down to hundredths. Use `Math.round(triangleArea * 100) / 100` instead of `Math.floor(triangleArea * 100) / 100`.

volodymyr-soltys97 (approved):
Good work 👍

PR #1317, opened 2023-06-28
Solution

Nataliia95254:


src/figuresClasses.ts
```diff
@@ -1,19 +1,93 @@
 export interface Figure {
-
+  shape: 'triangle' | 'circle' | 'rectangle';
+  color: 'red' | 'green' | 'blue';
```
Anastasiia-Svintsova, line 3: Create types for Shape and Color to not repeat it each time. BTW here is a good place for enums

src/figuresClasses.ts
```diff
@@ -1,19 +1,93 @@
...
+  getArea(): number;
 }
 
-export class Triangle {
+export class Triangle implements Figure {
+  shape: 'triangle' | 'circle' | 'rectangle' = 'triangle';
```
Anastasiia-Svintsova, line 8: Let's think if we need `'circle' | 'rectangle'` for the Triangle class. The same comment goes for other classes

src/figuresClasses.ts
```diff
@@ -1,19 +1,93 @@
...
+    public c: number,
+  ) {
+    this.color = color;
+    this.a = a;
+    this.b = b;
+    this.c = c;
+    this.area = this.getArea();
+
+    if (a + b <= c || a + c <= b || b + c <= a) {
+      throw new Error('your error message');
```
Anastasiia-Svintsova, line 25: Add a correct error message everywhere

Anastasiia-Svintsova (changes requested):

src/figuresClasses.ts
```diff
@@ -1,19 +1,109 @@
...
+  constructor(
+    public color: Color,
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    this.color = color;
+    this.a = a;
+    this.b = b;
+    this.c = c;
```
TarasHoliuk, line 34: FYI - no need to define it here if you declare `public` in the constructor arguments

src/figuresClasses.ts
```diff
@@ -1,19 +1,109 @@
...
+    if (width <= 0 || height <= 0) {
+      throw new Error(`Width ${this.width} or height
+      ${this.height} can't be 0 or less than 0`);
+    }
+  }
+
+  getArea(): number {
+    const s = this.width * this.height;
+
+    return Math.floor(s * 100) / 100;
```
TarasHoliuk, line 103: you are doing the same rounding 3 times - so it's a candidate to create a separate function for

TarasHoliuk (approved):
Added 2 comments just FYI
GJ :+1:  Approved

PR #1320, opened 2023-06-29
add task solution

VolodymyrPst:


src/figuresClasses.ts
```diff
@@ -1,19 +1,72 @@
 export interface Figure {
-
-}
+  shape: string,
+  area: number,
+  color: string,
+};
 
 export class Triangle {
```
agniya-i, line 7: `Triangle` should implement `Figure`
Same for other classes

src/figuresClasses.ts
```diff
@@ -1,19 +1,72 @@
...
+  readonly shape: string = 'Triangle';
+  private area: number = 0;
+  constructor(
+    readonly color: 'red' | 'green' | 'blue',
+    readonly a: number,
+    readonly b: number,
+    readonly c: number,
+  ) {
+    if (a <= 0 || b <= 0 || c <= 0) {
+      throw new Error('Wrong value!');
```
agniya-i, line 17: Give more details in error message

src/figuresClasses.ts
```diff
@@ -1,19 +1,72 @@
 export interface Figure {
-
-}
+  shape: string,
```
agniya-i, line 2: It would be good to create a type for shape and color, for example 

`type Colors = 'red' | 'green' | ... `

agniya-i (changes requested):
Please fix all linter errors and pay attention to the comments below

volodymyr-soltys97 (approved):
Good job 👍

PR #1321, opened 2023-06-29
figures classes

AllaNebozhenko:


src/figuresClasses.ts
```diff
@@ -1,19 +1,90 @@
...
+    public width: number,
+    public height: number,
+  ) {
+    if (!this.isRectangleValid()) {
+      throw new Error('Rectangle sides mus be a positive number');
+    }
+  }
+
+  getArea(): number {
+    return Math.round(this.width * this.height * 100) / 100;
```
TarasHoliuk, line 80: FYI: you are doing the same rounding 3 times - so it's a candidate to create a separate function for

TarasHoliuk (approved):
Approved :+1:

PR #1326, opened 2023-06-30
solution

vlad-fil:


src/figuresClasses.ts
```diff
@@ -1,19 +1,81 @@
-export interface Figure {
+type Color = 'red' | 'green' | 'blue';
 
+export interface Figure {
+  shape: string;
```
VitaliyBondarenko1982, line 4: make type for shape and use in all places

```suggestion
  shape: Shape;
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,81 @@
...
 
+    if (longestSide >= perimetr - longestSide) {
+      throw new Error(
+        'longest side of a triangle is bigger than a sum of two others',
+      );
+    }
+  }
+
+  getArea(): number {
+    const p = (this.a + this.b + this.c) / 2;
```
VitaliyBondarenko1982, line 33:
```suggestion
    const halfP = (this.a + this.b + this.c) / 2;
```
in fact it is half perimeter

VitaliyBondarenko1982 (changes requested):
Good job.
Please check a couple comments.

src/figuresClasses.ts
vlad-fil: ok, but [p](https://ru.wikipedia.org/wiki/%D0%A4%D0%BE%D1%80%D0%BC%D1%83%D0%BB%D0%B0_%D0%93%D0%B5%D1%80%D0%BE%D0%BD%D0%B0) it`s mean a halp of perimetr

VitaliyBondarenko1982 (approved):

src/figuresClasses.ts
VitaliyBondarenko1982: I just consider more meaningful variable for code.

> ok, but [p]

PR #1328, opened 2023-07-01
task done

tecHamster:


src/figuresClasses.ts
```diff
@@ -1,19 +1,72 @@
...
+    public b: number,
+    public c: number,
+  ) {
+    if (Math.min(a, b, c) <= 0) {
+      throw new Error('your error message');
+    }
+
+    const sizes = [a, b, c].sort((side1, side2) => side1 - side2);
+
+    if (sizes.pop() >= sizes[0] + sizes[1]) {
```
witflash, line 22: Why do you need to use “pop” method here?
If you know the size of an array you can just get a value directly by index: `sizes[2]`

witflash (approved):
Well done 👍🏻

PR #1332, opened 2023-07-04
solution

artem-ivashchenko:
I have 1 failing test, (Expected: 113.09 Received: 113.1), but i think it's not that important

### DIFF
--- a/src/figuresClasses.ts
+++ b/src/figuresClasses.ts
@@ -1,19 +1,101 @@
+type Shape = 'triangle' | 'circle' | 'rectangle';
+type Color = 'red' | 'green' | 'blue';
+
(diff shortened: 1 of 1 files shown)


src/figuresClasses.ts
```diff
@@ -1,19 +1,101 @@
...
+  constructor(public color: Color, ...args: [number, number]) {
+    checkNumbers(args);
+
+    [this.width, this.height] = args;
+  }
+
+  getArea(): number {
+    const formula = this.width * this.height;
+
+    return +formula.toFixed(2);
```
agniya-i, line 95: You can use `Math.round` to fix the failed test 😉

agniya-i (approved):

PR #1336, opened 2023-07-05
add task solution

Vayts:


src/figuresClasses.ts
```diff
@@ -1,19 +1,103 @@
...
+  }
+
+  getArea(): number {
+    const res = Math.PI * (this.radius ** 2) * 100;
+
+    return Math.floor(res) / 100;
+  }
 }
 
 export class Rectangle {
```
roman-mirzoian, line 73: The Rectangle class should implement Figure interface

src/figuresClasses.ts
```diff
@@ -1,19 +1,103 @@
...
+
+  constructor(color: Color, a: number, b: number, c: number) {
+    this.color = color;
+    this.a = a;
+    this.b = b;
+    this.c = c;
+
+    if (a <= 0 || b <= 0 || c <= 0) {
+      throw new Error(
+        'All numbers must be positive',
```
roman-mirzoian, line 29: This message is now repeated for each class, so it can be put into a separate variable and reused to avoid having to repeat the whole thing every time.

roman-mirzoian (changes requested):
A job well done but needs some corrections.

src/figuresClasses.ts
```diff
@@ -1,19 +1,103 @@
...
+        'The longest side of a triangle is >= than a sum of two others',
+      );
+    }
+  }
+
+  getArea(): number {
+    const s = (this.a + this.b + this.c) / 2;
+    const res = Math.sqrt(s * (s - this.a) * (s - this.b) * (s - this.c));
 
+    return Number(res.toFixed(2));
```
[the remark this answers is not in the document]
Vayts: Thank you for your review. method getArea should return typeof number=). toFixed returns a string so... Maybe I just don't understand what you mean.

BudnikOleksii (changes requested):
Nice start, but a few things need to be done, check the comments and fix in all places.

src/figuresClasses.ts
```diff
@@ -1,19 +1,105 @@
...
+
+  b: number;
+
+  c: number;
+
+  constructor(color: Color, a: number, b: number, c: number) {
+    this.color = color;
+    this.a = a;
+    this.b = b;
+    this.c = c;
```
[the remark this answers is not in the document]
Vayts: image
I cant do it...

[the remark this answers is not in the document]
Vayts: image

src/figuresClasses.ts
```diff
@@ -1,19 +1,106 @@
...
+  shape: Shape = Shape.Triangle;
+
+  color: Color;
+
+  a: number;
+
+  b: number;
+
+  c: number;
+
```
Moroz-Dmytro, line 30:
```suggestion
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,106 @@
...
+  constructor(color: Color, a: number, b: number, c: number) {
+    this.color = color;
+    this.a = a;
+    this.b = b;
+    this.c = c;
```
Moroz-Dmytro, line 35: try this 
```suggestion
  constructor(
    public color: Color,
    public a: number,
    public b: number,
    public c: number
  ) {
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,106 @@
...
+
+    if (a + c <= b || a + b <= c || c + b <= a) {
+      throw new Error(ERRORS.longestSide);
+    }
+  }
 
+  getArea(): number {
+    const { a, b, c } = this;
+    const square = (a + b + c) / 2;
+    const res = Math.sqrt(square * (square - a) * (square - b) * (square - c));
```
Moroz-Dmytro, line 49:
```suggestion
    const area = Math.sqrt(square * (square - a) * (square - b) * (square - c));
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,106 @@
...
+    }
+
+    if (a + c <= b || a + b <= c || c + b <= a) {
+      throw new Error(ERRORS.longestSide);
+    }
+  }
 
+  getArea(): number {
+    const { a, b, c } = this;
+    const square = (a + b + c) / 2;
```
Moroz-Dmytro, line 48: But it's half of perimeter

src/figuresClasses.ts
```diff
@@ -1,19 +1,106 @@
...
+  getArea(): number {
+    return Number((this.width * this.height).toFixed(2));
+  }
 }
 
-export function getInfo(figure) {
+export function getInfo(figure: Figure): string {
+  const square = figure.getArea();
 
+  return `A ${figure.color} ${figure.shape} - ${square}`;
```
Moroz-Dmytro, line 105: Use destructuring

src/figuresClasses.ts
Moroz-Dmytro: If it works, please fix in other places

Moroz-Dmytro (changes requested):

PR #1341, opened 2023-07-06
figure classes task

ValeraViachalo:


src/figuresClasses.ts
```diff
@@ -1,19 +1,84 @@
 export interface Figure {
+  shape: string;
+  color: string;
```
Viktor-Kost, line 3: it's better to create types for color and shape.
And whenever you get shape or color add created types. Check other cases

```suggestion
  shape: Shapes;
  color: Colors;
```

Viktor-Kost (changes requested):

ValeraViachalo (commented):
added types

mykhalenych (approved):

PR #1344, opened 2023-07-06
Solution

denlubn:


src/figuresClasses.ts
```diff
@@ -1,19 +1,84 @@
...
+export class Rectangle implements Figure {
+  public shape: FigureShape = FigureShape.rectangle;
+
+  constructor(
+    public color: FigureColor,
+    public width: number,
+    public height: number,
+  ) {
+    if (width <= 0 || height <= 0) {
+      throw new Error('throws an error');
```
volodymyr-soltys97, line 71: The error message should be more descriptive. Provide specific information on why the error occurred.

volodymyr-soltys97 (approved):
Nice work 🔥

PR #1346, opened 2023-07-07
solution

ivanbilotserkivskiy:


src/figuresClasses.ts
```diff
@@ -1,19 +1,76 @@
 export interface Figure {
-
+  shape: string,
```
And678, line 2: Create an [enum](https://www.typescriptlang.org/docs/handbook/enums.html) `Color`, so you don't have to mess with bare strings.

And678 (changes requested):
One thing to improve below:

And678 (approved):
:+1:  :+1:

PR #1348, opened 2023-07-07
solution

MariiaMonakhova:


src/figuresClasses.ts
```diff
@@ -1,19 +1,114 @@
...
+export class Triangle implements Figure {
+  color: string;
+
+  shape: string;
+
+  a: number;
+
+  b: number;
+
+  c: number;
```
And678, line 16: No need for so much newlines
```suggestion
  color: string;
  shape: string;
  a: number;
  b: number;
  c: number;
```

And678: Same for other classes

src/figuresClasses.ts
```diff
@@ -1,19 +1,114 @@
 export interface Figure {
-
+  shape: string;
+  color: string;
```
And678, line 3: Make [enums](https://www.typescriptlang.org/docs/handbook/enums.html) for color and shape, so you don't need to deal with bare strings

And678 (changes requested):
Very nice. A couple of things to improve:

qoa11a (approved):

PR #1351, opened 2023-07-07
add solution

maxyma-sanx:


src/figuresClasses.ts
```diff
@@ -1,19 +1,83 @@
 export interface Figure {
-
+  shape: 'triangle' | 'circle' | 'rectangle';
+  color: 'red' | 'green' | 'blue';
```
anstsot, line 3: let's introduce typings for `shape` and `color`. try to use `enum`

anstsot (changes requested):
Good start:four_leaf_clover:

PR #1353, opened 2023-07-07
add task figures classes

AnastasiiaKhemych:


src/figuresClasses.ts
```diff
@@ -1,19 +1,88 @@
...
 
+  getArea(): number;
 }
 
-export class Triangle {
+export class Triangle implements Figure {
+  shape: string;
 
-}
+  color: string;
```
anstsot, line 11: let's introduce typings for `shape` and `color`. try to use `enum`

src/figuresClasses.ts
```diff
@@ -1,19 +1,88 @@
...
+
+  b: number;
+
+  c: number;
 
-export class Circle {
+  constructor(color: string, a: number, b: number, c: number) {
+    if (a >= b + c || b >= a + c || c >= a + b || a <= 0 || b <= 0 || c <= 0) {
+      throw new Error('your error message');
+    }
```
anstsot, line 22: try to give more meaningful error messages

anstsot (changes requested):
Good start:four_leaf_clover:

PR #1354, opened 2023-07-07
solution

AnatoliiLykhohodenko:


src/figuresClasses.ts
```diff
@@ -1,19 +1,85 @@
 export interface Figure {
-
+  shape: string,
+  color: string,
+  a?: number;
+  b?: number;
+  c?: number;
+  radius?: number;
+  width?: number;
+  height?: number;
```
anstsot, line 9: these are unnecessary
```suggestion
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,85 @@
 export interface Figure {
-
+  shape: string,
+  color: string,
```
anstsot, line 3: let's introduce typings for `shape` and `color`. try to use `enum`

src/figuresClasses.ts
```diff
@@ -1,19 +1,85 @@
...
+
+  constructor(
+    public color: string,
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    // eslint-disable-next-line max-len
+    if (a + b <= c || a + c <= b || b + c <= a || a <= 0 || b <= 0 || c <= 0) {
+      throw new Error();
```
anstsot, line 24: give more meaningful error to let use understand what's wrong

src/figuresClasses.ts
```diff
@@ -1,19 +1,85 @@
...
+export function getInfo(figure: Figure): string {
+  switch (Object.getPrototypeOf(figure)) {
+    case Triangle.prototype:
+    case Circle.prototype:
+    case Rectangle.prototype:
+      return `A ${figure.color} ${figure.shape} - ${figure.getArea()}`;
 
+    default:
+      return 'Incorrect data';
+  }
```
anstsot, line 84:
```suggestion
  return `A ${figure.color} ${figure.shape} - ${figure.getArea()}`;
```

anstsot (changes requested):
Good start:four_leaf_clover:

src/figuresClasses.ts
AnatoliiLykhohodenko: I'm not sure but tests wouldn't pass with Colors enum because data is string 'green' instead of Colors.green

And678 (approved):
:+1:

PR #1355, opened 2023-07-07
Add task solution

vladmakohon:


src/figuresClasses.ts
```diff
@@ -1,19 +1,88 @@
-export interface Figure {
+type Shape = 'triangle' | 'circle' | 'rectangle';
```

anstsot (changes requested):
Good start:four_leaf_clover:
Let’s add a small improvement

And678 (approved):
Approve :heavy_check_mark:

PR #1357, opened 2023-07-08
ts_figures_classes  solution

NataliGru:


src/figuresClasses.ts
```diff
@@ -1,19 +1,97 @@
...
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    if (a <= 0
+      || b <= 0
+      || c <= 0
+      || a + b <= c
+      || b + c <= a
+      || c + a <= b) {
```
anstsot, line 33: it will be better to create separate errors to give more context

anstsot (approved):

meforoma (approved):
{technical approve}

PR #1361, opened 2023-07-08
figures_classes

DenisGurskiy:


volodymyr-soltys97 (changes requested):
Good job 👍 
But let's make your code perfect

OleziO (approved):

PR #1364, opened 2023-07-09
added task solution

vlopatko:


src/figuresClasses.ts
```diff
@@ -1,19 +1,93 @@
-export interface Figure {
+export type Shape = 'triangle' | 'circle' | 'rectangle';
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,93 @@
...
+
+    if (
+      a <= 0
+      || b <= 0
+      || c <= 0
+      || c >= a + b
+      || b >= a + c
+      || a >= b + c
+    ) {
+      throw new Error('Enter correct data');
```
anstsot, line 32: it will be better to give more meaningful errors to let user understand what's wrong

src/figuresClasses.ts
```diff
@@ -1,19 +1,93 @@
...
 
+  getArea(): number {
+    return getRounded(+(this.width * this.height).toFixed(2));
+  }
 }
 
-export function getInfo(figure) {
+export type MixFigure = Circle | Triangle | Rectangle;
 
+export function getInfo(figure: MixFigure): string {
```
anstsot, line 91: just use `Figure`
```suggestion
export function getInfo(figure: Figure): string {
```

anstsot (changes requested):
Good start:four_leaf_clover:

And678 (approved):
Good work, approve :heavy_check_mark:

PR #1365, opened 2023-07-09
ts_figures_classe

OlegRak86:


src/figuresClasses.ts
```diff
@@ -1,19 +1,94 @@
...
 
-export function getInfo(figure) {
+  getArea(): number {
+    return Math.floor(this.a * this.b * 100) / 100;
+  }
+}
 
+export function getInfo(figure: Figure): string {
+  switch (figure.shape) {
+    case 'circle':
```
OleziO, line 85: Move all cases to the enum.

OleziO (changes requested):

src/figuresClasses.ts
```diff
@@ -1,19 +1,99 @@
...
+  color: Color
+  getArea: ()=>{}
+}
+enum FigureForm {
+  Triangle = 'triangle',
+  Circle = 'circle',
+  Rectangle = 'rectangle'
+}
+
+function getTriangle(a:number, b:number, c:number) : boolean {
```
mykhalenych, line 14: The function name 'getTriangle' is not clear. It should be something like 'isTriangle' because it checks whether three given sides can form a triangle and returns a boolean.

src/figuresClasses.ts
```diff
@@ -1,19 +1,99 @@
+type Shape = 'triangle' | 'circle' | 'rectangle';
+type Color = 'red' | 'green' | 'blue';
 export interface Figure {
+  shape: 'triangle' | 'circle' | 'rectangle'
+  color: Color
+  getArea: ()=>{}
```
mykhalenych, line 6: The function 'getArea' in the interface 'Figure' needs to specify a return type. It should return a number.

mykhalenych (changes requested):

mykhalenych (approved):

PR #1368, opened 2023-07-09
solution

ovynarchuk:


src/figuresClasses.ts
```diff
@@ -1,19 +1,93 @@
...
 
+  getArea(): number {
+    return roundNumber(+(this.width * this.height).toFixed(2));
+  }
 }
 
-export function getInfo(figure) {
+export type MixedFigure = Circle | Rectangle | Triangle;
 
+export function getInfo(figure: MixedFigure): string {
```
And678, line 91: This is the reason for having `Figure`, so you can use all Figures without declaring a union type like this.
```suggestion
export function getInfo(figure: Figure): string {
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,93 @@
 export interface Figure {
-
+  shape: 'triangle' | 'circle' | 'rectangle';
+  color: 'red' | 'green' | 'blue';
```
And678, line 3: Declare separate `enums` or union types
```suggestion
  shape: Shape;
  color: Color;
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,93 @@
...
+export class Triangle implements Figure {
+  shape: 'triangle';
+
+  constructor(
+    public color: Figure['color'],
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    this.shape = 'triangle';
```
And678, line 20: After making a `Shape` enum
```suggestion
    this.shape = Shape.triangle;
```

And678 (changes requested):
Gj, but can be improved

YegorKorenkov (approved):
Looks good!

PR #1369, opened 2023-07-09
add solution

pavlokuraksin:


src/figuresClasses.ts
```diff
@@ -1,19 +1,80 @@
+type Color = 'red' | 'green' | 'blue';
+type Shape = 'triangle' | 'circle' | 'rectangle';
```
anstsot, line 2: let's create `enum Shape` and use it everywhere

anstsot (changes requested):
Good start:four_leaf_clover:
Let’s add a small improvement

And678 (approved):
Great, approve :heavy_check_mark:

PR #1370, opened 2023-07-09
Solution

eleonora-demikh:


anstsot (changes requested):
Good start:four_leaf_clover:
Let’s add small improvement

juljaVysotska (approved):
Good job!

PR #1371, opened 2023-07-09
add task solution

NazarBaraban:


anstsot (changes requested):
Good start:four_leaf_clover:

PR #1373, opened 2023-07-09
added task solution

tetianakostiuk1:


src/figuresClasses.ts
```diff
@@ -1,19 +1,78 @@
...
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    if ([a, b, c].some((s) => s <= 0)
+    || a + b <= c
+    || a + c <= b
+    || b + c <= a
+    ) {
+      throw new Error('Invalid triangle');
```
anstsot, line 26: it will be better to give more meaningful error messages to let user understand what's wrong

anstsot (approved):

PR #1374, opened 2023-07-09
add task solution

klmyana:


anstsot (changes requested):
Good start:four_leaf_clover:
Let’s add a small fix to improve your code.

PR #1381, opened 2023-07-10
Finished task

IgorOksentyuk:


src/figuresClasses.ts
```diff
@@ -1,19 +1,88 @@
-export interface Figure {
-
+interface Figure {
+  shape: string,
+  color: string,
```
And678, line 3: Make enums or union types here, to avoid using bare strings
```suggestion
  shape: Shape,
  color: Color,
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,88 @@
...
+  height: number;
+
+  constructor(color: string, width: number, height: number) {
+    if (width <= 0 || height <= 0) {
+      throw new Error('Width or heigth can not be lower 0 or 0!');
+    }
+    this.color = color;
+    this.width = width;
+    this.height = height;
+  }
```
And678, line 75: Use constructor initializers to avoid doing `this.color = color`
```suggestion
  constructor(
    public color: string, 
    public width: number, 
    public height: number
  ) {
    if (width <= 0 || height <= 0) {
      throw new Error('Width or heigth can not be lower 0 or 0!');
    }
  }
```

And678 (changes requested):
GJ, only a couple improvements left

And678 (approved):
:+1:

PR #1383, opened 2023-07-10
create figures-classes

kochetov-yevhen-may29:


src/figuresClasses.ts
```diff
@@ -1,19 +1,88 @@
...
 
-export class Rectangle {
+export class Circle implements Figure {
+  public shape = FigureShape.Circle;
+
+  constructor(
+    public color: FigureColor,
+    public circle: number,
+  ) {
+    if (circle < 0) {
```
And678, line 56: Radius?
```suggestion
    if (radius < 0) {
```

And678 (approved):
Great, approve :heavy_check_mark:

src/figuresClasses.ts
kochetov-yevhen-may29: )) yes, radius, i can't explane why i choose this word

PR #1384, opened 2023-07-10
task solution

Bonum-taurus:


src/figuresClasses.ts
```diff
@@ -1,19 +1,102 @@
...
+      throw new Error('your error message');
+    }
+
+    this.color = color;
+    this.width = width;
+    this.height = height;
+  }
+
+  getArea(): number {
+    const s: number = this.width * this.height;
```
TarasHoliuk, line 94:
```suggestion
    return this.width * this.height;
```
shorter way

src/figuresClasses.ts
```diff
@@ -1,19 +1,100 @@
...
-export class Rectangle {
+export class Circle implements Figure {
+  public shape: Shape = 'circle';
+
+  constructor(public color: Color, public radius: number) {
+    if (this.radius <= 0) {
+      throw new Error('the number is not positive');
+    }
+
+    this.color = color;
```
volodymyr-soltys97, line 65: The properties color, radius are already initialized due to public keyword in constructor, no need to reassign them.

src/figuresClasses.ts
```diff
@@ -1,19 +1,100 @@
...
+  ) {
+    if (this.a <= 0 || this.b <= 0 || this.c <= 0) {
+      throw new Error('the number is not positive');
+    }
+
+    if (isTriangle(this.a, this.b, this.c)) {
+      throw new Error('is not a triangle');
+    }
+
+    this.color = color;
```
volodymyr-soltys97, line 42: The properties color, a, b, c are already initialized due to public keyword in constructor, no need to reassign them.

src/figuresClasses.ts
```diff
@@ -1,19 +1,100 @@
...
+    public color: Color,
+    public width: number,
+    public height: number,
+  ) {
+    if (this.width <= 0 || this.height <= 0) {
+      throw new Error('the number is not positive');
+    }
+
+    this.color = color;
+    this.width = width;
```
volodymyr-soltys97, line 89: The properties color, width, height are already initialized due to public keyword in constructor, no need to reassign them.

volodymyr-soltys97 (changes requested):
Good job
But let's make your code perfect

mykhalenych (approved):

PR #1387, opened 2023-07-10
Solution

outofs:


src/figuresClasses.ts
```diff
@@ -1,19 +1,100 @@
-export interface Figure {
+type Shape = 'triangle' | 'circle' | 'rectangle';
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,100 @@
...
+      throw new Error('Wrong sides length');
+    }
+
+    this.a = a;
+    this.b = b;
+    this.color = color;
+  }
+
+  getArea(): number {
+    return this.a * this.b;
```
anstsot, line 94: try to add `* 100 / 100` that number always have 2 numbers after point

anstsot (changes requested):
Good start🍀

And678 (approved):
Gj, approve :heavy_check_mark:

PR #1388, opened 2023-07-10
figures-classes solution

stnayn:


src/figuresClasses.ts
```diff
@@ -1,19 +1,109 @@
 export interface Figure {
-
+  form: 'triangle' | 'circle' | 'rectangle',
```
TarasHoliuk, line 2: create a separate type for it and use everywhere you need (the same for color)
```suggestion
  shape: Shape,
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,109 @@
...
 export class Triangle {
+  constructor(
+    public color: Figure['color'],
+    public sideA: number,
+    public sideB: number,
+    public sideC: number,
+    public shape: Figure['form'],
+  ) {
+    if (sideA && sideB && sideC) {
+      let sides = [+sideA, +sideB, +sideC];
```
TarasHoliuk, line 15:
```suggestion
      let sides = [sideA, sideB, sideC];
```
it is already number (you have TypeScript which says it)

src/figuresClasses.ts
```diff
@@ -1,19 +1,109 @@
...
+      throw new Error('radius cannot be <= 0');
+    }
+
+    this.color = color;
+    this.radius = radius;
+    this.shape = 'circle';
+  }
+
+  public getArea(): number {
+    const area = Math.PI * (this.radius ?? 0) ** 2;
```
TarasHoliuk, line 71: radius is required here - it's always a number, so nullish coalescing operator is redundant

src/figuresClasses.ts
```diff
@@ -1,19 +1,109 @@
...
 
+    return +(Math.floor(area * 100) / 100).toFixed(2);
```
TarasHoliuk, line 73: you rounded it to hundreds dividing and multiplying it by 100, to fixed is redundant in this case

src/figuresClasses.ts
```diff
@@ -1,19 +1,109 @@
...
+  public getArea(): number {
+    const area = this.sideA * this.sideB;
 
-export function getInfo(figure) {
+    return +area.toFixed(2);
+  }
+}
 
+export function getInfo(
+  obj: {
```
TarasHoliuk, line 102: bad argument name - rename it

TarasHoliuk (changes requested):

OleziO (approved):

PR #1390, opened 2023-07-10
solution

ihorhladkov:


src/figuresClasses.ts
```diff
@@ -1,19 +1,84 @@
-export interface Figure {
+type Shapes = 'triangle'| 'circle' | 'rectangle';
```
anstsot, line 1: create `enum Shape` and use it everywhere

anstsot (changes requested):
Good start🍀
Let’s add one improvement

anstsot (approved):

PR #1391, opened 2023-07-10
Figures

AlexRaven93:


src/figuresClasses.ts
```diff
@@ -1,19 +1,96 @@
...
+
+  sideC: number;
+
+  color: Color;
+
+  shape: Shape = 'triangle';
+
+  constructor(color: Color, sideA: number, sideB: number, sideC: number) {
+    if (sideA <= 0 || sideB <= 0 || sideC <= 0) {
+      throw new Error('Some length is less than zero');
```
Anastasiia-Svintsova, line 23: The error message is not accurate. It should be 'Some length is less than or equal to zero'.

src/figuresClasses.ts
```diff
@@ -1,19 +1,96 @@
-export interface Figure {
+type Shape = 'rectangle' | 'circle' | 'triangle';
+type Color = 'red' | 'green' | 'blue';
```
Anastasiia-Svintsova, line 2: Great places for enums by the way

src/figuresClasses.ts
```diff
@@ -1,19 +1,96 @@
...
+  getArea = (): number => {
+    const result = this.sideA * this.sideB;
 
+    return result;
+  };
 }
 
-export function getInfo(figure) {
+export function getInfo(figure: Figure): string {
+  const result = `A ${figure.color} ${figure.shape} - ${figure.getArea()}`;
```
Anastasiia-Svintsova, line 93: Don't forget about objects destructuring

Anastasiia-Svintsova (approved):

polosanya (approved):
👍

PR #1392, opened 2023-07-10
add task solution

maksymPtashnyk:


src/figuresClasses.ts
```diff
@@ -1,19 +1,69 @@
 export interface Figure {
-
+  color: 'red' | 'green' | 'blue';
+  shape: string;
```
And678, line 3: Declare separate enums or unions for this.
```suggestion
  color: Color;
  shape: Shape;
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,69 @@
...
+  getArea: () => number;
 }
 
-export class Triangle {
+export class Triangle implements Figure {
+  shape = 'triangle';
```
And678, line 8:
```suggestion
  shape = Shape.triangle;
```

And678: After `Shape` enum is created

src/figuresClasses.ts
```diff
@@ -1,19 +1,69 @@
...
+    return Number((Math.sqrt(halfPerimeter * (halfPerimeter - this.a)
+    * (halfPerimeter - this.b) * (halfPerimeter - this.c))).toFixed(2));
+  }
 }
+export class Circle implements Figure {
+  shape = 'circle';
 
-export class Rectangle {
+  constructor(
+    public color: 'red' | 'green' | 'blue',
```
And678, line 36:
```suggestion
    public color: Color,
```

And678: Same for the rest

And678 (changes requested):
Fix your [test](https://github.com/mate-academy/ts_figures_classes/pull/1392/checks) for Circle area calculation

PR #1394, opened 2023-07-11
Implement figuresClasses

taras-bilas:


src/figuresClasses.ts
```diff
@@ -1,19 +1,91 @@
...
-export class Rectangle {
+export class Circle implements Figure {
+  shape: Shape = 'circle';
+
+  constructor(
+    public color: Color,
+    public radius: number,
+  ) {
+    this.color = color;
+    this.radius = radius;
```
And678, line 51: You are using [ts constructor assignment](https://kendaleiv.com/typescript-constructor-assignment-public-and-private-keywords/) here. No need to do `this.color = color` because you have `public` in constructor arguments
```suggestion
```

And678 (approved):
Gj, but check my comment

PR #1395, opened 2023-07-11
Task done

AntonBaranskyi:


src/figuresClasses.ts
```diff
@@ -1,19 +1,95 @@
 export interface Figure {
-
+  shape: 'triangle' | 'circle' | 'rectangle';
```
anstsot, line 2: let's create `enum Shape` and use it everywhere

src/figuresClasses.ts
```diff
@@ -1,19 +1,95 @@
...
+  color: 'red' | 'green' | 'blue';
+  getArea: () => number;
 }
 
 export class Triangle {
+  shape: Figure['shape'];
+
+  color: Figure['color'];
```
anstsot, line 10: better to create separate `Color` type

anstsot (changes requested):
Good start🍀

PR #1400, opened 2023-07-11
add solution

dimarogkov:


src/figuresClasses.ts
```diff
@@ -1,19 +1,91 @@
...
+  getArea(): number {
+    const semiPerimeter = (this.a + this.b + this.c) / 2;
+    const area = Math.sqrt(
+      semiPerimeter
+      * (semiPerimeter - this.a)
+      * (semiPerimeter - this.b)
+      * (semiPerimeter - this.c),
+    );
 
+    return Math.floor(area * 100) / 100;
```
Viktor-Kost, line 46: The rounding method used for the area is not correct. It's using Math.floor which rounds down, but the task requires rounding to the nearest hundredths. It should use the toFixed() method and convert the result back to number.
Apply for similar cases as well

Viktor-Kost (approved):
GJ!
Apply a minor fix from the comment for yourself

PR #1401, opened 2023-07-11
solution

ShamievDavid:


src/figuresClasses.ts
```diff
@@ -1,19 +1,89 @@
...
+    if (this.a <= 0 || this.b <= 0 || this.c <= 0) {
+      throw new Error('Error');
+    }
+    this.color = color;
+  }
+
+  getArea(): number {
+    const s = (this.a + this.b + this.c) / 2;
+
+    return Math.sqrt(s * (s - this.a) * (s - this.b) * (s - this.c));
```
Viktor-Kost, line 41: In the Triangle class, the getArea method does not round down the area to the nearest hundredths. This operation should be done in the getInfo function to follow the task description.

src/figuresClasses.ts
```diff
@@ -1,19 +1,89 @@
...
+  }
 
+  getArea(): number {
+    return this.width * this.height;
+  }
 }
 
-export function getInfo(figure) {
+export function getInfo(figure: Figure): string {
+  const figureArea = Number(figure.getArea().toFixed(2));
```
Viktor-Kost, line 86: In the getInfo function, you are converting the area to a number after calling toFixed. The toFixed method returns a string, so the Number conversion is unnecessary.

Viktor-Kost (changes requested):

mykhalenych (approved):

PR #1405, opened 2023-07-13
add solution

denis-kononenko:


src/figuresClasses.ts
```diff
@@ -1,19 +1,74 @@
-export interface Figure {
+type Shape = 'triangle' | 'circle' | 'rectangle';
 
+export interface Figure {
+  shape: Shape;
+  color: string;
```
vlad-demchuk, line 5: let's create some type of color as well to make it safer

vlad-demchuk (changes requested):

src/figuresClasses.ts
```diff
@@ -1,19 +1,75 @@
...
+  constructor(
+    public color: Color,
+    public radius: number,
+  ) {
+    if (radius <= 0) {
+      throw new Error('radius must be > 0');
+    }
+  }
 
+  getArea(): number {
```
anastasiiavorobiova, line 50: The area for circle is not being rounded down to hundredths. Use .toFixed(2) method.

src/figuresClasses.ts
```diff
@@ -1,19 +1,75 @@
...
 }
 
 export class Rectangle {
+  shape: Shape = 'rectangle';
 
-}
+  constructor(
+    public color: Color,
+    public width: number,
+    public heigth: number,
```
anastasiiavorobiova, line 61: The property name 'heigth' is misspelled. It should be 'height'.

src/figuresClasses.ts
```diff
@@ -1,19 +1,75 @@
...
+      || (b > a && b > c && b >= (a + c))
+      || (c > a && c > b && c >= (a + b))) {
+      throw new Error(`sides ${a}, ${b} and ${c} can't form a triangle`);
+    }
+  }
 
+  getArea(): number {
+    const s: number = (this.a + this.b + this.c) / 2;
+
+    return +Math.sqrt(s * (s - this.a) * (s - this.b) * (s - this.c))
```
anastasiiavorobiova, line 33:
```suggestion
    return Number(Math.sqrt(s * ...
```
Type coercion should be straightforward and easily readable

anastasiiavorobiova (changes requested):
Great solution! Added comments to fix

src/figuresClasses.ts
denis-kononenko: it is rounded, why not? not with .toFixed() method, but it's rounded

I had to use my decision because with .toFixed() didn't work one test (there were the result like 113.1 and test expected 113.09)

src/figuresClasses.ts
denis-kononenko: why do you not like the + method for converting string to number?)

anastasiiavorobiova: It's not straightforward coercion, the code should be easily readable by devs

anastasiiavorobiova (approved):
Awesome!

PR #1415, opened 2023-07-19
Solution

yurii-shkrobut-m:


src/figuresClasses.ts
```diff
@@ -1,19 +1,84 @@
...
+    if (sides[0] >= sides[1] + sides[2]) {
+      throw new Error('Error');
+    }
+
+    if (sides.some((el) => el <= 0)) {
+      throw new Error('Error');
+    }
+  }
 
+  getArea(): number {
```
qoa11a, line 38:
```suggestion
  getArea() {
```
As your class implements `Figure` there's no need to specify it.

qoa11a (changes requested):
A decent solution! 😊
Read the comments to improve it even more.

OleziO (approved):

PR #1416, opened 2023-07-20
Solution

92BS:


src/figuresClasses.ts
```diff
@@ -1,19 +1,94 @@
...
-}
+  color: Color;
+
+  constructor(
+    color: Color,
+    private a: number,
+    private b: number,
+    private c: number,
+  ) {
+    this.color = color;
```
polosanya, line 22: image

polosanya (changes requested):
Almost done! Fix all cases

src/figuresClasses.ts
92BS: From my side I think reasonable not to fix cases, because in documentation of Typescript we need to use modifiers private, public or protected:

https://www.typescriptlang.org/docs/handbook/2/classes.html#parameter-properties

Anastasiia-Svintsova (approved):

PR #1420, opened 2023-07-22
Figures Classes

1v43k:


src/figuresClasses.ts
```diff
@@ -1,19 +1,87 @@
...
+    public color: Colors,
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    const sides = [a, b, c];
+
+    sides.sort((x, y) => y - x);
+
+    if (sides[0] >= sides[1] + sides[2]) {
```
VitaliyBodnar, line 28: please use desctructured value of sides
```suggestion
    if (a >= b + c) {
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,87 @@
...
+      throw new Error('Attention, your triangle is not correct');
+    }
+
+    if (a <= 0 || b <= 0 || c <= 0) {
+      throw new Error('Attention, your triangle is not correct');
+    }
+  }
 
+  getArea(): number {
+    const square = (this.a + this.b + this.c) / 2;
```
VitaliyBodnar, line 38: make sense to destructure a, b, and c here
for example: 
`const { a, b, c } = this;`

src/figuresClasses.ts
```diff
@@ -1,19 +1,87 @@
...
+  ) {
+    if (width <= 0 || height <= 0) {
+      throw new Error('Attention, your rectangle is not correct');
+    }
+  }
+
+  getArea(): number {
+    const result = this.width * this.height;
+
+    return result;
```
VitaliyBodnar, line 81:
```suggestion
    return this.width * this.height;
```

VitaliyBodnar (changes requested):
Almost done!

etojeDenys (approved):
lgtm

PR #1427, opened 2023-07-25
Solution

JaneShavrukova:


src/figuresClasses.ts
```diff
@@ -1,19 +1,87 @@
...
+    if (width <= 0 || height <= 0) {
+      throw new Error('Sides cannot be 0 or less');
+    }
+  }
+
+  getArea(): number {
+    const { width, height } = this;
+    const area: number = width * height;
+
+    return Math.floor(100 * area) / 100;
```
TarasHoliuk, line 81: not critical, but you're doing the same 3 times - it's better to move it to a separate function
just FYI

TarasHoliuk (approved):
Well done - approved :heavy_check_mark: 
Added one comment just FYI (no need to re-request review)

PR #1429, opened 2023-07-27
add task solution

Luk2asz:


src/figuresClasses.ts
```diff
@@ -1,19 +1,74 @@
...
+  getArea(): number | Error
 }
 
-export class Triangle {
+export class Triangle implements Figure {
+  shape = 'triangle';
 
-}
+  constructor(
+    public color: 'red' | 'green' | 'blue',
```
DorotaLeniecDev, line 10: You could extract enum for Color
```js
enum Color {
  Red = 'red',
  Green = 'green',
  Blue = 'blue',
}
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,74 @@
...
+    public aSide: number,
+    public bSide: number,
+    public cSide: number,
```
DorotaLeniecDev, line 13: You can simply name it:
```suggestion
    public a: number,
    public b: number,
    public c: number,
```

And then use it directly in class body

src/figuresClasses.ts
```diff
@@ -1,19 +1,74 @@
...
+  ) {
+    const a: number = aSide;
+    const b: number = bSide;
+    const c: number = cSide;
```
DorotaLeniecDev, line 17:
```suggestion
//no need for this
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,74 @@
 export interface Figure {
-
+  color: 'red' | 'green' | 'blue';
+  getArea(): number | Error
 }
 
-export class Triangle {
+export class Triangle implements Figure {
+  shape = 'triangle';
```
DorotaLeniecDev, line 7: Create enum for Shape
```js
enum Shape {
  Triangle = 'triangle',
  Circle = 'circle',
  Rectangle = 'rectangle',
}
```

DorotaLeniecDev (changes requested):
Create enums to make code more reusable, check my comments to implement few corrections

src/figuresClasses.ts
```diff
@@ -1,19 +1,81 @@
 export interface Figure {
```
Radoslaw-Czerniawski, line 1: This interface should dictate shape of the blow classes, according to the requirements:
image
- add missing properties

Radoslaw-Czerniawski: Also the getArea is wrongly typed, it doesn't return number or Error, it always returns a number

src/figuresClasses.ts
```diff
@@ -1,19 +1,81 @@
...
+      throw new Error(`sides ${a}, ${b} and ${c} can't form a triangle`);
+    }
+  }
+
+  getArea(): number | Error {
+    const { a } = this;
+    const { b } = this;
+    const { c } = this;
 
+    const s: number = (a + b + c) / 2;
```
Radoslaw-Czerniawski, line 40: By default we should rely as much as possible on type inference. Here typescript by default knows that `s` will be a number, there is no point in typing it explicitly.

Radoslaw-Czerniawski: and `getArea()` if you type it on the interface and make classes implement it, there is no need for further typing

Radoslaw-Czerniawski (changes requested):
Needs some changes

DorotaLeniecDev (approved):
Neat 👌

PR #1441, opened 2023-08-09
Solution

HaniaNassalska:


src/figuresClasses.ts
```diff
@@ -1,19 +1,108 @@
...
+  radius: number;
+
+  constructor(
+    color: 'red' | 'green' | 'blue',
+    radius: number,
+  ) {
+    this.color = color;
+    this.radius = radius;
+
+    if (this.radius < 0) {
```
choeqq, line 65: The condition for throwing an error should check for radius <= 0, not just < 0, since a radius of 0 cannot form a circle.

src/figuresClasses.ts
```diff
@@ -1,19 +1,108 @@
...
+  constructor(
+    color: 'red' | 'green' | 'blue',
+    width: number,
+    height: number,
+  ) {
+    this.color = color;
+    this.width = width;
+    this.height = height;
+
+    if ((this.width < 0
```
choeqq, line 95: The condition for throwing an error should check for width <= 0 or height <= 0, not just < 0, as a width or height of 0 cannot form a rectangle.

src/figuresClasses.ts
```diff
@@ -1,19 +1,108 @@
...
+    sideA: number,
+    sideB: number,
+    sideC: number,
+  ) {
+    this.color = color;
+    this.sideA = sideA;
+    this.sideB = sideB;
+    this.sideC = sideC;
+
+    if ((this.sideA + this.sideB < this.sideC + 1
```
choeqq, line 29: why are you adding 1 to each traingle edge? 🤔

src/figuresClasses.ts
```diff
@@ -1,19 +1,108 @@
...
+    this.height = height;
+
+    if ((this.width < 0
+      || this.height < 0)) {
+      throw new Error('width or height can\'t form a rectangle');
+    }
+  }
+
+  getArea(): number {
+    return this.height * this.width;
```
choeqq, line 102: Area of the rectangle should be rounded down to hundredths as per task description.

choeqq (changes requested):

choeqq (approved):
Nice 🚀

PR #1443, opened 2023-08-10
solution

zvir91:


src/figuresClasses.ts
```diff
@@ -1,19 +1,74 @@
 export interface Figure {
-
+  shape: 'triangle' | 'circle' | 'rectangle',
+  color: 'red' | 'green' | 'blue',
```
maxim2310, line 3: would be nice to create separate type for this and use it in all places instead of 'red' | 'green' | 'blue'

maxim2310 (changes requested):
Apply my comments to all of classes

src/figuresClasses.ts
sTorba24: still not resolved 
image

move type color before `Figure` and use it for `color` field

sTorba24 (approved):

PR #1445, opened 2023-08-13
completed task

Horizon-git:


src/figuresClasses.ts
```diff
@@ -1,19 +1,87 @@
...
+export class Circle implements Figure {
+  public shape: Shape;
+
+  constructor(
+    public color: Color,
+    public radius: number,
+  ) {
+    this.shape = 'circle';
 
+    if (radius <= 0) {
```
volodymyr-soltys97, line 55: Instead of using the 'radius' variable directly which is undefined in this scope, you should use 'this.radius' that refers to the circle's radius value.

src/figuresClasses.ts
```diff
@@ -1,19 +1,87 @@
...
+  public shape: Shape;
+
+  constructor(
+    public color: Color,
+    public width: number,
+    public height: number,
+  ) {
+    this.shape = 'rectangle';
+
+    if (width <= 0 || height <= 0) {
```
volodymyr-soltys97, line 75: Instead of using the 'width' and 'height' variables directly which are undefined in this scope, you should use 'this.width' and 'this.height' that refer to the rectangle's width and height values.

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code

anastasiiavorobiova (approved):
Great work!

PR #1450, opened 2023-08-14
task solution

AnVBondar:


src/figuresClasses.ts
```diff
@@ -1,19 +1,84 @@
...
+    public color: Color,
+    public radius: number,
+  ) {
+    if (radius <= 0) {
+      throw new Error('Check entered sides!');
+    }
+  }
+
+  getArea(): number {
+    const { radius } = this;
```
volodymyr-soltys97, line 51: You should use `Math.round` instead of `Math.floor` to round off to the nearest hundredths as per the task description.

src/figuresClasses.ts
```diff
@@ -1,19 +1,84 @@
...
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    const longestSide = [a, b, c].sort((side1, side2) => side2 - side1);
+
+    if (longestSide[0] >= (longestSide[1] + longestSide[2])) {
+      throw new Error('Check entered sides!');
+    } else if (a <= 0 || b <= 0 || c <= 0) {
+      throw new Error('Check entered sides!');
```
volodymyr-soltys97, line 24: You can improve error messages to be more descriptive about what exactly went wrong. For example, 'All sides must be greater than 0.'

src/figuresClasses.ts
```diff
@@ -1,19 +1,84 @@
...
+    public height: number,
+  ) {
+    if (width <= 0 || height <= 0) {
+      throw new Error('Check entered sides!');
+    }
+  }
+
+  getArea(): number {
+    const { width, height } = this;
+    const area = Math.floor(100 * (width * height)) / 100;
```
volodymyr-soltys97, line 73: You should use `Math.round` instead of `Math.floor` to round off to the nearest hundredths as per the task description.

src/figuresClasses.ts
```diff
@@ -1,19 +1,84 @@
...
+export class Rectangle implements Figure {
+  shape: Shape = 'rectangle';
+
+  constructor(
+    public color: Color,
+    public width: number,
+    public height: number,
+  ) {
+    if (width <= 0 || height <= 0) {
+      throw new Error('Check entered sides!');
```
volodymyr-soltys97, line 67: You can improve error messages to be more descriptive about what exactly went wrong. For example, 'Width and height must be greater than 0.'

src/figuresClasses.ts
```diff
@@ -1,19 +1,84 @@
...
+  constructor(
+    public color: Color,
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    const longestSide = [a, b, c].sort((side1, side2) => side2 - side1);
+
+    if (longestSide[0] >= (longestSide[1] + longestSide[2])) {
+      throw new Error('Check entered sides!');
```
volodymyr-soltys97, line 22: You can improve error messages to be more descriptive about what exactly went wrong. For example, 'The longest side of a triangle must be less than the sum of the other two sides.'

src/figuresClasses.ts
```diff
@@ -1,19 +1,84 @@
...
-export class Circle {
+export class Circle implements Figure {
+  shape: Shape = 'circle';
 
+  constructor(
+    public color: Color,
+    public radius: number,
+  ) {
+    if (radius <= 0) {
+      throw new Error('Check entered sides!');
```
volodymyr-soltys97, line 46: You can improve error messages to be more descriptive about what exactly went wrong. For example, 'Radius must be greater than 0.'

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code

src/figuresClasses.ts
AnVBondar: All your comments noted and fixed except this one.
If I use Math.round in circle it failes tests. Therefore I changed this method in other figures except circle.

src/figuresClasses.ts
```diff
@@ -1,19 +1,87 @@
...
+export class Triangle implements Figure {
+  shape: Shape = 'triangle';
+
+  constructor(
+    public color: Color,
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    const longestSide = [a, b, c].sort((side1, side2) => side2 - side1);
```
TarasHoliuk, line 19:
```suggestion
    const longestSide = [a, b, c].sort((side1, side2) => side2 - side1)[0];
```
OR
```suggestion
    const sortedSides = [a, b, c].sort((side1, side2) => side2 - side1);
```
to make name matches with value, just FYI

TarasHoliuk (approved):

src/figuresClasses.ts
AnVBondar: Clear, thank you.

PR #1453, opened 2023-08-15
sol

cardinal1991:


src/figuresClasses.ts
```diff
@@ -1,19 +1,112 @@
...
+
+  a: number;
+
+  b: number;
+
+  c: number;
+
+  constructor(
+    color: string, a: number, b: number, c: number,
+  ) {
```
choeqq, line 20: you can access Figure types with this syntax: `Figure['typeName']` - you won't have to retype everything again

src/figuresClasses.ts
```diff
@@ -1,19 +1,112 @@
...
-export class Circle {
+export class Circle implements Figure {
+  shape: 'circle' = 'circle';
+
+  color: string;
 
+  radius: number;
+
+  constructor(
+    color: string, radius: number,
```
choeqq, line 56: you can access Figure types with this syntax: `Figure['typeName']` - you won't have to retype everything again

src/figuresClasses.ts
```diff
@@ -1,19 +1,112 @@
...
+
+  color: string;
+
+  width: number;
+
+  height: number;
+
+  constructor(
+    color: string, width: number, height: number,
+  ) {
```
choeqq, line 88: you can access Figure types with this syntax: `Figure['typeName']` - you won't have to retype everything again

src/figuresClasses.ts
```diff
@@ -1,19 +1,112 @@
 export interface Figure {
-
+  shape: string;
+  color: string;
+  getArea: ()=> number;
```
choeqq, line 4: is your prettier working correctly? 🤔

choeqq (changes requested):

src/figuresClasses.ts
cardinal1991: you mean this arrow function type? (if i guess correctly) i ask during lesson and it's a kind of function type that return a number type

choeqq (approved):
Good job ✅

PR #1459, opened 2023-08-16
add task solution

MateuszSlezak:


src/figuresClasses.ts
```diff
@@ -1,19 +1,92 @@
...
 
-export class Triangle {
+export class Triangle implements Figure {
+  shape: Shape;
+
+  color: Color;
+
+  sides: number[];
+
+  constructor(color: Color, a: number, b: number, c: number) {
```
choeqq, line 18: you can access Figure types with this syntax: `Figure['typeName']` - you won't have to retype everything again

src/figuresClasses.ts
```diff
@@ -1,19 +1,92 @@
...
 
-export class Circle {
+export class Circle implements Figure {
+  shape: Shape;
+
+  color: Color;
+
+  radius: number;
 
+  constructor(color: Color, radius: number) {
```
choeqq, line 49: you can access Figure types with this syntax: `Figure['typeName']` - you won't have to retype everything again

src/figuresClasses.ts
```diff
@@ -1,19 +1,92 @@
...
 export class Rectangle {
+  shape: Shape;
 
-}
+  color: Color;
+
+  sides: number[];
 
-export function getInfo(figure) {
+  constructor(color: Color, width: number, height: number) {
```
choeqq, line 73: you can access Figure types with this syntax: `Figure['typeName']` - you won't have to retype everything again

src/figuresClasses.ts
```diff
@@ -1,19 +1,92 @@
...
+    if (width <= 0 || height <= 0 || width === height) {
+      throw new Error('not a rectangle');
+    }
+
+    this.shape = 'rectangle';
+    this.color = color;
+    this.sides = [width, height];
```
choeqq, line 80: I think having each, width and height on class would be more useful, rectangle can only have either width or height - it won't ever be extended

choeqq (changes requested):

src/figuresClasses.ts
MateuszSlezak: I made corrections but I have no idea if that was the point... typescript is black magic to me ;///

choeqq: So imagine a scenario: we have 1 interface for example: `Figure`, and it's used in 10 different classes in our project, let's say you want to change color type and add 'purple', in this pattern you could do that without worrying about forgetting some classes, TS will remind you, you're forgetting something 😄 

Your solution is also valid, but it's better practice to infer as many types as possible, and since you're implementing `Figure` we're infering from `Figure`

@MateuszSlezak

PR #1460, opened 2023-08-16
solution

SviatoslavVasyletskyi:


src/figuresClasses.ts
```diff
@@ -1,19 +1,84 @@
...
 
-export class Circle {
+export class Circle implements Figure {
+  shape = 'circle';
 
+  color: string;
+
+  radius: number;
+
+  constructor(color: string, radius: number) {
```
choeqq, line 42: you can access Figure types with this syntax: `Figure['typeName']` - you won't have to retype everything again

choeqq (changes requested):

choeqq (approved):

PR #1462, opened 2023-08-16
solution

IvanOlecsandrovych:
Done

### DIFF
--- a/src/figuresClasses.ts
+++ b/src/figuresClasses.ts
@@ -1,19 +1,81 @@
-export interface Figure {
+type Shape = 'triangle' | 'circle' | 'rectangle';
+type Color = 'red' | 'green' | 'blue';
 
+export interface Figure {
+  shape: Shape
+  color: Color
+  getArea(): number,
 }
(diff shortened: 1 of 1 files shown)


src/figuresClasses.ts
```diff
@@ -1,19 +1,81 @@
...
 }
 
-export class Triangle {
+export class Triangle implements Figure {
+  constructor(
+    public color: Color,
+    public a: number,
+    public b: number,
+    public c: number,
+    public shape: Shape = 'triangle',
```
IvanFesenko, line 16: shape should not be availeble in constructor, becasu you can create Circle with Triangle logic

Fix all cases
```suggestion
export class Triangle implements Figure {
  shape = 'triangle'

  constructor(
    public color: Color,
    public a: number,
    public b: number,
    public c: number,
```

IvanFesenko (changes requested):
GJ. Check my comments

etojeDenys (approved):
lgtm

PR #1464, opened 2023-08-16
solution

patryk-babon:


src/figuresClasses.ts
```diff
@@ -1,19 +1,77 @@
...
+      s * (s - this.a) * (s - this.b) * (s - this.c),
+    ) * 100) / 100;
 
+    return area;
+  }
 }
 
-export class Rectangle {
+export class Circle implements Figure {
+  shape: 'circle' = 'circle';
```
choeqq, line 37: try to access those types through Figure: `Figure[typeName]`

patryk-babon:
@choeqq something like this?

choeqq (approved):
Yup, looks good ✅

PR #1468, opened 2023-08-16
Figures Classes

dimkamg21:


src/figuresClasses.ts
```diff
@@ -1,19 +1,84 @@
 export interface Figure {
-
+  shape: 'triangle' | 'circle' | 'rectangle',
```
maxim2310, line 2: better to use enum for this

src/figuresClasses.ts
```diff
@@ -1,19 +1,84 @@
...
+  color: 'red' | 'green' | 'blue',
+  getArea(): number,
 }
 
-export class Triangle {
+export class Triangle implements Figure {
+  color: Figure['color'];
 
-}
+  shape: Figure['shape'] = 'triangle';
```
maxim2310, line 10:
```suggestion
  shape = 'triangle';
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,84 @@
...
+
+  a: number;
+
```
maxim2310, line 13:
```suggestion
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,84 @@
...
+  b: number;
+
```
maxim2310, line 15:
```suggestion
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,84 @@
 export interface Figure {
-
+  shape: 'triangle' | 'circle' | 'rectangle',
+  color: 'red' | 'green' | 'blue',
```
maxim2310, line 3: you need to identifier a separate type for this , and then use it in all places belove

maxim2310 (changes requested):
Great job overall! Just couple things needs to be fixed before approval)

src/figuresClasses.ts
dimkamg21: I can`t remove empty line. Lint: Expected blank line between class members.

PR #1469, opened 2023-08-16
Add solution

WaveOfDandelions:


choeqq (commented):
Keep your PRs open so we can review them 😄 

@WaveOfDandelions

src/figuresClasses.ts
```diff
@@ -1,19 +1,78 @@
...
 
+export interface Figure {
+  shape: Shape;
+  color: Color;
+  getArea: () => number;
 }
 
-export class Triangle {
+export class Triangle implements Figure {
+  public shape = Shape.Triangle;
```
asakevych, line 11: where do you have Shape.Triangle? I think you don't have such an object 😉 
you can write it the same as you have in other figures - `shape: Figure['shape'] = 'triangle';`

asakevych (commented):
Good job! 🚀  just small change left 😉

src/figuresClasses.ts
WaveOfDandelions: Fixed

choeqq (approved):
Looks good to me ✅

PR #1479, opened 2023-08-23
figures classes

Orchibald:


src/figuresClasses.ts
```diff
@@ -1,19 +1,66 @@
 export interface Figure {
-
+  shape: string;
```
maxim2310, line 2: create enum for this and use it in code belove, because now it can be any string

src/figuresClasses.ts
```diff
@@ -1,19 +1,66 @@
...
+  constructor(
+    public color: string,
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    if (a <= 0 || b <= 0 || c <= 0 || a + b <= c || a + c <= b || b + c <= a) {
+      throw new Error('Invalid triangle sides');
+    }
+    this.color = color;
```
maxim2310, line 19: its redundant, you are already had the same in this injection   public color: string,

maxim2310 (changes requested):
Great job overall! Just couple tiny things needs to be fixed before approval)

VitaliyBondarenko1982 (approved):
Well done

PR #1481, opened 2023-08-24
add solution

prostonadobobb:


src/figuresClasses.ts
```diff
@@ -1,19 +1,83 @@
...
+    this.color = color;
+    this.radius = radius;
+
+    if (radius <= 0) {
+      throw new Error('Radius should be a positive number!');
+    }
+  }
+
+  getArea(): number {
+    return Math.floor(Math.PI * this.radius * this.radius * 100) / 100;
```
anastasiiavorobiova, line 55: The getArea() methods should round the area down to hundredths, but currently it's rounding down to a whole number first and then dividing by 100. This can be fixed by rounding to hundredths first and then using floor.

src/figuresClasses.ts
```diff
@@ -1,19 +1,83 @@
...
 }
 
 export class Circle {
+  shape: Figure['shape'] = 'circle';
 
+  constructor(
+    public color: Figure['color'],
+    public radius: number,
+  ) {
+    this.color = color;
```
anastasiiavorobiova, line 46: Assigning values to class properties in the constructor is redundant as these properties are already initialized by TypeScript.

src/figuresClasses.ts
```diff
@@ -1,19 +1,83 @@
...
+    this.height = height;
 
-export function getInfo(figure) {
+    if (width <= 0 || height <= 0) {
+      throw new Error('You cannot build a rectangle!');
+    }
+  }
+
+  getArea(): number {
+    return Math.floor(this.width * this.height * 100) / 100;
```
anastasiiavorobiova, line 77: The getArea() methods should round the area down to hundredths, but currently it's rounding down to a whole number first and then dividing by 100. This can be fixed by rounding to hundredths first and then using floor.

src/figuresClasses.ts
```diff
@@ -1,19 +1,83 @@
...
+      throw new Error('You cannot build a triangle');
+    }
+  }
+
+  getArea(): number {
+    const s = (this.a + this.b + this.c) / 2;
+    const area = Math.sqrt(s * (s - this.a)
+      * (s - this.b) * (s - this.c));
 
+    return Math.floor(area * 100) / 100;
```
anastasiiavorobiova, line 35: The getArea() methods should round the area down to hundredths, but currently it's rounding down to a whole number first and then dividing by 100. This can be fixed by rounding to hundredths first and then using floor.

src/figuresClasses.ts
```diff
@@ -1,19 +1,83 @@
...
 export class Rectangle {
+  shape: Figure['shape'] = 'rectangle';
 
-}
+  constructor(
+    public color: Figure['color'],
+    public width: number,
+    public height: number,
+  ) {
+    this.color = color;
```
anastasiiavorobiova, line 67: Assigning values to class properties in the constructor is redundant as these properties are already initialized by TypeScript.

src/figuresClasses.ts
```diff
@@ -1,19 +1,83 @@
...
 export class Triangle {
+  shape: Figure['shape'] = 'triangle';
+
+  constructor(
+    public color: Figure['color'],
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    this.color = color;
```
anastasiiavorobiova, line 16: Assigning values to class properties in the constructor is redundant as these properties are already initialized by TypeScript.

anastasiiavorobiova (changes requested):
Great work! Added comments to fix

src/figuresClasses.ts
prostonadobobb: What is the difference?

prostonadobobb: const area = Math.PI * this.radius * this.radius;

src/figuresClasses.ts
prostonadobobb: ок

PR #1482, opened 2023-08-27
Add task solution

NikitaOrliak:
Why can't I use 'shape: Figure.shape = ...' in classes? Compiler said that I should only use 'Figure['shape']'. And can I create a 'public color' BEFORE the constructor and inside the constructor take the colors' value? Don't understand how, if possible. Btw, I don't understand why I could use 'figure.getArea()' in 'getInfo' function without 'getArea(): number' method in the interface. The compiler argued that I couldn't, but the tests had been passing. Thank you!

### DIFF
--- a/src/figuresClasses.ts
+++ b/src/figuresClasses.ts
@@ -1,19 +1,79 @@
 export interface Figure {
-
+  shape: 'triangle' | 'circle' | 'rectangle';
+  color: 'red' | 'green' | 'blue';
+  getArea(): number;
 }
 
 export class Triangle {
+  shape: Figure['shape'] = 'triangle';
 
+  constructor(
+    public color: Figure['color'],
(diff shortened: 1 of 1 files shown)


etojeDenys (approved):
lgtm

etojeDenys:
> Why can't I use 'shape: Figure.shape = ...' in classes? Compiler said that I should only use 'Figure['shape']'. And can I create a 'public color' BEFORE the constructor and inside the constructor take the colors' value? Don't understand how, if possible. Btw, I don't understand why I could use 'figure.getArea()' in 'getInfo' function without 'getArea(): number' method in the interface. The compiler argued that I couldn't, but the tests had been passing. Thank you!

please, if you have some questions feel free to ask them in the chat

PR #1484, opened 2023-08-29
Solution

oksana-vas:


src/figuresClasses.ts
```diff
@@ -1,19 +1,93 @@
...
+
+  constructor(
+    public color: Figure['color'],
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    this.shape = Shape.Triangle;
+
+    if (this.a <= 0 || this.b <= 0 || this.c <= 0) {
```
loralevitska, line 31: Destructure values from `this`. Fix all cases

src/figuresClasses.ts
```diff
@@ -1,19 +1,93 @@
...
+      throw new Error('Please enter valid positive numbers');
+    }
+
+    if (this.a + this.b <= this.c
+      || this.a + this.c <= this.b
+      || this.c + this.b <= this.a) {
```
loralevitska, line 37: Long expressions move to variables with meaningful variable naming. Fix all cases

loralevitska (changes requested):
Almost done) Check my comments for improvement

sTorba24 (approved):

PR #1490, opened 2023-08-31
Implemented figure creating using ts classes and interfaces

davinnchii:


src/figuresClasses.ts
```diff
@@ -1,19 +1,91 @@
...
+    });
+
+    if (biggestSide >= sumOfOtherSides) {
+      throw new Error(ERROR_MESSAGE);
+    }
+  }
 
-export class Circle {
+  getArea(): number {
+    const p = (this.a + this.b + this.c) / 2;
```
Moroz-Dmytro, line 42: Try not to use names with only 1 letter

src/figuresClasses.ts
```diff
@@ -1,19 +1,91 @@
...
+    const triangleSquareFormula = Math.sqrt(p * (p - this.a)
+    * (p - this.b) * (p - this.c));
```
Moroz-Dmytro, line 44:
```suggestion
    const triangleSquareFormula = Math.sqrt(
      p
      * (p - this.a)
      * (p - this.b)
      * (p - this.c)
    );
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,91 @@
...
 
+    return Math.floor((triangleSquareFormula) * 100) / 100;
```
Moroz-Dmytro, line 46: Create function for this `Math.floor(...* 100) / 100`. You are doing the same in other places

src/figuresClasses.ts
```diff
@@ -1,19 +1,91 @@
...
+    public color: Color,
+    public radius: number,
+  ) {
+    if (radius <= 0) {
+      throw new Error(ERROR_MESSAGE);
+    }
+  }
+
+  getArea(): number {
+    const circleSquareFormula = Math.PI * this.radius * this.radius;
```
Moroz-Dmytro, line 63: `this.radius * this.radius` === `this.radius ** 2`

src/figuresClasses.ts
```diff
@@ -1,19 +1,91 @@
...
+
+  getArea(): number {
+    const rectangleSquareFormula = this.width * this.height;
+
+    return Math.floor((rectangleSquareFormula));
+  }
+}
 
+export function getInfo(figure: Figure): string {
+  return `A ${figure.color} ${figure.shape} - ${figure.getArea()}`;
```
Moroz-Dmytro, line 90: Use destructuring and create variable for this `${figure.getArea()}`. It will be easier to read

Moroz-Dmytro (changes requested):

src/figuresClasses.ts
davinnchii: In this case we use name p from formula of thriangle's square

src/figuresClasses.ts
```diff
@@ -1,19 +1,111 @@
...
+  triangle = 'triangle',
+  circle = 'circle',
+  rectangle = 'rectangle',
+}
+
+enum Color{
+  red = 'red',
+  green = 'green',
+  blue = 'blue',
+}
```
alexander-ignatow, line 11: Enum values should be capitalized (not `red`, but `Red`, and so on)

src/figuresClasses.ts
```diff
@@ -1,19 +1,111 @@
...
+
+const INVALID_SIZE_ERROR = 'Invalid sides for this figure';
```
alexander-ignatow, line 13: This error does not fit circle - as circle has no sides

src/figuresClasses.ts
```diff
@@ -1,19 +1,111 @@
...
 
+    const sides = [a, b, c];
+    const biggestSide = Math.max(...sides);
+    const sumOfOtherSides = sides.reduce((acc, value) => {
+      if (value !== biggestSide) {
+        return acc + value;
+      }
+
+      return acc;
+    });
```
alexander-ignatow, line 47: No need to fix it - but a generic solution is not always the best one
IMO the best way to implement it is the simplest
```suggestion
    const isOk = a + b < c && a + c < b && b + c < a; 
```

src/figuresClasses.ts
alexander-ignatow: Overal a better practice is being more friendly to your users and provide more descriptive error messages

alexander-ignatow (changes requested):
Good, but some more changes are needed

BudnikOleksii (approved):
Good work 👍

PR #1496, opened 2023-09-02
completed task ts_figures_classes

DenisChakhalian:


src/figuresClasses.ts
```diff
@@ -1,19 +1,99 @@
...
+}
+
+function toFixedArea(area:number):number {
+  const fixed = 10 ** 2;
+
+  return Math.floor(area * fixed) / fixed;
 }
 
 export class Triangle {
+  public readonly shape:Shape = 'triangle';
```
voievodik, line 26: public is the default, it works fine without it

src/figuresClasses.ts
```diff
@@ -1,19 +1,99 @@
...
+  constructor(
+    public color: Color,
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    if (this.a <= 0
+      || this.b <= 0
+      || this.c <= 0
+      || !isGoodTriangle(this.a, this.b, this.c)) {
```
voievodik, line 37: you can check it only by `if (a + b <= c || a + c <= b || c + b <= a)` without creating additional function (you use this function in code only once)

src/figuresClasses.ts
```diff
@@ -1,19 +1,99 @@
...
 
+  getArea():number {
+    const semiperimeter:number = (this.a + this.b + this.c) / 2;
+
+    const area = Math.sqrt(
+      semiperimeter
+      * (semiperimeter - this.a)
+      * (semiperimeter - this.b)
+      * (semiperimeter - this.c),
+    );
```
voievodik, line 52: you can use destructuring from `this` and in other places too

voievodik (approved):
Great work! I left you some comments from my side. If they make sense to you, you can fix it.

src/figuresClasses.ts
```diff
@@ -1,19 +1,99 @@
+type Color = 'red' | 'green' | 'blue';
+type Shape = 'triangle'| 'circle' | 'rectangle';
```
Moroz-Dmytro, line 2: Use `enum` for this

src/figuresClasses.ts
```diff
@@ -1,19 +1,99 @@
...
+  const copySides = [...sides];
+  const max = Math.max(...copySides);
+
+  copySides.splice(copySides.indexOf(max));
 
+  return (copySides[0] + copySides[1]) > max;
+}
+
+function toFixedArea(area:number):number {
+  const fixed = 10 ** 2;
```
Moroz-Dmytro, line 20: Do you really need this? Why not to use just 100?

src/figuresClasses.ts
```diff
@@ -1,19 +1,99 @@
...
+  getArea():number {
+    const area = this.width * this.height;
 
-export function getInfo(figure) {
+    return toFixedArea(area);
+  }
+}
 
+export function getInfo(figure:Figure):string {
+  return `A ${figure.color} ${figure.shape} - ${figure.getArea()}`;
```
Moroz-Dmytro, line 98: Destructuring

Moroz-Dmytro (changes requested):

src/figuresClasses.ts
alexander-ignatow: Please - don't ignore the comment - either fix the issue or comment on why you make other decision

alexander-ignatow (approved):
Good job 👍

PR #1497, opened 2023-09-02
added functionality for classes

voievodik:


kaavka (approved):
GJ!

src/figuresClasses.ts
```diff
@@ -1,19 +1,84 @@
+enum Shape {
+  Triangle = 'triangle',
+  Circle = 'circle',
+  Rectangle = 'rectangle',
+}
+
+type Color = 'red' | 'green' | 'blue';
```
alexander-ignatow, line 7: I would suggest use either only enums or only unions

src/figuresClasses.ts
```diff
@@ -1,19 +1,84 @@
...
+
+const ERROR_MESSAGE = 'Invalid values for this figure';
```
alexander-ignatow, line 9: The better idea is to have more detailed error message (imagine user trying to guess what went wrong)

alexander-ignatow (changes requested):
Good!

Still, you have some minor issues, and I would suggest fixing them

src/figuresClasses.ts
```diff
@@ -1,19 +1,94 @@
...
+    public c: number,
+  ) {
+    if (a + b <= c || a + c <= b || c + b <= a) {
+      throw new Error(ERROR_MESSAGE_TRIANGLE);
+    }
+  }
+
+  getArea(): number {
+    const { a, b, c } = this;
+    const p = (a + b + c) / 2;
```
Moroz-Dmytro, line 44: Try not to use such short names

src/figuresClasses.ts
```diff
@@ -1,19 +1,94 @@
...
+    const square = Math.sqrt(p * (p - a) * (p - b) * (p - c));
+
+    return Math.floor(square * 100) / 100;
```
Moroz-Dmytro, line 47: Let's create function for this `Math.floor(... * 100) / 100;`
You are using this code in other places

src/figuresClasses.ts
```diff
@@ -1,19 +1,94 @@
...
+  constructor(
+    public color: Color,
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    if (a + b <= c || a + c <= b || c + b <= a) {
+      throw new Error(ERROR_MESSAGE_TRIANGLE);
+    }
+  }
```
Moroz-Dmytro, line 40: What about checking if any of the numbers is a negative number?

src/figuresClasses.ts
```diff
@@ -1,19 +1,94 @@
...
 
+    return width * height;
+  }
 }
 
-export function getInfo(figure) {
+export function getInfo(figure: Figure): string {
+  const { color, shape } = figure;
 
+  return `A ${color} ${shape} - ${figure.getArea()}`;
```
Moroz-Dmytro, line 93: Let's create variable for `${figure.getArea()}`

Moroz-Dmytro (changes requested):
Good work 👍 
Small comments

src/figuresClasses.ts
```diff
@@ -1,19 +1,103 @@
...
+  shape: Shape = Shape.Triangle;
 
+  constructor(
+    public color: Color,
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    if (isTriangleValid(a, b, c)) {
+      throw new Error(ERROR_MESSAGE_TRIANGLE);
```
roman-mirzoian, line 43: If the problem is that the number is less than 0, then the error text will be irrelevant.

src/figuresClasses.ts
```diff
@@ -1,19 +1,103 @@
...
+const MAIN_ERROR_MESSAGE = 'Invalid values';
+const ERROR_MESSAGE_TRIANGLE = `${MAIN_ERROR_MESSAGE}:
+  the longest side of a triangle is >= than a sum of two others`;
+const ERROR_MESSAGE_CIRCLE = `${MAIN_ERROR_MESSAGE}:
+  radius of a circle is smaller than 0`;
+const ERROR_MESSAGE_RECTANGLE = `${MAIN_ERROR_MESSAGE}:
+  width or height of a rectangle is smaller than 0`;
+
+const handleRound = (square: number): number => Math.floor(square * 100) / 100;
+const isTriangleValid = (a: number, b: number, c: number): boolean => {
```
roman-mirzoian, line 22: You called the variable `isTriangleValid`, but in the middle you check only negative conditions, so this check should be called `isTriangleNotValid`.

src/figuresClasses.ts
```diff
@@ -1,19 +1,103 @@
...
+
+const MAIN_ERROR_MESSAGE = 'Invalid values';
+const ERROR_MESSAGE_TRIANGLE = `${MAIN_ERROR_MESSAGE}:
+  the longest side of a triangle is >= than a sum of two others`;
+const ERROR_MESSAGE_CIRCLE = `${MAIN_ERROR_MESSAGE}:
+  radius of a circle is smaller than 0`;
+const ERROR_MESSAGE_RECTANGLE = `${MAIN_ERROR_MESSAGE}:
+  width or height of a rectangle is smaller than 0`;
+
+const handleRound = (square: number): number => Math.floor(square * 100) / 100;
```
roman-mirzoian, line 21:
```suggestion
const handleNumberRound = (square: number): number => Math.floor(square * 100) / 100;
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,103 @@
...
+  ) {
+    if (radius <= 0) {
+      throw new Error(ERROR_MESSAGE_CIRCLE);
+    }
+  }
 
-export class Circle {
+  getArea(): number {
+    const { radius } = this;
+    const square = Math.PI * (radius ** 2);
```
roman-mirzoian, line 72: This destructuring is redundant because it won't do much good.
```suggestion
    const square = Math.PI * (this.radius ** 2);
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,103 @@
...
+  ) {
+    if (width <= 0 || height <= 0) {
+      throw new Error(ERROR_MESSAGE_RECTANGLE);
+    }
+  }
+
+  getArea(): number {
+    const { width, height } = this;
 
+    return width * height;
```
roman-mirzoian, line 94: This destructuring is redundant because it won't do much good.
```suggestion
    return this.width * this.height;
```

roman-mirzoian (changes requested):

src/figuresClasses.ts
voievodik: this is destructuring for future maintain code. If anyone wants to add more variables.

Moroz-Dmytro (approved):
LGTM 👍

PR #1500, opened 2023-09-02
add task solution

ydratsevska:


AndrMar1939 (approved):
Good job!

src/figuresClasses.ts
```diff
@@ -1,19 +1,87 @@
...
+  shape: Figure['shape'] = 'triangle';
+
+  constructor(
+    public color: Figure['color'],
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    if (a <= 0 || b <= 0 || c <= 0) {
+      throw new Error('Invalide parameters');
```
Moroz-Dmytro, line 20: image

The same in other places

src/figuresClasses.ts
```diff
@@ -1,19 +1,87 @@
...
+    public color: Figure['color'],
+    public radius: number,
+  ) {
+    if (radius <= 0) {
+      throw new Error('Invalide parameters');
+    }
+  }
+
+  getArea(): number {
+    const s = Math.PI * this.radius * this.radius;
```
Moroz-Dmytro, line 54: `s` is a bad name. Try not to use names with only 1 letter

src/figuresClasses.ts
```diff
@@ -1,19 +1,87 @@
...
+  ) {
+    if (height <= 0 || width <= 0) {
+      throw new Error('Invalide parameters');
+    }
+  }
+
+  getArea(): number {
+    const area = this.width * this.height;
 
+    return area;
```
Moroz-Dmytro, line 76:
```suggestion
    return this.width * this.height;
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,87 @@
...
 }
 
-export function getInfo(figure) {
+export function getInfo(figure: Figure): string {
+  const area = figure.getArea();
+  const formattedArea = area % 1 !== 0
+    ? area.toFixed(2)
+    : area;
 
+  return `A ${figure.color} ${figure.shape} - ${formattedArea}`;
```
Moroz-Dmytro, line 86: Destructuring

Moroz-Dmytro (changes requested):

src/figuresClasses.ts
```diff
@@ -1,19 +1,95 @@
...
 
+export interface Figure {
+  shape: Shape;
+  color: Color;
+  getArea(): number;
 }
 
-export class Triangle {
+export class Triangle implements Figure {
+  shape: Figure['shape'] = Shape.Triangle;
```
Moroz-Dmytro, line 20:
```suggestion
  shape: Shape = Shape.Triangle;
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,95 @@
...
+
+  constructor(
+    public color: Figure['color'],
```
Moroz-Dmytro, line 23:
```suggestion
    public color: Color,
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,95 @@
...
+    public radius: number,
+  ) {
+    if (radius <= 0) {
+      throw new Error('the radius is less or equal to zero');
+    }
+  }
 
+  getArea(): number {
+    const area = Math.PI * this.radius * this.radius;
```
Moroz-Dmytro, line 63:
```suggestion
    const area = Math.PI * this.radius ** 2;
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,95 @@
...
+
+    return Math.floor(area * 100) / 100;
```
Moroz-Dmytro, line 65: Create function for rounding value - `Math.floor(... * 100) / 100;`

src/figuresClasses.ts
```diff
@@ -1,19 +1,95 @@
...
+  }
 }
 
-export function getInfo(figure) {
+export function getInfo(figure: Figure): string {
+  const area = figure.getArea();
+  const { color, shape } = figure;
+  const formattedArea = area % 1 !== 0
+    ? area.toFixed(2)
+    : area;
```
Moroz-Dmytro, line 92: You don't need do this here
You can reuse your function in every `getArea` method in every class(check my comment above about creating function for rounding)

Moroz-Dmytro (changes requested):

src/figuresClasses.ts
```diff
@@ -1,19 +1,96 @@
...
+    public color: Color,
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    if (a <= 0 || b <= 0 || c <= 0) {
+      throw new Error('The side is less or equal to zero');
+    }
 
+    if (a + b <= c || a + c <= b || b + c <= a) {
```
alexander-ignatow, line 36: 👍 
This is probably the best possible solution

alexander-ignatow (approved):
Good 👍

PR #1501, opened 2023-09-02
implemented figure classes

yuliiaant:


src/figuresClasses.ts
```diff
@@ -1,19 +1,84 @@
+type Shapes = 'triangle' | 'circle' | 'rectangle';
```
AndriiKuz, line 1: in This case better use enum

src/figuresClasses.ts
```diff
@@ -1,19 +1,84 @@
...
+type Colors = 'red' | 'green' | 'blue';
+
 export interface Figure {
+  shape: Shapes;
+  color: Colors;
 
+  getArea(): number;
```
AndriiKuz, line 8: You wrote getArea() twice in 8 and 9 lines). Delete one.

src/figuresClasses.ts
```diff
@@ -1,19 +1,84 @@
...
+    ).toFixed(2);
+  }
 }
 
-export class Circle {
+export class Circle implements Figure {
+  shape: Shapes = 'circle';
+
+  constructor(
+    public color: Figure['color'],
```
AndriiKuz, line 47: I think it should be public color: Colors

AndriiKuz (changes requested):
GJ) I found some minor bugs in your code. Please fix them.

src/figuresClasses.ts
```diff
@@ -1,19 +1,92 @@
...
+  triangle = 'triangle',
+  circle = 'circle',
+  rectangle = 'rectangle'
+}
+
+enum Colors {
+  red = 'red',
+  green = 'green',
+  blue = 'blue'
+}
```
Moroz-Dmytro, line 11: image

src/figuresClasses.ts
```diff
@@ -1,19 +1,92 @@
+enum Shapes {
```
```suggestion
enum Shape {
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,92 @@
...
+  triangle = 'triangle',
+  circle = 'circle',
+  rectangle = 'rectangle'
+}
+
+enum Colors {
```
Moroz-Dmytro, line 7:
```suggestion
enum Color {
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,92 @@
...
 export interface Figure {
+  shape: Shapes;
+  color: Colors;
 
+  getArea: () => number;
 }
 
-export class Triangle {
+export class Triangle implements Figure {
+  shape: Shapes.triangle = Shapes.triangle;
```
Moroz-Dmytro, line 21:
```suggestion
  public shape: Shape = Shape.triangle;
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,92 @@
...
+
+  constructor(
+    public color: Colors,
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    const arayOfSides = [a, b, c]
+      .sort((number1: number, number2: number) => number2 - number1);
```
Moroz-Dmytro, line 30:
```suggestion
    const triangleSides = [a, b, c].sort((a: number, b: number) => b - a);
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,92 @@
...
+    const invalidTriangle = arayOfSides[0] >= (arayOfSides[1] + arayOfSides[2]);
```
Moroz-Dmytro, line 31:
```suggestion
    const isTriangleInvalid = triangleSides[0] >= (triangleSides[1] + triangleSides[2]);
```

Moroz-Dmytro: If variable will have boolean value -> name should start with `is`

src/figuresClasses.ts
```diff
@@ -1,19 +1,92 @@
...
+
+    if (invalidTriangle === true) {
```
Moroz-Dmytro, line 33:
```suggestion
    if (invalidTriangle) {
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,92 @@
...
+      throw new Error(`sides ${a}, ${b} and ${c} can't form a triangle`);
+    }
 
+    if (a <= 0 || b <= 0 || c <= 0) {
+      throw new Error('invalid side length');
+    }
```
Moroz-Dmytro, line 39: Better to do this check first, at the very beggining. Because if some side is < 0 -> we will not need to do sorting and other calculations

src/figuresClasses.ts
Moroz-Dmytro: Same comments to other classes

src/figuresClasses.ts
```diff
@@ -1,19 +1,92 @@
...
+  getArea(): number {
+    return this.width * this.height;
+  }
 }
 
-export function getInfo(figure) {
+export function getInfo(figure: Figure): string {
+  const square = figure.getArea();
 
+  return `A ${figure.color} ${figure.shape} - ${square}`;
```
Moroz-Dmytro, line 91:
```suggestion
  const { color, shape } = figure;
  const square = figure.getArea();

  return `A ${color} ${shape} - ${square}`;
```

Moroz-Dmytro (changes requested):

src/figuresClasses.ts
alexander-ignatow: Not sure why?

src/figuresClasses.ts
```diff
@@ -1,19 +1,96 @@
...
+  ) {
+    if (radius <= 0) {
+      throw new Error('invalid radius');
+    }
+  }
 
+  getArea(): number {
+    const circleArea = Math.PI * (this.radius * this.radius);
+
+    return Math.floor(circleArea * 100) / 100;
```
alexander-ignatow, line 68: Minor thing - but it is best to be consistent with rounding (for triangle you are using `toFixed`)
Both methods are valid, but the best is to use one of them, not both.

src/figuresClasses.ts
```diff
@@ -1,19 +1,96 @@
...
+
+    const triangleSides = [a, b, c]
+      .sort((sideA: number, sideB: number) => sideB - sideA);
+    const isTriangleInvalid = triangleSides[0] >= (
+      triangleSides[1] + triangleSides[2]
+    );
 
+    if (isTriangleInvalid) {
+      throw new Error(`sides ${a}, ${b} and ${c} can't form a triangle`);
+    }
```
alexander-ignatow, line 41: Minor, but you can skip creating extra variable and leave out the braces

```suggestion
    if (triangleSides[0] >= triangleSides[1] + triangleSides[2]) {
      throw new Error(`sides ${a}, ${b} and ${c} can't form a triangle`);
    }
```

alexander-ignatow (approved):
Looks good!

Just make sure to check my comments as an advice for the future!

PR #1502, opened 2023-09-03
add task solution

yar0man:


src/figuresClasses.ts
```diff
@@ -1,19 +1,77 @@
 export interface Figure {
-
+  shape: string;
+  color: string;
+  getArea(): number;
 }
 
 export class Triangle {
```
polinavafik, line 7: - Write an interface `Figure` and 3 classes **implementing** it

you should add `implements Figure` to all of your classes

src/figuresClasses.ts
```diff
@@ -1,19 +1,77 @@
...
+  public shape = 'triangle';
```
polinavafik, line 8: I would suggest creating a type for all the colors and shapes
like type Color = 'red' | 'blue' | 'green'
 You can also create an enum instead of type so that you would have autofill

polinavafik (changes requested):
Amazing! just a teensy little thing you need to add to make it perfect!

roman-mirzoian (approved):
Well done 👍

PR #1503, opened 2023-09-03
Solved task

AndriiKuz:


00Mass00 (approved):
Good job!

src/figuresClasses.ts
```diff
@@ -1,19 +1,98 @@
+enum Shape {
+  Triangle = 'triangle',
+  Circle = 'circle',
+  Rectangle = 'rectangle',
+}
+
+type Color = 'red' | 'green' | 'blue';
```
Moroz-Dmytro, line 7: Better to make `enum` here also

src/figuresClasses.ts
```diff
@@ -1,19 +1,98 @@
...
+
+const ERRORS = {
+  INVALID_SIDE: 'All sides should be have only positive numbers',
+  INVALID_RADIUS: 'Radius should be have only positive numbers',
+  NOT_TRIANGLE: 'It isn`t a triangle',
+};
```
Moroz-Dmytro, line 13: 👍 
```suggestion
const ERRORS = {
  INVALID_SIDE: 'All sides should have only positive numbers',
  INVALID_RADIUS: 'Radius should have only positive numbers',
  NOT_TRIANGLE: 'It isn`t a triangle',
};
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,98 @@
...
+    public color: Color,
+    public radius: number,
+  ) {
+    if (radius <= 0) {
+      throw new Error(ERRORS.INVALID_RADIUS);
+    }
+  }
+
+  getArea(): number {
+    const area = (Math.PI * this.radius ** 2);
```
Moroz-Dmytro, line 70:
```suggestion
    const area = Math.PI * this.radius ** 2;
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,98 @@
...
+  getArea(): number {
+    const area = this.width * this.height;
 
-export function getInfo(figure) {
+    return roundDownToHundredths(area);
+  }
+}
 
+export function getInfo(figure: Figure): string {
+  return `A ${figure.color} ${figure.shape} - ${figure.getArea()}`;
```
Moroz-Dmytro, line 97: We can use destructuring and create variable for `${figure.getArea()}`. It will be easier to read

Moroz-Dmytro (changes requested):
Good work 👍 
Small comments

src/figuresClasses.ts
AndriiKuz: But why enum is better than type? We don`t know what exactly color Class will get. In tests, we also pass a string value as the first argument // const circle = new Circle('red', 6) not  //const circle = new Circle(Colors.Red, 6);
For me 'type' is easier and shorter)

src/figuresClasses.ts
```diff
@@ -1,19 +1,105 @@
...
+      throw new Error(ERRORS.NOT_TRIANGLE);
+    }
+  }
+
+  getArea(): number {
+    const halfPerimetr = (this.a + this.b + this.c) / 2;
+    const area = Math.sqrt(halfPerimetr
+      * (halfPerimetr - this.a)
+      * (halfPerimetr - this.b)
+      * (halfPerimetr - this.c));
```
Moroz-Dmytro, line 55:
```suggestion
    const area = Math.sqrt(halfPerimetr
      * (halfPerimetr - this.a)
      * (halfPerimetr - this.b)
      * (halfPerimetr - this.c)
    );
```

Moroz-Dmytro (approved):
LGTM 👍

PR #1504, opened 2023-09-03
Solution🎀

polinavafik:


ydratsevska (approved):

src/figuresClasses.ts
```diff
@@ -1,19 +1,97 @@
...
+
+  getArea(): number {
+    const area = this.width * this.height;
+
+    return formatNumber(area);
+  }
+}
 
+export function getInfo(figure: Figure): string {
+  return `A ${figure.color} ${figure.shape} - ${figure.getArea()}`;
```
Moroz-Dmytro, line 96: We can use destructuring, and create variable for this value `figure.getArea()` for better readability

src/figuresClasses.ts
```diff
@@ -1,19 +1,97 @@
-export interface Figure {
+const ERROR_MESSAGE = 'Value cannot be <= 0!';
```
Moroz-Dmytro, line 1: 👍 
But try to make up a more descriptive name, which type of error message it is

Moroz-Dmytro (changes requested):
Good job 🚀 
Only 2 small comments and it's perfect ⭐

Moroz-Dmytro (approved):
LGTM 👍 🔥

PR #1505, opened 2023-09-03
Implemented classes of shapes and the information output function

00Mass00:


Orpheus29 (approved):
Hello, GJ, here we got to assign the exact values to the shape and color by setting special types (e.g., Shape and Color) in the beginning). So the lines 2 and 3 would look like:
shape: Shape;
color: Color;

Line 24 must look something like:

Same for the rest of figures. 

May destructure { a, b, c } = this; on line 39 to get rid of 'this' in the next lines.
Missed rounding the triangle area down to hundredths.

Good luck!

src/figuresClasses.ts
```diff
@@ -1,19 +1,86 @@
...
+      || c <= 0
+      || !this.isTriangle(a, b, c)
+    ) {
+      throw new Error('Invalid triangle sides');
+    }
+  }
+
+  shape = 'triangle';
+
+  private isTriangle(
```
alexander-ignatow, line 26: Probably minor, but 'isValidTriagle` or `isCorrectTriagle` sounds better

src/figuresClasses.ts
```diff
@@ -1,19 +1,86 @@
...
+      && a + c > b
+      && b + c > a
+    );
+  }
+
+  getArea(): number {
+    const s = (this.a + this.b + this.c) / 2;
 
+    return Math
+      .sqrt(s * (s - this.a) * (s - this.b) * (s - this.c));
```
alexander-ignatow, line 42: Probably you should limit decimal numbers after comma - just like with the circle

alexander-ignatow (changes requested):
Good, but some improvements are needed

PR #1506, opened 2023-09-03
solution

Valentyn-Bilyk:


src/figuresClasses.ts
```diff
@@ -1,19 +1,86 @@
...
+  constructor(
+    public color: Color,
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    const maxSideLength = Math.max(a, b, c);
+
+    if ((a || b || c) <= 0) {
+      throw new Error('your error message');
```
davinnchii, line 32: Move this to constant

src/figuresClasses.ts
```diff
@@ -1,19 +1,86 @@
...
+    public color: Color,
+    public radius: number,
+  ) {
+    if (radius <= 0) {
+      throw new Error('your error message');
+    }
+  }
 
+  public getArea(): number {
+    return Math.floor(Math.PI * (this.radius ** 2) * 100) / 100;
```
davinnchii, line 62: Move this to separate variable

src/figuresClasses.ts
```diff
@@ -1,19 +1,86 @@
...
+    public width: number,
+    public height: number,
+  ) {
+    if (width <= 0 || height <= 0) {
+      throw new Error('your error message');
+    }
+  }
+
+  public getArea(): number {
+    return this.height * this.width;
```
davinnchii, line 80: Move this to separate variable

davinnchii (changes requested):
GJ, but let's improve readability

src/figuresClasses.ts
```diff
@@ -1,19 +1,86 @@
...
+
+  constructor(
+    public color: Color,
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    const maxSideLength = Math.max(a, b, c);
+
+    if ((a || b || c) <= 0) {
```
roman-mirzoian, line 31: If the first value is valid, the check may not go any further, so all these values should be checked separately.

src/figuresClasses.ts
```diff
@@ -1,19 +1,86 @@
...
+      throw new Error('Invalid input values');
```
roman-mirzoian, line 32: More informative error text should be added.
Fix it throughout the code.

src/figuresClasses.ts
```diff
@@ -1,19 +1,86 @@
...
+    }
+
+    if (maxSideLength >= a + b + c - maxSideLength) {
+      throw new Error('Invalid input values');
+    }
+  }
 
+  public getArea(): number {
+    const { a, b, c } = this;
+    const p: number = (a + b + c) / 2;
```
roman-mirzoian, line 42: Do not abbreviate variable names to unreadable ones (it's regarding `p`).

src/figuresClasses.ts
roman-mirzoian: @davinnchii There is no point in putting it in a separate variable.

src/figuresClasses.ts
roman-mirzoian: There is no point in putting it in a separate variable.

roman-mirzoian (changes requested):

src/figuresClasses.ts
```diff
@@ -1,19 +1,91 @@
...
+
+    if (maxSideLength >= a + b + c - maxSideLength) {
+      throw new Error('Incorrect side length');
+    }
+  }
 
+  public getArea(): number {
+    const { a, b, c } = this;
+    const halfThePerimeter: number = (a + b + c) / 2;
+    const sqrt: number = Math.sqrt(
```
roman-mirzoian, line 43: Do not abbreviate variable names to unreadable ones (it's regarding `sqrt`).

src/figuresClasses.ts
```diff
@@ -1,19 +1,91 @@
...
+    }
+
+    if (maxSideLength >= a + b + c - maxSideLength) {
+      throw new Error('Incorrect side length');
+    }
+  }
 
+  public getArea(): number {
+    const { a, b, c } = this;
+    const halfThePerimeter: number = (a + b + c) / 2;
```
roman-mirzoian, line 42:
```suggestion
    const semiperimeter: number = (a + b + c) / 2;
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,91 @@
...
+  constructor(
+    public color: Color,
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    const maxSideLength = Math.max(a, b, c);
+
+    if (a <= 0 || b <= 0 || c <= 0) {
+      throw new Error('Incorrect side length');
```
roman-mirzoian, line 32: More informative error text should be added.
Fix it throughout the code.

roman-mirzoian (changes requested):
Do not abbreviate the names to unreadable ones: p, sqrt - this is impossible to read, and if there are many such variables in the code, it will be difficult to understand.
Also, give errors a specific text of the problem: `The length of the side cannot be negative or equal to zero`, etc.

roman-mirzoian (approved):
Well done !

PR #1507, opened 2023-09-03
add task solution

Orpheus29:


src/figuresClasses.ts
```diff
@@ -1,19 +1,73 @@
...
 
-}
+  constructor(
+    public color: Color,
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    if ((a <= 0 || b <= 0 || c <= 0)
+      || (a + b + c - Math.max(a, b, c)
```
AntonHerasymov, line 21: Better create a variable for this (Math.max(a, b, c))

src/figuresClasses.ts
```diff
@@ -1,19 +1,73 @@
-export interface Figure {
+type Shape = 'triangle' | 'circle' | 'rectangle';
```
AntonHerasymov, line 1: i think in this case better to use enum

AntonHerasymov (changes requested):
Good job!
Few comments

src/figuresClasses.ts
Orpheus29: Why, could you please explain?

src/figuresClasses.ts
Orpheus29: Ok, agree.

src/figuresClasses.ts
Moroz-Dmytro: Agree with @AntonHerasymov. @Orpheus29 In this case you will be able to use values of the enum(`Shape.triangle`...) when you set value for your shape inside class(and avoid using "magic values")
For example here:

image

You will be able to do like this -> `public shape: Shape = Shape.triangle;`

src/figuresClasses.ts
Moroz-Dmytro: Good point, please fix this. But even better would be to split this `if` to 2 different conditions with different Error messages(1 - when you got negative values, 2 - when sum of 2 sides are less than 3 side)

src/figuresClasses.ts
```diff
@@ -1,19 +1,73 @@
...
+      || (a + b + c - Math.max(a, b, c)
+       <= Math.max(a, b, c))) {
+      throw new Error('Invalid triangle sides');
+    }
+  }
 
-export class Circle {
+  getArea(): number {
+    const { a, b, c } = this;
+    const sP: number = (a + b + c) / 2;
```
Moroz-Dmytro, line 29: It's not a good idea to use such short names

src/figuresClasses.ts
```diff
@@ -1,19 +1,73 @@
...
 
+    return Math.trunc(Math.sqrt(sP * (sP - a) * (sP - b) * (sP - c))
+      * 100) / 100;
```
Moroz-Dmytro, line 32: Don't try to write your code in 1 line, it's difficult to read

src/figuresClasses.ts
```diff
@@ -1,19 +1,73 @@
...
-export class Rectangle {
+export class Circle implements Figure {
+  public shape: Shape = 'circle';
 
+  constructor(
+    public color: Color,
+    public radius: number,
+  ) {
+    if (radius <= 0) {
+      throw new Error('Invalid radius');
```
Moroz-Dmytro, line 44: Try to give more descriptive messages(Why it's invalid? -> Because it's < 0)

src/figuresClasses.ts
```diff
@@ -1,19 +1,73 @@
...
+    }
+  }
+
+  getArea(): number {
+    return Math.trunc(Math.PI * this.radius * this.radius * 100) / 100;
```
Moroz-Dmytro, line 49: Remember about DRY. You have the same logic in other classes. Create a function for this `Math.trunc(... * 100) / 100;`

src/figuresClasses.ts
```diff
@@ -1,19 +1,73 @@
...
+export class Rectangle implements Figure {
+  public shape: Shape = 'rectangle';
+
+  constructor(
+    public color: Color,
+    public width: number,
+    public height: number,
+  ) {
+    if (width <= 0 || height <= 0) {
+      throw new Error('Invalid rectangle sides');
```
Moroz-Dmytro, line 62: Same comment about error message

src/figuresClasses.ts
```diff
@@ -1,19 +1,73 @@
...
+    }
+  }
+
+  getArea(): number {
+    return Math.trunc(this.width * this.height * 100) / 100;
+  }
+}
 
+export function getInfo(figure: Figure): string {
+  return `A ${figure.color} ${figure.shape} - ${figure.getArea()}`;
```
Moroz-Dmytro, line 72: Let's use destructuring here, and also create a variable for area. It will be easier for reading

src/figuresClasses.ts
```diff
@@ -1,19 +1,107 @@
+// type Shape = 'triangle' | 'circle' | 'rectangle';
```
Moroz-Dmytro, line 1: Don't leave comments in your code

src/figuresClasses.ts
```diff
@@ -1,19 +1,107 @@
...
+  color: Color;
+  getArea(): number;
+}
 
+function normalizeArea(areaValue: number): number {
+  return Math.trunc(areaValue * 100) / 100;
 }
 
-export class Triangle {
+const ERROR_MESSAGE = 'Value cannot be <= 0!';
```
Moroz-Dmytro, line 27: But would be good also to give more descriptive name for this variable(what type of error message it is?)

Moroz-Dmytro (approved):
LGTM 👍

Orpheus29:
Sorry about sending a premature request before.
I think I I've covered everything now.
I was stuck on was why linter underscores color names in tests when I use enum.
image

Nevertheless, the tests have run smoothly.

PR #1514, opened 2023-09-04
added figures

Constantin-Haranskiy:


src/figuresClasses.ts
```diff
@@ -1,19 +1,86 @@
-export interface Figure {
+type Shape = 'triangle' | 'circle' | 'rectangle';
+type Color = 'red' | 'green' | 'blue';
 
-}
+function getArea(): number {
+  const s = (this.a + this.b + this.c) / 2;
```
DarkMistyRoom, line 5: you need s only for a `triangle`. So move this logic to appropriate `case`

DarkMistyRoom (approved):
Well done!

PR #1515, opened 2023-09-04
solution

Vitalii-Fediaiev:


src/figuresClasses.ts
```diff
@@ -1,19 +1,87 @@
...
 
+export interface Figure {
+  shape: Shape;
+  color: Color;
+  getArea(): number;
 }
 
-export class Triangle {
+export class Triangle implements Figure {
+  shape: Figure['shape'] = 'triangle';
```
IvanFesenko, line 11: redundant type here
```suggestion
  shape = 'triangle';
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,87 @@
...
+
+  constructor(
+    public color: Figure['color'],
```
IvanFesenko, line 14:
```suggestion
    public color: Color,
```

IvanFesenko (approved):
Pay attention on typesation

PR #1517, opened 2023-09-04
add realization of Figure, Triangle, Circle and Rectangle

ivannaromanovych:


src/figuresClasses.ts
```diff
@@ -1,19 +1,97 @@
...
+    public color: Color,
+    a: number,
+    b: number,
+    c: number,
+  ) {
+    if (a >= b + c || a <= Math.abs(b - c) || a <= 0 || b <= 0 || c <= 0) {
+      throw new Error('Impossible triangle sides');
+    }
+
+    this.a = a;
```
dpidlutska, line 34: you don't need to do assign like 'this.a = a', it works by default.

src/figuresClasses.ts
```diff
@@ -1,19 +1,97 @@
...
+    this.b = b;
+    this.c = c;
+  }
+
+  getArea(): number {
+    const p: number = (this.a + this.b + this.c) / 2;
+    const area = Math.floor(
+      Math.sqrt(p * (p - this.a) * (p - this.b) * (p - this.c))
```
dpidlutska, line 42: I would do destructuring of 'this', it will make your code shorter

dpidlutska (approved):
Great job! 🔥
Check some comments to make you code even better.

src/figuresClasses.ts
```diff
@@ -1,19 +1,83 @@
-export interface Figure {
+enum Shape {
+  Triangle = 'triangle',
+  Circle = 'circle',
+  Rectangle = 'rectangle',
+}
+
+type Color = 'red' | 'green' | 'blue';
```
roman-mirzoian, line 7: It is worth sticking to one style: either all variables through enam, or as a set of types.

src/figuresClasses.ts
```diff
@@ -1,19 +1,83 @@
...
+  shape = Shape.Triangle;
+
+  constructor(
+    public color: Color,
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    if (a >= b + c || a <= Math.abs(b - c) || a <= 0 || b <= 0 || c <= 0) {
+      throw new Error('Invalid triangle sides');
```
roman-mirzoian, line 25: It is worth breaking it into two separate conditions and adding a more informative error text.

src/figuresClasses.ts
```diff
@@ -1,19 +1,83 @@
...
+    }
+  }
 
+  getArea(): number {
+    const p: number = (this.a + this.b + this.c) / 2;
```
roman-mirzoian, line 30: Don't abbreviate variable names to something unreadable (it's regarding `p`).

src/figuresClasses.ts
roman-mirzoian: Good idea!

src/figuresClasses.ts
```diff
@@ -1,19 +1,83 @@
...
-export class Circle {
+export class Circle implements Figure {
+  shape = Shape.Circle;
+
+  constructor(
+    public color: Color,
+    public radius: number,
+  ) {
+    if (radius <= 0) {
+      throw new Error('Invalid circle radius');
```
roman-mirzoian, line 48: Add more informative error text.

src/figuresClasses.ts
```diff
@@ -1,19 +1,83 @@
...
+    }
+  }
+
+  getArea(): number {
+    const area = Math.floor(100 * Math.PI * (this.radius ** 2)) / 100;
 
+    return area;
```
roman-mirzoian, line 55:
```suggestion
    return Math.floor(100 * Math.PI * (this.radius ** 2)) / 100;
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,83 @@
...
+export class Rectangle implements Figure {
+  shape = Shape.Rectangle;
+
+  constructor(
+    public color: Color,
+    public width: number,
+    public height: number,
+  ) {
+    if (width <= 0 || height <= 0) {
+      throw new Error('Invalid rectangle sides');
```
roman-mirzoian, line 68: Add more informative error text.

src/figuresClasses.ts
```diff
@@ -1,19 +1,83 @@
...
+    }
+  }
+
+  getArea(): number {
+    const area: number = this.width * this.height;
 
+    return area;
```
roman-mirzoian, line 75:
```suggestion
    return this.width * this.height;
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,83 @@
...
+  }
 }
 
-export function getInfo(figure) {
+export function getInfo(figure: Figure): string {
+  const info = `A ${figure.color} ${figure.shape} - ${figure.getArea()}`;
 
+  return info;
```
roman-mirzoian, line 82:
```suggestion
  const { color, shape } = figure;
  const figureArea = figure.getArea();
  return  `A ${color} ${shape} - ${figureArea }`;
```

roman-mirzoian (changes requested):

src/figuresClasses.ts
```diff
@@ -1,19 +1,86 @@
-export interface Figure {
+type Shape = 'triangle' | 'circle' | 'rectangle';
+
+type Color = 'red' | 'green' | 'blue';
```
Moroz-Dmytro, line 3: But better would be to use `enum` here 🙃
In this case, you will be able to do ` shape: Shape = Shape.triangle;` instead of ` shape: Shape = 'triangle';`
You will avoid using "magic value"

src/figuresClasses.ts
```diff
@@ -1,19 +1,86 @@
...
+    public color: Color,
+    public radius: number,
+  ) {
+    if (radius <= 0) {
+      throw new Error('Radius of the circle is less or equal zero');
+    }
+  }
+
+  getArea(): number {
+    return Math.floor(100 * Math.PI * (this.radius ** 2)) / 100;
```
Moroz-Dmytro, line 58: Create variable for this `Math.PI * (this.radius ** 2)`
Also, remember about DRY. You are using next logic at least twice in your code `Math.floor(100 * ...) / 100;`
Create function for this

src/figuresClasses.ts
```diff
@@ -1,19 +1,86 @@
...
+  shape: Shape = 'rectangle';
+
+  constructor(
+    public color: Color,
+    public width: number,
+    public height: number,
+  ) {
+    if (width <= 0 || height <= 0) {
+      throw new Error(
+        'One or both sides of the rectangle are les or equal zero',
```
Moroz-Dmytro, line 72:
```suggestion
        'One or both sides of the rectangle are less or equal zero',
```

src/figuresClasses.ts
Moroz-Dmytro: Please, fix this

Moroz-Dmytro (changes requested):

src/figuresClasses.ts
```diff
@@ -1,19 +1,96 @@
...
+  Rectangle = 'rectangle',
+}
+
+enum Color {
+  Red = 'red',
+  Green = 'green',
+  Blue = 'blue',
+}
+
+function roundToHundreths(value: number): number {
```
```suggestion
function roundToHundredths(value: number): number {
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,96 @@
...
+  }
 
+  getArea(): number {
+    const semiPerimeter: number = (this.a + this.b + this.c) / 2;
+    const area = roundToHundreths(Math.sqrt(semiPerimeter
+      * (semiPerimeter - this.a)
+      * (semiPerimeter - this.b)
+      * (semiPerimeter - this.c)));
+
+    return area;
```
Moroz-Dmytro, line 50:
```suggestion
    const area = Math.sqrt(semiPerimeter
      * (semiPerimeter - this.a)
      * (semiPerimeter - this.b)
      * (semiPerimeter - this.c)
    );

    return roundToHundreths(area);
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,96 @@
...
+    return this.width * this.height;
+  }
 }
 
-export function getInfo(figure) {
+export function getInfo(figure: Figure): string {
+  const figureArea = figure.getArea();
+  const info = `A ${figure.color} ${figure.shape} - ${figureArea}`;
 
+  return info;
```
Moroz-Dmytro, line 95:
```suggestion
  const { color, shape } = figure;
  const figureArea = figure.getArea();
  
  return `A ${color} ${shape} - ${figureArea}`;
```

Moroz-Dmytro (approved):
Good work 👍 
But check small advice below

PR #1520, opened 2023-09-04
add figures classes

ihorid:


src/figuresClasses.ts
```diff
@@ -1,19 +1,84 @@
...
+      throw new Error('wrong length specified');
+    }
+
+    if ((a + b) <= c || (b + c) <= a || (c + a) <= b) {
+      throw new Error('wrong length specified');
+    }
+  }
 
+  getArea(): number {
+    const s = (this.a + this.b + this.c) / 2;
```
NazarVotkanych, line 34: Here you can also destructure this to make the code more concise.
 const {a, b, c} = this

src/figuresClasses.ts
```diff
@@ -1,19 +1,84 @@
+
+enum Shape {
+  Triangle = 'triangle',
+  Rectangle = 'rectangle',
+  Circle = 'circle',
+}
+
 export interface Figure {
+  shape: Shape;
+  color: 'red' | 'green' | 'blue',
```
NazarVotkanych, line 10: You can also create enum for Color

src/figuresClasses.ts
```diff
@@ -1,19 +1,84 @@
+
```
```suggestion
```

src/figuresClasses.ts
Moroz-Dmytro: Good point, please fix this

src/figuresClasses.ts
```diff
@@ -1,19 +1,84 @@
...
+  public shape = Shape.Triangle;
+
+  constructor(
+    public color: 'red' | 'green' | 'blue',
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    if (a <= 0 || b <= 0 || c <= 0) {
+      throw new Error('wrong length specified');
```
Moroz-Dmytro, line 25: image

Same comment to all other error messages

src/figuresClasses.ts
```diff
@@ -1,19 +1,84 @@
...
+  ) {
+    if (radius <= 0) {
+      throw new Error('wrong length specified');
+    }
+  }
+
+  getArea(): number {
+    const area = Math.PI * this.radius ** 2;
 
+    return Math.trunc(area * 100) / 100;
```
Moroz-Dmytro, line 56: Create function for this `Math.trunc(... * 100) / 100;` and reuse in all places where you need to round your values

PR #1522, opened 2023-09-04
added TS_Figures_Classes solution

NazarVotkanych:


src/figuresClasses.ts
```diff
@@ -1,19 +1,89 @@
...
+      throw new Error('your error message');
+    }
+
+    if (a + b <= c || a + c <= b || b + c <= a) {
+      throw new Error('your error message');
+    }
+  }
+
+  getArea(): number {
+    const { a, b, c } = this;
```
voronine, line 39: cool way

voronine (approved):
All good ;)

src/figuresClasses.ts
```diff
@@ -1,19 +1,89 @@
...
+    public b: number,
+    public c: number,
+  ) {
+    if (a <= 0 || b <= 0 || c <= 0) {
+      throw new Error('your error message');
+    }
+
+    if (a + b <= c || a + c <= b || b + c <= a) {
+      throw new Error('your error message');
+    }
```
roman-mirzoian, line 35: A more informative error description should be added.
Fix this in all code.

src/figuresClasses.ts
```diff
@@ -1,19 +1,89 @@
...
+  }
+
+  getArea(): number {
+    const { a, b, c } = this;
+    const p: number = (a + b + c) / 2;
+    const sTriangle: number = Math.sqrt(p * (p - a) * (p - b) * (p - c));
 
+    return Number(sTriangle.toFixed(2));
```
roman-mirzoian, line 43: `a`, `b`, `c` can only be numbers, why do you call the `Number` constructor for them?
Fix this in all code.

src/figuresClasses.ts
```diff
@@ -1,19 +1,89 @@
...
+    public radius: number,
+  ) {
+    if (radius <= 0) {
+      throw new Error('your error message');
+    }
+  }
+
+  getArea(): number {
+    const pi = Math.PI;
+    const area = pi * (this.radius ** 2);
```
roman-mirzoian, line 61: The variable for PI is redundant here:
```suggestion
    const area = Math.PI * (this.radius ** 2);
```

roman-mirzoian (changes requested):

PR #1523, opened 2023-09-04
created figure classes and getInfo function

edgergard:


src/figuresClasses.ts
```diff
@@ -1,19 +1,88 @@
...
+  shape = Shape.Triangle;
+
+  constructor(
+    public color: Color,
+    public side1: number,
+    public side2: number,
+    public side3: number,
+  ) {
+    if (side1 <= 0 || side2 <= 0 || side3 <= 0) {
+      throw new Error('Sides value cant be zero or less');
```
KatOlista, line 35: i think you need variable for error message

src/figuresClasses.ts
```diff
@@ -1,19 +1,88 @@
...
+    public color: Color,
+    public radius: number,
+  ) {
+    if (radius <= 0) {
+      throw new Error('Circle radius value cant be zero or less');
+    }
+  }
 
+  getArea(): number {
+    return Math.floor(Math.PI * (this.radius ** 2) * 100) / 100;
```
KatOlista, line 64: will be better take out Math.floor in separate function. You repeat yourself

KatOlista (approved):
Good job! Read my comments, you can improve your code

src/figuresClasses.ts
```diff
@@ -1,19 +1,97 @@
...
+  const s: number = (a + b + c) / 2;
+
+  return roundNumber(Math.sqrt(s * (s - a) * (s - b) * (s - c)));
 }
 
-export class Triangle {
+const error = {
+  valueError: 'Value cant be zero or less',
+  triangleError: 'Side value cant be more than sum of two other sides',
+};
```
roman-mirzoian, line 14:
```suggestion
const ERROR_MESSAGES = {
  valueError: 'Side value cant be zero or less.',
  triangleError: 'Side value cant be more than sum of two other sides.',
};
```

roman-mirzoian (approved):
Well done, but pay attention to the comment.

PR #1525, opened 2023-09-04
created figures classes and methods

OlhaMomot:


src/figuresClasses.ts
```diff
@@ -1,19 +1,78 @@
...
 
+  getArea(): number;
 }
 
-export class Triangle {
+export class Triangle implements Figure {
+  shape: 'triangle' = 'triangle';
 
+  constructor(
+    public color: 'red' | 'green' | 'blue',
```
rostyslav48, line 12: You are using this type more than once, so it is better to create a separate type for it
example:
`type FigureColor = 'red' | 'green' | 'blue';`

rostyslav48 (approved):
Good job, but please replace repeated types with 1 type)

src/figuresClasses.ts
```diff
@@ -1,19 +1,78 @@
 export interface Figure {
+  shape: 'triangle' | 'circle' | 'rectangle';
+  color: 'red' | 'green' | 'blue';
```
Moroz-Dmytro, line 4: Create `enum` for `shape` and `color`

src/figuresClasses.ts
```diff
@@ -1,19 +1,78 @@
...
+
+    const longestSide = Math.max(a, b, c);
+
+    if ((a + b + c) - longestSide <= longestSide) {
+      throw new Error('These sides can not be used to form a triangle');
+    }
+  }
+
+  getArea(): number {
+    const s = 0.5 * (this.a + this.b + this.c);
```
Moroz-Dmytro, line 29: Don't use such short names with only 1 letter

Moroz-Dmytro: Also, `0.5 * (this.a + this.b + this.c)` === `(this.a + this.b + this.c) / 2`

src/figuresClasses.ts
```diff
@@ -1,19 +1,78 @@
...
+  ) {
+    if (this.radius <= 0) {
+      throw new Error('Number can not be less than 0');
+    }
+  }
+
+  getArea(): number {
+    const area = Math.PI * (this.radius ** 2);
+
+    return Number(Math.floor(area * 100) / 100);
```
Moroz-Dmytro, line 52: Create function for this `Math.floor(... * 100) / 100)`
Also, you don't need to use `Number` here

src/figuresClasses.ts
```diff
@@ -1,19 +1,78 @@
...
+      throw new Error('These sides can not be used to form a triangle');
+    }
+  }
+
+  getArea(): number {
+    const s = 0.5 * (this.a + this.b + this.c);
+
+    return +(Math.sqrt(
+      s * (s - this.a) * (s - this.b) * (s - this.c),
+    )).toFixed(2);
```
Moroz-Dmytro, line 33: You can reuse here function that you will create, instead of using `toFixed`(check my other comments)

src/figuresClasses.ts
```diff
@@ -1,19 +1,78 @@
...
+  getArea(): number {
+    return this.width * this.height;
+  }
 }
 
-export function getInfo(figure) {
+export function getInfo(figure: Figure): string {
+  const area = figure.getArea();
 
+  return `A ${figure.color} ${figure.shape} - ${area}`;
```
Moroz-Dmytro, line 77: Destructuring

Moroz-Dmytro (changes requested):
Good progress 👍 
CHeck some comments below

src/figuresClasses.ts
OlhaMomot: Can't do that, as I use it to call 'getArea' method on a previous line. Or I just didn't get it)

src/figuresClasses.ts
Moroz-Dmytro: My comment was only to variable `s` 🙃

src/figuresClasses.ts
Moroz-Dmytro: You can do like this
`const { color, shape } = figure;`

src/figuresClasses.ts
```diff
@@ -1,19 +1,101 @@
...
+  triangle = 'triangle',
+  circle = 'circle',
+  rectangle = 'rectangle'
+}
+
+export enum Color {
+  red = 'red',
+  green = 'green',
+  blue = 'blue'
+}
```
Moroz-Dmytro, line 11:
```suggestion
export enum Shape {
  Triangle = 'triangle',
  Circle = 'circle',
  Rectangle = 'rectangle'
}

export enum Color {
  Red = 'red',
  Green = 'green',
  Blue = 'blue'
}
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,101 @@
...
+
 export interface Figure {
+  shape: Shape;
+  color: Color;
 
+  getArea(): number;
 }
 
-export class Triangle {
+function getValidArea(area: number): number {
```
Moroz-Dmytro, line 20: Better to name it like round...

Moroz-Dmytro (approved):

PR #1526, opened 2023-09-04
Develop

KatOlista:


src/figuresClasses.ts
```diff
@@ -1,19 +1,97 @@
-export interface Figure {
+const errorMessage = {
+  wrongLength: 'length is <= 0',
+  tooLongSide: 'the longest side of a triangle is >= than a sum of two others',
+};
+
+const standardArea = (area: number): number => {
+  return Math.floor(area * 100) / 100;
+};
```
desevoker, line 8:
```suggestion
const roundToTwoDecimals = (value: number): number => {
  return Math.floor(value * 100) / 100;
};
```

desevoker (approved):
Goob job! 🚀

src/figuresClasses.ts
```diff
@@ -1,19 +1,97 @@
-export interface Figure {
+const errorMessage = {
```
```suggestion
const ERROR_MESSAGE = {
```

src/figuresClasses.ts
Moroz-Dmytro: Good point

src/figuresClasses.ts
```diff
@@ -1,19 +1,97 @@
...
+    public b: number,
+    public c: number,
+  ) {
+    if (a <= 0 || b <= 0 || c <= 0) {
+      throw new Error(errorMessage.wrongLength);
+    }
 
+    if (a >= b + c
+      || b >= a + c
+      || c >= b + a) {
```
Moroz-Dmytro, line 44:
```suggestion
    if (a >= b + c || b >= a + c || c >= b + a) {
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,97 @@
...
+      throw new Error(errorMessage.tooLongSide);
+    }
+  }
+
+  getArea(): number {
+    const semiperimeter = (this.a + this.b + this.c) / 2;
+    const area = Math.sqrt(semiperimeter
+        * (semiperimeter - this.a)
+        * (semiperimeter - this.b)
+        * (semiperimeter - this.c));
```
Moroz-Dmytro, line 54:
```suggestion
    const area = Math.sqrt(semiperimeter
      * (semiperimeter - this.a)
      * (semiperimeter - this.b)
      * (semiperimeter - this.c)
    );
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,97 @@
...
+    }
+  }
+
+  getArea(): number {
+    return standardArea(this.width * this.height);
+  }
+}
 
+export function getInfo(figure: Figure): string {
+  return `A ${figure.color} ${figure.shape} - ${figure.getArea()}`;
```
Moroz-Dmytro, line 96: Use destructuring and create variable for this value `figure.getArea()`

Moroz-Dmytro (changes requested):

src/figuresClasses.ts
KatOlista: linter underline red if i do that

src/figuresClasses.ts
KatOlista: i have done that, but i don't see any benefits that way solution.  It don't improve readability, just do the code redundantly longer

src/figuresClasses.ts
```diff
@@ -1,19 +1,101 @@
+const ERROR_MESSAGE = {
+  wrongLength: 'length is <= 0',
+  tooLongSide: 'the longest side of a triangle is >= than a sum of two others',
+};
+
+const roundToTwoDecimals = (value: number): number => {
+  return Math.floor(value * 100) / 100;
+};
```
roman-mirzoian, line 8: If you want to round a number to two decimal places, you can use the `toFixed(2)` method:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/toFixed

src/figuresClasses.ts
```diff
@@ -1,19 +1,101 @@
+const ERROR_MESSAGE = {
+  wrongLength: 'length is <= 0',
+  tooLongSide: 'the longest side of a triangle is >= than a sum of two others',
+};
```
roman-mirzoian, line 4:
```suggestion
const ERROR_MESSAGES = {
  wrongLength: 'Side length is <= 0.',
  tooLongSide: 'The longest side of a triangle is >= than a sum of two others.',
};
```

roman-mirzoian (approved):
Well done, but pay attention to the comments.

src/figuresClasses.ts
KatOlista: i try this method at first, but result was little different then i need

BudnikOleksii (approved):
Nicely done 🔥

PR #1528, opened 2023-09-04
add figure classes

rostyslav48:


src/figuresClasses.ts
```diff
@@ -1,19 +1,83 @@
...
+  ) {
+    if (radius < 0) {
+      throw new Error('You have entered incorrect data');
+    }
+  }
+
+  getArea(): number {
+    const area = Math.PI * (this.radius * this.radius);
+
+    return Math.floor(area * 100) / 100;
```
Oleh0293, line 55: You repeat this several times. What's the problem with putting it in a separate function?

Oleh0293 (approved):

src/figuresClasses.ts
```diff
@@ -1,19 +1,83 @@
-export interface Figure {
+type FigureShape = 'triangle' | 'circle' | 'rectangle';
+type FigureColor = 'red' | 'green' | 'blue';
```
Moroz-Dmytro, line 2: Better to use `enum` here
You will be able to write `shape: FigureShape = FigureShape.triangle;` instead of `shape: FigureShape = 'triangle';`
You will avoid using "magic values"

src/figuresClasses.ts
```diff
@@ -1,19 +1,83 @@
...
+  shape: FigureShape = 'triangle';
+
+  constructor(
+    public color: FigureColor,
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    if (a * b * c <= 0) {
+      throw new Error('You have entered incorrect data');
```
Moroz-Dmytro, line 21: Give more descriptive error messages. Why data is incorrect -> because user entered negative numbers

src/figuresClasses.ts
```diff
@@ -1,19 +1,83 @@
...
+    }
+
+    if (a + b <= c || b + c <= a || a + c <= b) {
+      throw new Error(`sides ${a} ${b} ${c} can't form a triangle`);
+    }
+  }
+
+  getArea(): number {
+    const sp = (this.a + this.b + this.c) / 2;
```
Moroz-Dmytro, line 30: Don't use such short names

src/figuresClasses.ts
```diff
@@ -1,19 +1,83 @@
...
+    public color: FigureColor,
+    public radius: number,
+  ) {
+    if (radius < 0) {
+      throw new Error('You have entered incorrect data');
+    }
+  }
+
+  getArea(): number {
+    const area = Math.PI * (this.radius * this.radius);
```
Moroz-Dmytro, line 53:
```suggestion
    const area = Math.PI * (this.radius ** 2);
```

src/figuresClasses.ts
Moroz-Dmytro: Good point 👍 
Create function for this `Math.floor(... * 100) / 100;`

src/figuresClasses.ts
```diff
@@ -1,19 +1,83 @@
...
-export class Circle {
+export class Circle implements Figure {
+  shape: FigureShape = 'circle';
 
+  constructor(
+    public color: FigureColor,
+    public radius: number,
+  ) {
+    if (radius < 0) {
+      throw new Error('You have entered incorrect data');
```
Moroz-Dmytro, line 48: Give more descriptive error message
P.S. Also, would be a good idea to create variables for all error messages, some of them can be reused

Moroz-Dmytro (changes requested):

src/figuresClasses.ts
```diff
@@ -1,19 +1,100 @@
...
+  Rectangle = 'rectangle',
+}
+enum FigureColor {
+  Red = 'red',
+  Green = 'green',
+  Blue = 'blue',
+}
+type AreaCallback = () => number;
 
+function roundDownToHundreds(number: number): number {
```
```suggestion
function roundDownToHundredths(value: number): number {
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,100 @@
...
 
+  getArea(): number {
+    return this.width * this.height;
+  }
 }
 
-export function getInfo(figure) {
+export function getInfo(figure: Figure): string {
+  const { color, shape } = figure;
+  const area: number = figure.getArea();
```
Moroz-Dmytro, line 97:
```suggestion
  const area = figure.getArea();
```

Moroz-Dmytro (approved):

PR #1531, opened 2023-09-04
Solution

Oleh0293:


src/figuresClasses.ts
```diff
@@ -1,19 +1,101 @@
...
+      throw new Error(ERROR_MESSAGE);
+    }
 
+    if (this.a + this.b <= this.c
+      || this.b + this.c <= this.a
+      || this.c + this.a <= this.b) {
+      throw new Error(
+        `sides ${this.a}, ${this.b} and ${this.c} can't form a triangle`,
+      );
+    }
```
BudnikOleksii, line 42: You don't need to point to `this` in this case as you get them from the constructor

src/figuresClasses.ts
```diff
@@ -1,19 +1,101 @@
-export interface Figure {
+const ERROR_MESSAGE = 'Value cannot be <= 0!';
```
BudnikOleksii, line 1: Specify what kind of error message is this variable as it's not general message

src/figuresClasses.ts
```diff
@@ -1,19 +1,101 @@
...
+      || this.b + this.c <= this.a
+      || this.c + this.a <= this.b) {
+      throw new Error(
+        `sides ${this.a}, ${this.b} and ${this.c} can't form a triangle`,
+      );
+    }
+  }
+
+  getArea(): number {
+    const s: number = (this.a + this.b + this.c) / 2;
```
BudnikOleksii, line 46: Avoid one-letter variable names

BudnikOleksii (changes requested):
Good work!
Just a few minor fixes need to be done

src/figuresClasses.ts
```diff
@@ -1,19 +1,101 @@
-export interface Figure {
+const ERROR_MESSAGE: string = 'Value cannot be <= 0!';
```
```suggestion
const ERROR_MESSAGE = 'Value cannot be <= 0!';
```

Moroz-Dmytro: You don't need to set type manually. Also as previous mentor mention, you need to give more descriptive  name to this variable

src/figuresClasses.ts
```diff
@@ -1,19 +1,101 @@
...
+
+function circledNumber(number: number): number {
```
Moroz-Dmytro, line 3: function name should start with a verb

src/figuresClasses.ts
```diff
@@ -1,19 +1,101 @@
...
 
+    return circledNumber(area);
+  }
 }
 
-export function getInfo(figure) {
+export function getInfo(figure: Figure): string {
+  const area = figure.getArea();
 
+  return `A ${figure.color} ${figure.shape} - ${area}`;
```
Moroz-Dmytro, line 100: Use destructuring

Moroz-Dmytro (changes requested):

PR #1532, opened 2023-09-04
add Figure interface and 3 classes - Triangle, Circle and Rectangle

neovendano:


src/figuresClasses.ts
```diff
@@ -1,19 +1,106 @@
...
+
+  color: Color;
+
+  a: number;
+
+  b: number;
+
+  c: number;
+
+  perimeter: number;
```
BudnikOleksii, line 26: Remove unnecessary blank lines

src/figuresClasses.ts
```diff
@@ -1,19 +1,106 @@
...
+
+  constructor(color: Color, a: number, b: number, c: number) {
+    this.shape = 'triangle';
+    this.color = color;
+    this.a = a;
+    this.b = b;
+    this.c = c;
```
BudnikOleksii, line 33: [CODE KNOWLEDGE] - When you declare a constructor parameter with an access modifier TypeScript implicitly creates a property on the class with the same name as the parameter and assigns the parameter value to that property. You don't need to do it manually.
image

BudnikOleksii (changes requested):
Good work 👍 
Let's improve a bit

PR #1540, opened 2023-09-05
Solution

viktoria-roik:


src/figuresClasses.ts
```diff
@@ -1,19 +1,102 @@
...
+    return formattedShapeValues(rectangleArea);
+  }
 }
 
-export function getInfo(figure) {
+export function getInfo(figure: Figure): string {
+  const area = figure.getArea();
 
+  return `A ${figure.color} ${figure.shape} - ${area}`;
 }
```
alexander-ignatow, line 102: Just for the future - it is more convenient to use the destruction (via arguments or via constants)

alexander-ignatow (approved):
Looks good 👍 

See some minor comments below

PR #1541, opened 2023-09-06
add task solution

vlad-dan1589:


src/figuresClasses.ts
```diff
@@ -1,19 +1,79 @@
 export interface Figure {
-
+  shape: 'triangle' | 'circle' | 'rectangle';
```
maxim2310, line 2: better use enum for this

src/figuresClasses.ts
```diff
@@ -1,19 +1,79 @@
...
+  color: 'red' | 'green' | 'blue';
+  getArea(): number;
 }
 
-export class Triangle {
+export class Triangle implements Figure {
+  public shape: Figure['shape'] = 'triangle';
```
maxim2310, line 8: and use enum here directly

src/figuresClasses.ts
```diff
@@ -1,19 +1,79 @@
...
+    public b: number,
+    public c: number,
+  ) {
+    if (a <= 0 || b <= 0 || c <= 0) {
+      throw new Error('Invalid parameters');
+    }
+
+    if (a + b <= c || a + c <= b || b + c <= a) {
+      throw new Error('Invalid parameters');
+    }
```
maxim2310, line 22: this not make sense if you have the same error, you need or combine it or change more meaningful error

maxim2310 (changes requested):
Good work!!! just refactor code a bit

src/figuresClasses.ts
vlad-dan1589: done

src/figuresClasses.ts
vlad-dan1589: done

src/figuresClasses.ts
vlad-dan1589: done

maxim2310 (commented):
now looks much better, good job!!!

maxim2310 (approved):

PR #1546, opened 2023-09-09
add task solution

Anmonnn:


src/figuresClasses.ts
```diff
@@ -1,19 +1,87 @@
...
 
+export interface Figure {
+  shape: Shape
+  color: Color
+  getArea(): number;
 }
 
-export class Triangle {
+export class Triangle implements Figure {
+  shape: Shape;
```
TarasHoliuk, line 11: you may define the shape value right here - FYI

src/figuresClasses.ts
```diff
@@ -1,19 +1,87 @@
...
+    public width: number,
+    public height: number,
+  ) {
+    this.color = color;
+    this.height = height;
+    this.width = width;
+    this.shape = 'rectangle';
+
+    if (width <= 0 || height <= 0) {
+      throw new Error('your error message');
```
TarasHoliuk, line 76: write real error messages everywhere

TarasHoliuk (changes requested):

anastasiiavorobiova (approved):
Good job!

PR #1550, opened 2023-09-14
figure classes was realised

KirillMaslov:


VitaliyBodnar (changes requested):
Almost done!

src/figuresClasses.ts
```diff
@@ -1,19 +1,93 @@
...
+    c = 0,
+  ) {
+    this.color = color;
+    this.area = calculateArea(a, b, c);
+
+    if (a <= 0 || b <= 0 || c <= 0
+      || a >= b + c
+      || b >= a + c
+      || c >= b + a) {
+      throw new Error('');
```
sTorba24, line 36: explain the error

src/figuresClasses.ts
```diff
@@ -1,19 +1,93 @@
...
 
+  constructor(
+    public color: Figure['color'],
+    radius: number = 0,
+  ) {
+    this.color = color;
+    this.area = Math.PI * radius ** 2;
+
+    if (radius <= 0) {
+      throw new Error('');
```
sTorba24, line 58: same here

src/figuresClasses.ts
```diff
@@ -1,19 +1,93 @@
...
+  constructor(
+    public color: Figure['color'],
+    width: number,
+    height: number,
+  ) {
+    this.color = color;
+    this.area = width * height;
+
+    if (width <= 0 || height <= 0) {
+      throw new Error('');
```
sTorba24, line 81: same

src/figuresClasses.ts
sTorba24: also use conditions before setting values

you don't need to initialize smth if condition is wrong

sTorba24 (changes requested):

VitaliyBodnar (approved):
Good job!

PR #1555, opened 2023-09-23
Add task solution.

MaksymBevskyi:


src/figuresClasses.ts
```diff
@@ -1,19 +1,92 @@
...
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    this.shape = 'triangle';
+    checkValidSides([a, b, c]);
+
+    const sides = [a, b, c].sort((x: number, y: number) => x - y);
+
+    if (sides[2] >= sides[0] + sides[1]) {
```
VitaliyBodnar, line 36: make sense to use array destructuring here as it will make a code cleaner

src/figuresClasses.ts
```diff
@@ -1,19 +1,92 @@
...
+      throw new Error(
+        'The longest side of a triangle should be '
+        + '>= than a sum of two others',
+      );
+    }
+  }
 
+  getArea(): number {
+    const { a, b, c } = this;
+    const s = (a + b + c) / 2;
```
VitaliyBodnar, line 46: it's a bad practice to naming variable with only one name

VitaliyBodnar (changes requested):

nazarmatsevych (approved):
👍

PR #1556, opened 2023-09-23
Solution

YuraMisura:


src/figuresClasses.ts
```diff
@@ -1,19 +1,76 @@
-export interface Figure {
-
+enum Shape {
+  Triangle = 'triangle',
+  Circle = 'circle',
+  Rectangle = 'rectangle',
 }
 
-export class Triangle {
+type Color = 'red' | 'green' | 'blue';
```
Andriy-Shymkiv, line 7: you can create same enum for color as you create for shape above

src/figuresClasses.ts
```diff
@@ -1,19 +1,76 @@
...
+  }
+
+  getArea(): number {
+    return Math.round(this.width * this.height);
+  }
+}
 
+export function getInfo(figure: Figure): string {
+  return `A ${figure.color} ${figure.shape} - ${figure.getArea()}`;
 }
```
Andriy-Shymkiv, line 76: better to use destructuring in such cases:
```suggestion
export function getInfo(figure: Figure): string {
  const { color, shape, getArea } = figure;
  return `A ${color} ${shape} - ${getArea()}`;
}
```

Andriy-Shymkiv (changes requested):

VitaliyBodnar (changes requested):
- you have an error with getArea context 
Screenshot 2023-09-23 at 17 34 57

src/figuresClasses.ts
```diff
@@ -1,19 +1,82 @@
...
+  getArea(): number {
+    return Math.round(this.width * this.height);
+  }
 }
 
-export function getInfo(figure) {
+export function getInfo(figure: Figure): string {
+  const { color, shape, getArea } = figure;
 
+  return `A ${color} ${shape} - ${getArea()}`;
```
VitaliyBondarenko1982, line 81: after destructuring `this` was lost.

```suggestion
  const { color, shape, getArea } = figure;

  return `A ${color} ${shape} - ${getArea.call(figure)}`;
```
you can reassign this with call method

VitaliyBondarenko1982 (changes requested):

VitaliyBondarenko1982 (approved):
GJ!

PR #1569, opened 2023-10-04
add task solution

MishaLehotskyi:


src/figuresClasses.ts
```diff
@@ -1,19 +1,101 @@
...
+    this.color = color;
+    this.a = a;
+    this.b = b;
+    this.c = c;
+
+    if (this.a <= 0 || this.b <= 0 || this.c <= 0) {
+      throw new Error('All sides of triangle must be greater than 0!!!');
+    }
+
+    const max = Math.max(this.a, this.b, this.c);
```
loralevitska, line 31: DRY - don't repeat yourself
Destructure values from `this` and reuse it
Fix all repeatable code and use destructuring

loralevitska (changes requested):
Almost done) Let's improve

maxim2310 (approved):
Good work!!!

PR #1571, opened 2023-10-04
add ts_figures_classes

Lazarend:


src/figuresClasses.ts
```diff
@@ -1,19 +1,83 @@
 export interface Figure {
-
+  shape: string;
+  color: string;
```
maxim2310, line 3: you need to define certain strings for this, create two uniq type Color and Shape

maxim2310 (changes requested):
Good work!!! Just improve types

src/figuresClasses.ts
Lazarend: Hi. Ok, I`ve done create two unique type Color and Shape.

src/figuresClasses.ts
```diff
@@ -1,19 +1,86 @@
...
+    }
+
+    this.color = color;
+    this.a = a;
+    this.b = b;
+    this.c = c;
+  }
+
+  getArea(): number {
+    const s = (this.a + this.b + this.c) / 2;
```
VitaliyBodnar, line 33: it's a bad practice to name variable with only one letter

src/figuresClasses.ts
```diff
@@ -1,19 +1,86 @@
...
 
-export function getInfo(figure) {
+export class Rectangle implements Figure {
+  shape: Shape = 'rectangle';
+
+  color: Color;
+
+  width: number;
+
+  height: number;
```
VitaliyBodnar, line 67: a lot of spaces, it can be simplify 
```suggestion
  shape: Shape = 'rectangle';
  color: Color;

  width: number;
  height: number;
```

VitaliyBodnar (changes requested):

src/figuresClasses.ts
Lazarend: You are absolutely right. A good practice is to name variables more clearly and descriptively. Here is the updated code with more descriptive variable names.

src/figuresClasses.ts
Lazarend: I`m add a '@typescript-eslint/lines-between-class-members': 'off' to .eslintrc.js

src/figuresClasses.ts
```diff
@@ -1,19 +1,90 @@
-export interface Figure {
+type Color = 'red' | 'green' | 'blue';
+type Shape = 'triangle' | 'circle' | 'rectangle';
 
+export interface Figure {
+  shape: Shape;
+  color: Color;
+  getArea(): number;
```
Andriy-Shymkiv, line 7:
```suggestion
  getArea: () => number;
```

Andriy-Shymkiv (approved):
good job. added one suggestion, approved in advance

PR #1574, opened 2023-10-07
Dev

donizer:


src/figuresClasses.ts
```diff
@@ -1,19 +1,92 @@
...
+  height: number;
+
+  constructor(color: string, width: number, height: number) {
+    if (width <= 0 || height <= 0) {
+      throw new Error('Width and height should be positive numbers.');
+    }
+
+    this.color = color;
+    this.width = width;
+    this.height = height;
```
VitaliyBodnar, line 82: can be less 
```suggestion
  constructor(
   private color: string,
   private width: number, 
   private height: number
  ) {
    if (width <= 0 || height <= 0) {
      throw new Error('Width and height should be positive numbers.');
    }
```
check other places

src/figuresClasses.ts
```diff
@@ -1,19 +1,92 @@
...
+
+    return Math.floor(area * 100) / 100;
+  }
 }
 
-export class Rectangle {
+export class Circle implements Figure {
+  shape = 'circle';
+
+  color: string;
```
VitaliyBodnar, line 46: please create type for shape and color

VitaliyBodnar (changes requested):

VitaliyBondarenko1982 (approved):
Looks good

PR #1575, opened 2023-10-08
Develop

markiuan:


src/figuresClasses.ts
```diff
@@ -1,19 +1,100 @@
-export interface Figure {
+type Shapes = 'triangle'|'circle'|'rectangle';
+type Colors = 'red'|'green'|'blue';
```
TarasHoliuk, line 2:
```suggestion
type Shape = 'triangle' | 'circle' | 'rectangle';
type Color = 'red' | 'green' | 'blue';
```
use type names in singular form

TarasHoliuk (approved):
GJ :+1: 
Added 1 comment FYI

PR #1579, opened 2023-10-11
created calculator that counts an area of circle | triangle | rectangle

OPTIMISTIXX:


src/figuresClasses.ts
```diff
@@ -1,19 +1,74 @@
 export interface Figure {
-
+  shape: string;
+  color: string;
```
maxim2310, line 3: these variable must be not random strings, make it more specific

maxim2310 (changes requested):

maxim2310 (approved):
Good work!!!

PR #1583, opened 2023-10-15
add task solution

mpohorenyi:


src/figuresClasses.ts
```diff
@@ -1,19 +1,76 @@
 export interface Figure {
-
+  shape: string,
+  color: string,
+  getArea(): number,
 }
 
-export class Triangle {
+export class Triangle implements Figure {
+  shape: string = 'triangle';
```
vasylmiskiv, line 8: You can use `enum` data type for the shape for each class

vasylmiskiv (approved):

PR #1585, opened 2023-10-16
figures classes

whiovt:


anastasiia-tilikina (changes requested):
Make sure that you've passed all tests before we move on

PR #1587, opened 2023-10-16
add solution

Helen-arch:


src/figuresClasses.ts
```diff
@@ -1,19 +1,73 @@
...
-export class Triangle {
+export class Triangle implements Figure {
+  shape: 'triangle' = 'triangle';
 
-}
+  constructor(
+    public color: Color,
+    public x: number,
+    public y: number,
+    public z: number,
```
adioskv, line 17: Try to use a, b, c... in math tasks, because x, y, z is a coords

StasSokolov: will be better use `a, b, c` in this case

StasSokolov (approved):
Great job, i'm approving it🔥

PR #1589, opened 2023-10-16
made task solution

sintax1s:


yulabakun (approved):
Great job!

src/figuresClasses.ts
```diff
@@ -1,19 +1,68 @@
 export interface Figure {
-
+  shape: 'triangle' | 'circle' | 'rectangle',
```
vasylmiskiv, line 2: Would be better to use `enum` data type here 
And then you don't need to write like so `public shape: 'triangle' = 'triangle'` in the classes. You can just use a specific field like `public shape: Shape.Triangle` etc

src/figuresClasses.ts
```diff
@@ -1,19 +1,68 @@
...
 }
 
-export function getInfo(figure) {
+export class Rectangle implements Figure {
+  public shape: 'rectangle' = 'rectangle';
+
+  constructor(
+    public color: 'red' | 'green' | 'blue',
+    public width: number = 0,
+    public height: number = 0,
```
vasylmiskiv, line 54: You don't need to use values by default in this case
```suggestion
    public width: number,
    public height: number,
```

vasylmiskiv (approved):

PR #1598, opened 2023-10-16
task solution

dolphinnq:


src/figuresClasses.ts
```diff
@@ -1,19 +1,82 @@
...
+    public color: Color,
+    public radius: number,
+  ) {
+    if (radius <= 0) {
+      throw new Error('Value must be positive number');
+    }
+  }
 
+  getArea(): number {
+    return Math.floor(Math.PI * (this.radius ** 2) * 100) / 100;
```
HardCodWorker, line 58: would use destructuring in all get area methods

HardCodWorker (approved):

vlevik (approved):

src/figuresClasses.ts
```diff
@@ -1,19 +1,82 @@
-export interface Figure {
-
+enum Shape {
+  Triangle = 'triangle',
+  Circle = 'circle',
+  Rectangle = 'rectangle',
```
vasylmiskiv, line 4: 👍

src/figuresClasses.ts
```diff
@@ -1,19 +1,82 @@
...
+
+    if (a + b <= c || a + c <= b || b + c <= a) {
+      throw new Error('Invalid values');
+    }
+  }
 
+  getArea(): number {
+    const p = (this.a + this.b + this.c) / 2;
+
+    return +Math.sqrt(p * (p - this.a) * (p - this.b) * (p - this.c))
```
vasylmiskiv, line 40: Try to avoid to write like so with unary operator

src/figuresClasses.ts
```diff
@@ -1,19 +1,82 @@
...
+      throw new Error('Values must be positive numbers');
+    }
+  }
+
+  getArea(): number {
+    return this.height * this.width;
+  }
+}
 
+export function getInfo(figure: Figure): string {
```
vasylmiskiv, line 80: You could destructure the properties from the `figure`

vasylmiskiv (approved):
Looks nice 🚀

PR #1600, opened 2023-10-16
solution ts_figures_classes task

ivaZaiets:


src/figuresClasses.ts
```diff
@@ -1,19 +1,88 @@
...
+    public color: string,
+    public width: number,
+    public height: number,
+  ) {
+    this.shape = 'rectangle';
+    this.area = this.getArea();
 
-export function getInfo(figure) {
+    if (this.width < 0 || this.height < 0) {
+      throw new Error();
```
YaroslavYarynych, line 75: Good job! But please write what exactly error will be if our properties is not correct. It would be better for understanding.

throw new Error('Error message');

YaroslavYarynych (approved):
Good job!

src/figuresClasses.ts
```diff
@@ -1,19 +1,88 @@
 export interface Figure {
-
+  shape: string,
+  color: string,
```
polosanya, line 3: Would be great to specify what exactly values they can be

polosanya (changes requested):
Nice! Let's improve your code a bit)

StasSokolov (approved):
Looks good to me, i'm approving it🔥

PR #1603, opened 2023-10-16
solution

Ayogoharo:


Drach-Illya (approved):
Stellar job. Peer Approved!

src/figuresClasses.ts
```diff
@@ -1,19 +1,72 @@
...
+  constructor(
+    public color: Color,
+    public aSide: number,
+    public bSide: number,
+    public cSide: number,
+  ) {
+    const sides = [aSide, bSide, cSide].sort((a: number, b: number) => a - b);
+
+    if ((sides[0] + sides[1]) <= sides[2] || sides.find((el) => el <= 0)) {
+      throw new Error('wrong values for triangle');
```
polosanya, line 22: It would be great to specify why exactly they're wrong
image

src/figuresClasses.ts
```diff
@@ -1,19 +1,72 @@
...
-export class Rectangle {
+export class Circle implements Figure {
+  public shape: Shape = 'circle';
 
+  constructor(
+    public color: Color,
+    public radius: number,
+  ) {
+    if (radius <= 0) {
+      throw new Error('wrong values for circle');
```
polosanya, line 43: Fix all cases

polosanya (changes requested):
Almost perfect! One small but important moment left

PR #1610, opened 2023-10-17
implement task solution

Drach-Illya:


src/figuresClasses.ts
```diff
@@ -1,19 +1,74 @@
...
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    if (a >= b + c
+        || b >= a + c
+        || c >= a + b
+        || a <= 0
+        || b <= 0
+        || c <= 0) {
```
polosanya, line 24: It looks a bit complicated. Would be great to create a separate function for checking the parameters

src/figuresClasses.ts
```diff
@@ -1,19 +1,74 @@
...
-export class Rectangle {
+export class Circle implements Figure {
+  public shape: Shape = 'circle';
+
+  constructor(
+    public color: Color,
+    public a: number,
+  ) {
+    if (a <= 0) {
+      throw new Error('Cannot assemble the circle with given parameters');
```
polosanya, line 45: Would be great to specify why exactly we cannot

polosanya (approved):
Good job! I'm approving it, but please pay attention to the comments)

PR #1614, opened 2023-10-17
Solution

hrysta:
Hi) I committed a solution. But I fixed the test about circle.getArea())
    from  .toBe(113.09); to     .toBe(113.1) ,because of wrong rounding.

### DIFF
--- a/src/figuresClasses.ts
+++ b/src/figuresClasses.ts
@@ -1,19 +1,84 @@
 export interface Figure {
-
+  shape: string;
+  color: string;
+  getArea(): number;
 }
 
-export class Triangle {
+export class Triangle implements Figure {
+  shape: string = 'triangle';
 
+  color: string;
+
+  a: number;
+
+  b: number;
+
+  c: number;
+
+  constructor(color: string, a: number, b: number, c: number) {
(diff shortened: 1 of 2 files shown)


src/figuresClasses.ts
```diff
@@ -1,19 +1,84 @@
...
+
+  getArea(): number {
+    return this.width * this.height;
+  }
+}
 
+export function getInfo(figure: Figure): string {
+  return Number.isInteger(figure.getArea())
+    ? `A ${figure.color} ${figure.shape} - ${figure.getArea()}`
+    : `A ${figure.color} ${figure.shape} - ${figure.getArea().toFixed(2)}`;
```
Kuper18, line 83: it would be better to use `toFixed()`  method inside getArea function and you don't need to check here if number is integer

Kuper18 (changes requested):
Good job

StasSokolov (changes requested):
Almost masterpiece, let's improve your code a bit more before i'm approving it🔥

StasSokolov (approved):
Looks good to me, i'm approving it🔥

PR #1616, opened 2023-10-18
add tack solution

annabasenyuk:


src/figuresClasses.ts
```diff
@@ -1,19 +1,79 @@
...
+
+  constructor(
+    public color: Color,
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    if (a <= 0 || b <= 0 || c <= 0 || a + b <= c || a + c <= b || b + c <= a) {
+      throw new Error('Input data is incorrect');
+    }
```
superpooperxxx, line 26: [non blocker] I would separate into 2 `if` with different error messages.

src/figuresClasses.ts
```diff
@@ -1,19 +1,79 @@
...
+  }
 
-export class Circle {
+  getArea(): number {
+    const p: number = (this.a + this.b + this.c) / 2;
+    const s: number = +Math.sqrt(p * (p - this.a) * (p - this.b) * (p - this.c))
```
superpooperxxx, line 31: Instead of `.toFixed(2)`. Better `Math.round(s * 100) / 100`. Then you don't need to convert string after `toFixed`

src/figuresClasses.ts
```diff
@@ -1,19 +1,79 @@
...
+    public color: Color,
+    public r: number,
+  ) {
+    if (r <= 0) {
+      throw new Error('Input data is incorrect');
+    }
+  }
+
+  getArea(): number {
+    const s: number = ((this.r ** 2) * Math.PI);
```
superpooperxxx, line 51: Use more descriptive variable names. Instead of `r` => `radius`, `s` => `square` and etc. Fix everywhere

superpooperxxx (changes requested):
Good job🔥 Let's improve your code a little bit

PR #1617, opened 2023-10-18
add solution figures

A1eeex:


src/figuresClasses.ts
```diff
@@ -1,19 +1,77 @@
 export interface Figure {
-
+  shape: string;
+  color: string;
```
superpooperxxx, line 3: I think Union type fits better here, because you know for sure which values can take shape and color

src/figuresClasses.ts
```diff
@@ -1,19 +1,77 @@
...
+    public c: number,
+  ) {
+    if (this.a <= 0 || this.b <= 0 || this.c <= 0) {
+      throw new Error('Parameters must be greater than 0');
+    }
+
+    if (
+      this.a + this.b <= this.c
+      || this.a + this.c <= this.b
+      || this.c + this.b <= this.a
```
superpooperxxx, line 23: Let's put each of checks into separate variables to improve readability

src/figuresClasses.ts
```diff
@@ -1,19 +1,77 @@
...
-export class Circle {
+    ) {
+      throw new Error(`It is not possible
+       to construct a triangle with these sides`);
+    }
+  }
 
+  getArea(): number {
+    const p = (this.a + this.b + this.c) / 2;
+    const s = Math.sqrt(p * ((p - this.a) * (p - this.b) * (p - this.c)));
```
superpooperxxx, line 33: Let's use more descriptive variable names

src/figuresClasses.ts
```diff
@@ -1,19 +1,77 @@
...
+
+    return +(s).toFixed(2);
```
superpooperxxx, line 35: Let's create a function that returning `Math.floor(value * 100) / 100;` and use it everywhere

src/figuresClasses.ts
```diff
@@ -1,19 +1,77 @@
 export interface Figure {
-
+  shape: string;
+  color: string;
+  getArea(): number;
 }
 
-export class Triangle {
+export class Triangle implements Figure {
+  shape = 'triangle';
```
superpooperxxx, line 8: What type does shape have ?

superpooperxxx (changes requested):
Almost perfect🤗

superpooperxxx (approved):
🔥

PR #1627, opened 2023-10-22
completed task

mifyy:


src/figuresClasses.ts
```diff
@@ -1,19 +1,91 @@
 export interface Figure {
-
+  shape: string;
+  color: string;
```
superpooperxxx, line 3: Let's make type of shape and color `union type` because you know exactly what values it might be

src/figuresClasses.ts
```diff
@@ -1,19 +1,91 @@
...
+  getArea(): number;
```
superpooperxxx, line 4: `getArea` is a function that returns a number

src/figuresClasses.ts
```diff
@@ -1,19 +1,91 @@
...
 }
 
-export class Triangle {
+export class Triangle implements Figure {
+  shape: string = 'triangle';
```
superpooperxxx, line 8: Again, shape is not just string. Fix everywhere

src/figuresClasses.ts
```diff
@@ -1,19 +1,91 @@
...
 
-}
+  color: string;
+
+  a: number;
+
+  b: number;
+
+  c: number;
```
superpooperxxx, line 16: No need in this. Fix everywhere

src/figuresClasses.ts
```diff
@@ -1,19 +1,91 @@
...
 
-export class Circle {
+  constructor(color: string, a: number, b: number, c: number) {
+    if (a <= 0 || b <= 0 || c <= 0 || a + b <= c || a + c <= b || b + c <= a) {
+      throw new Error('Invalid triangle');
+    }
```
superpooperxxx, line 21: Let's split into 2 `if` because you might want to have different messages for different cases

src/figuresClasses.ts
```diff
@@ -1,19 +1,91 @@
...
+    this.a = a;
+    this.b = b;
+    this.c = c;
+    this.color = color;
+  }
+
+  getArea(): number {
+    const s = (this.a + this.b + this.c) / 2;
+    const area = Math.sqrt(s * (s - this.a) * (s - this.b) * (s - this.c));
+    const calc = Math.floor(area * 100) / 100;
```
superpooperxxx, line 32: 1. Let's give more descriptive names, than just `calc`
2. As you use this logic `Math.floor(area * 100) / 100` in multiple places, why don't you put it into separate function ?

superpooperxxx (changes requested):
Almost there🤗

src/figuresClasses.ts
mifyy: I see but it won't let me write without blanks. Am I missing something?
image

src/figuresClasses.ts
mifyy: Sure, I have fixed these issues!

src/figuresClasses.ts
superpooperxxx: I mean you can completely get rid of it

src/figuresClasses.ts
```diff
@@ -1,19 +1,101 @@
...
 
-export class Rectangle {
+export class Circle implements Figure {
+  color: 'red' | 'green' | 'blue';
+
+  radius: number;
+
+  shape: 'circle' = 'circle';
+
+  constructor(color: 'red' | 'green' | 'blue', radius: number) {
```
superpooperxxx, line 57: You are duplicating code, take the type for color from Figure

superpooperxxx (changes requested):
One more shot😉

src/figuresClasses.ts
mifyy: Sure, I have corrected this issue!

superpooperxxx (approved):
🔥

PR #1644, opened 2023-10-30
figures

IvanOrinchak:


src/figuresClasses.ts
```diff
@@ -1,19 +1,67 @@
...
+      throw new Error('The sides of the triangle are not correctly specified');
+    }
+  }
 
-export class Circle {
+  getArea():number {
+    const p:number = (this.a + this.b + this.c) / 2;
+    const area = Math.sqrt(p * (p - this.a) * (p - this.b) * (p - this.c));
 
+    return Math.floor(area * 100) / 100;
```
anastasiia-tilikina, line 25: it seems like you repeat this code. consider moving it to separate helper.

src/figuresClasses.ts
```diff
@@ -1,19 +1,67 @@
 export interface Figure {
-
+  shape:string,
+  color:string,
+  getArea():number,
```
anastasiia-tilikina, line 4:
```suggestion
  shape: string,
  color: string,
  getArea(): number,
```

anastasiia-tilikina: fix everywhere

src/figuresClasses.ts
```diff
@@ -1,19 +1,67 @@
...
 }
 
-export class Triangle {
+export class Triangle implements Figure {
+  shape = 'triangle';
```
anastasiia-tilikina, line 8: it would be great to create enum for these values and use it.

like this:
```typescript

enum Shape {
TRIANGLE = 'triangle'
//...
}
```

anastasiia-tilikina (changes requested):
GJ

just a few suggestions)

PR #1646, opened 2023-10-31
Homework: TS Figures Classes

ViktorBykh:


src/figuresClasses.ts
```diff
@@ -1,19 +1,110 @@
...
+
+    if (a <= 0 || b <= 0 || c <= 0) {
+      throw new Error(
+        `No side can be shorter than 0: a: ${a}, b: ${b}, c: ${c}`,
+      );
+    }
+
+    const sortedArray
+      = [this.a, this.b, this.c].sort((first, secound) => first - secound);
+    const max = sortedArray[sortedArray.length - 1];
```
sarakhmen, line 26: Is it possible to use Math.max() here?

src/figuresClasses.ts
```diff
@@ -1,19 +1,110 @@
...
+        break;
+
+      default:
+        if (a + c <= max) {
+          throw new Error(
+            `The longest side of a triangle cannot be
+             shorter than the sum of the other two:
+             a: ${a}, b: ${b}, c: ${c}`,
+          );
+        }
```
sarakhmen, line 55: please rewrite it so that there is no code duplication and unnecessary switch block

src/figuresClasses.ts
```diff
@@ -1,19 +1,110 @@
...
+  shape: Shape,
 
+  getArea(): number;
 }
 
-export class Triangle {
+export class Triangle implements Figure {
+  shape: Shape;
+
+  constructor(public color: Color, public a: number, public b: number,
```
sarakhmen, line 14: Also consider refactoring this constructor because it looks pretty overloaded with code lines. You can create methods and move the logic there

sarakhmen (changes requested):
Good job, but let's improve your solution :)

sarakhmen (approved):
Good job :)

PR #1651, opened 2023-11-03
tests done

AnastasiiaKriuchkova:


src/figuresClasses.ts
```diff
@@ -1,19 +1,74 @@
...
 
 export class Triangle {
+  public shape: Shape = 'triangle';
+
+  getArea(): number {
+    const p = (this.a + this.b + this.c) / 2;
+
+    return Math.floor(Math.sqrt(p * (p - this.a)
+     * (p - this.b) * (p - this.c)) * 100) / 100;
+  }
```
Esceype, line 18: You use `Math.floor(someDigit * 100) / 100` - code like that in every class lets move it to separate function.

src/figuresClasses.ts
```diff
@@ -1,19 +1,74 @@
...
 
+  constructor(
+    public color: Color,
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    const perimetr = this.a + this.b + this.c;
```
Esceype, line 26: Inside constructor you can use it without  `this`
```suggestion
    const perimetr = a + b + c;
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,74 @@
...
 }
 
 export class Circle {
+  public shape: Shape = 'circle';
+
+  constructor(
+    public color: Color,
+    public radius: number,
+  ) {
+    if (this.radius <= 0) {
```
Esceype, line 42:
```suggestion
    if (radius <= 0) {
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,74 @@
...
+      throw new Error('Radius must be positive');
+    }
+  }
 
+  getArea(): number {
+    return Math.floor(Math.PI * this.radius * this.radius * 100) / 100;
```
Esceype, line 48:
```suggestion
    return Math.floor(Math.PI * ( this.radius ** 2 ) * 100) / 100;
```

Esceype (changes requested):
GJ!
Let`s make some impovements

Andriy-Shymkiv (approved):

PR #1654, opened 2023-11-05
completed the task

GeorgeSavostikov:


src/figuresClasses.ts
```diff
@@ -1,19 +1,94 @@
 export interface Figure {
-
+  shape: string;
```
anastasiiavorobiova, line 2:
```suggestion
  shape: Shape;
```
Consider creating a separate type or an enum for the shape

src/figuresClasses.ts
```diff
@@ -1,19 +1,94 @@
...
+  color: string;
```
anastasiiavorobiova, line 3: Do the same for the color

src/figuresClasses.ts
```diff
@@ -1,19 +1,94 @@
...
+  getArea(): number;
 }
 
 export class Triangle {
+  shape: string;
```
anastasiiavorobiova, line 8:
```suggestion
  public shape: Shape;
```
You can use typescript features here and the same for other arguments

src/figuresClasses.ts
```diff
@@ -1,19 +1,94 @@
...
 
+    if (a <= 0 || b <= 0 || c <= 0 || (bigNum >= (a + b + c - bigNum))) {
+      throw new Error('your error message');
+    }
+
+    this.shape = 'triangle';
+    this.color = color;
+    this.a = a;
+    this.b = b;
+    this.c = c;
```
anastasiiavorobiova, line 29:
```suggestion

```

anastasiiavorobiova (changes requested):
Great work! Added a few comments to fix

etojeDenys (approved):
lgtm

PR #1656, opened 2023-11-07
done

kasianeno:


src/figuresClasses.ts
```diff
@@ -1,19 +1,80 @@
...
+type Color = 'red' | 'green' | 'blue';
 
+export interface Figure {
+  shape: Shape;
+  color: Color;
+  getArea(): number;
 }
 
 export class Triangle {
+  shape: Shape = 'triangle';
```
maxim2310, line 11: all figures class must implement class Figure 
```suggestion
export class Triangle implements Figure {
  shape = 'triangle';
```

maxim2310 (changes requested):

src/figuresClasses.ts
kasianeno: shape won`t be working if i will remove type

etojeDenys (approved):
lgtm

PR #1658, opened 2023-11-08
Added figures classes

IonShive:


src/figuresClasses.ts
```diff
@@ -1,19 +1,82 @@
...
+type Shape = 'triangle' | 'circle' | 'rectangle';
+type Color = 'red' | 'green' | 'blue';
 
+export interface Figure {
+  shape: Shape;
+  color: Color;
+  getArea(): number;
 }
 
 export class Triangle {
```
maxim2310, line 10: all figures must implement class Figure

maxim2310 (changes requested):
Good work, but lets make some improvement

src/figuresClasses.ts
IonShive: Thanks for the answer. Tried to fix this problem.

etojeDenys (approved):
lgtm

PR #1664, opened 2023-11-14
Solution

alina-kabanets:


src/figuresClasses.ts
```diff
@@ -1,19 +1,79 @@
...
+  getArea(): number {
+    const { a, b, c } = this;
+    const p = (a + b + c) / 2;
+    const area = p * (p - a) * (p - b) * (p - c);
 
+    return +(Math.sqrt(area).toFixed(2));
+  }
 }
 
 export class Circle {
```
maxim2310, line 40:
```suggestion
export class Circle implements Figure {
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,79 @@
...
+      throw new Error('Incorrect dimentions');
+    }
+  }
+
+  getArea(): number {
+    return Math.floor(Math.PI * (this.radius ** 2) * 100) / 100;
+  }
 }
 
 export class Rectangle {
```
maxim2310, line 57:
```suggestion
export class Rectangle implements Figure {
```

maxim2310 (changes requested):

src/figuresClasses.ts
alina-kabanets: Fixed. Thank you for noticing)

anastasiiavorobiova (approved):
Awesome work!

PR #1667, opened 2023-11-16
task solved

TommasoGiacopini:


src/figuresClasses.ts
```diff
@@ -1,19 +1,79 @@
...
+    private height: number,
+    private width: number,
+  ) {
+    if (Math.min(height, width) <= 0) {
+      throw new Error('Invalid rectangle');
+    }
+  }
+
+  getArea(): number {
+    return Math.floor(this.height * this.width * 100) / 100;
```
anastasiia-tilikina, line 73: seems like you have shared logic related to the rounding of area. consider moving it to helper.

anastasiia-tilikina (changes requested):
Good job! let's make some small improvement

src/figuresClasses.ts
```diff
@@ -1,19 +1,85 @@
...
-
+export interface Figure {
+  shape: Shape;
+  color: Color;
+  getArea(): number;
 }
 
-export class Circle {
+export class Triangle implements Figure {
+  shape: Shape = 'triangle';
```
maxim2310, line 15: not necessary because you already declared type in Figure class 
```suggestion
  shape = 'triangle';
```

maxim2310 (approved):
Good work!!!

PR #1676, opened 2023-11-21
add task solution

Albert-Kifiuk:


src/figuresClasses.ts
```diff
@@ -1,19 +1,78 @@
 export interface Figure {
-
+  shape: string;
+  color: string;
+  getArea(): number;
 }
 
 export class Triangle implements Figure {
+  shape = 'triangle';
```
VitaliyBodnar, line 8: Please create enum for shape as it  will make a code cleaner

src/figuresClasses.ts
```diff
@@ -1,19 +1,78 @@
...
+    if (a <= 0 || b <= 0 || c <= 0 || a + b <= c || b + c <= a || c + a <= b) {
+      throw new Error('Invalid triangle sides');
+    }
 
+    this.color = color;
+    this.sides = [a, b, c];
+  }
+
+  getArea(): number {
+    const s = (this.sides[0] + this.sides[1] + this.sides[2]) / 2;
```
VitaliyBodnar, line 27: Let's make your code cleaner 
```suggestion
    const s = (this.a + this.b + this.c) / 2;
```

VitaliyBodnar (changes requested):

src/figuresClasses.ts
```diff
@@ -1,19 +1,79 @@
...
+  constructor(public color: string,
+    public width: number,
+    public height: number) {
+    if (width <= 0 || height <= 0) {
+      throw new Error('Invalid rectangle dimensions');
+    }
 
+    this.color = color;
+    this.width = width;
+    this.height = height;
```
etojeDenys, line 67: looks redundant, 'cause you're using here Typescript Constructor Shorthand

etojeDenys (changes requested):

src/figuresClasses.ts
```diff
@@ -1,19 +1,72 @@
...
+    return Math.floor(area * 100) / 100;
+  }
 }
 
 export class Rectangle implements Figure {
+  shape = Shape.Rectangle;
 
+  constructor(public color: string,
+    public width: number,
+    public height: number) {
```
VitaliyBodnar, line 57:
```suggestion
  constructor(
    public color: string,
    public width: number,
    public height: number
  ) {
```

VitaliyBodnar (approved):

PR #1684, opened 2023-11-23
Solution

zaura333:


src/figuresClasses.ts
```diff
@@ -1,19 +1,72 @@
-export interface Figure {
+function roundToHundreths(number: number): number {
+  return +(Math.floor(number * 100) / 100).toFixed(2);
```
Barosz30, line 2: return Math.floor(number * 100) / 100)

If you just leave it like that, it completes same task, without changing to string and back to number ;)

Barosz30 (approved):
Najsu!

mbruhler (approved):
Well done, approved

PR #1685, opened 2023-11-24
Solution

Barosz30:


DanyloSemenov (approved):
looks good!

src/figuresClasses.ts
```diff
@@ -1,19 +1,78 @@
...
+    public color: Color,
+    public radius: number,
+  ) {
+    if (radius <= 0) {
+      throw new Error('That isnt realy circle');
+    }
+  }
+
+  getArea():number {
+    const area = Math.PI * this.radius * this.radius;
```
mvjl000, line 48: This could be done with `Math.pow` or `**`
```suggestion
    const area = Math.PI * this.radius ** 2;
```

mvjl000 (approved):
Well done. I noticed one small thing to change but I don't think it's worth requesting changes ;)

PR #1686, opened 2023-11-24
Solution

patrykyo20:


src/figuresClasses.ts
```diff
@@ -1,19 +1,71 @@
 export interface Figure {
-
+  shape: 'triangle' | 'circle' | 'rectangle',
+  color : 'red' | 'green' | 'blue',
+  getArea: () => number;
 }
 
 export class Triangle implements Figure {
+  shape = 'triangle';
```
zaura333, line 8: If you have TypeScript error while using this notation, you can add type:
```suggestion
  shape: 'triangle' | 'circle' | 'rectangle' = 'triangle';
```

And maybe storing it in another custom type would make it cleaner, at least that was my solution :)

zaura333 (approved):
Nice, I added two suggestions

src/figuresClasses.ts
```diff
@@ -1,19 +1,71 @@
...
+  shape = 'rectangle';
+
+  constructor(public color : string,
+    public width: number, public height: number) {
+    this.color = color;
+    this.width = width;
+    this.height = height;
+
+    if (width <= 0 || height <= 0) {
+      throw new Error('Invalid triangle sides');
```
mbruhler, line 60: It's not a triangle

src/figuresClasses.ts
```diff
@@ -1,19 +1,71 @@
...
 export class Circle implements Figure {
+  shape = 'circle';
 
+  constructor(public color : string,
+    public radius: number) {
+    this.color = color;
+    this.radius = radius;
+
+    if (radius <= 0) {
+      throw new Error('Invalid triangle sides');
```
mbruhler, line 41: It's not a triangle

src/figuresClasses.ts
```diff
@@ -1,19 +1,71 @@
...
 
 export class Triangle implements Figure {
+  shape = 'triangle';
+
+  constructor(public color : string,
+    public a: number, public b: number, public c: number) {
+    this.color = color;
+    this.a = a;
+    this.b = b;
+    this.c = c;
```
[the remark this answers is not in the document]
mbruhler: that's true

src/figuresClasses.ts
mbruhler: Yes, it's best to use another type here
```
export type Shape = 'triangle' | 'circle' | 'rectangle'
```
then in class use the type in the parameter
```
  shape: Shape = 'triangle';
}
```

Of course you can do `Figure['shape']` as type annotation but it's ugly, it's better to export another type.

mbruhler: And in `Figure` use 
```
export interface Figure {
  shape: Shape,
  color : 'red' | 'green' | 'blue', //color can be also exported as another type
  getArea: () => number;
}
```
Rule of thumb is to use as much types as you can in my opinion so don't hestitate to split it further

mbruhler (changes requested):
Changes are needed

PR #1693, opened 2023-11-27
solution

aureliagorzna:


src/figuresClasses.ts
```diff
@@ -1,19 +1,95 @@
...
+    this.color = color;
+    this.radius = radius;
+  }
+
+  getArea(): number {
+    const result: number = parseFloat(
+      (Math.PI * this.radius * this.radius).toFixed(2),
+    );
+
+    if (result === 113.1) {
```
mbruhler, line 60: that's a very bad thing to do, you are using toFixed which rounds the value. Use what @MarcinLigmann suggested

```
const result: number = (this.radius * this.radius * Math.PI);
return Math.floor(result * 100) / 100;
```

mbruhler (changes requested):
Hi,
Your code looks good, but you use a tests hack, I provided you with solution so you won't have to to pass the tests.

mbruhler (approved):

PR #1695, opened 2023-11-27
Solution

MarcinLigmann:


src/figuresClasses.ts
```diff
@@ -1,19 +1,94 @@
 export interface Figure {
-
+  shape: string;
+  color: string;
+  getArea(): number;
```
kate-prakofyeva, line 4: type Shapes = 'triangle' | 'circle' | 'rectangle';
shape: Shapes;
color: Colors;

if you define enumerations, you won’t have to give types every time in Class, only change a shape type

mbruhler (approved):
Approved

PR #1698, opened 2023-11-27
task by spy

kalkvova:


src/figuresClasses.ts
```diff
@@ -1,19 +1,80 @@
-export interface Figure {
+type Shape = 'triangle' | 'circle' | 'rectangle';
```
maxim2310, line 1: better to use enum here instead union type

maxim2310 (changes requested):

lerastarynets (approved):
You've got this, keep pushing forward!

PR #1699, opened 2023-11-27
solution

RadoslawBielski:


mvjl000 (changes requested):
You need to change the way you round the returned values in order to pass the tests

src/figuresClasses.ts
```diff
@@ -1,19 +1,87 @@
...
 export class Triangle implements Figure {
+  shape: string = 'triangle';
+
+  color: string;
+
+  a: number;
+
+  b: number;
+
+  c: number;
```
JoyCoffeeAddict, line 16: delete the empty spaces between the props

src/figuresClasses.ts
```diff
@@ -1,19 +1,87 @@
 export interface Figure {
-
+  shape: string;
```
JoyCoffeeAddict, line 2: define shape as `type alias` or `enum`
the same for the `color`

src/figuresClasses.ts
```diff
@@ -1,19 +1,87 @@
...
+    this.c = c;
+  }
+
+  getArea(): number {
+    const semiPerimeter = (this.a + this.b + this.c) / 2;
+
+    const area = Math.sqrt(semiPerimeter * (semiPerimeter - this.a)
+     * (semiPerimeter - this.b) * (semiPerimeter - this.c));
+
+    return parseFloat(area.toFixed(2));
```
JoyCoffeeAddict, line 34: rounding here, Miłosz mentioned that.

JoyCoffeeAddict (changes requested):

mbruhler (approved):
Approved

PR #1709, opened 2023-11-29
solution

OstinM:


yefimov77 (approved):
Good Job! 🚀

src/figuresClasses.ts
```diff
@@ -1,19 +1,80 @@
-export interface Figure {
+type Shape = 'triangle' | 'circle' | 'rectangle';
+type Color = 'red' | 'green' | 'blue';
 
+export interface Figure {
+  shape: Shape,
+  color: Color,
+  getArea(): number,
```
olya-shyian, line 7: please correct the syntax, commas should be replaced by semicolons

src/figuresClasses.ts
```diff
@@ -1,19 +1,80 @@
...
 }
 
 export class Triangle implements Figure {
+  shape: Shape;
```
olya-shyian, line 11: instead of writing this.shape = 'triangle' you can indicate which shape is used here :
shape: Shape = 'triangle';

do the same with 'circle' and 'rectangle'

src/figuresClasses.ts
```diff
@@ -1,19 +1,80 @@
...
+    if (a + b <= c || a + c <= b || b + c <= a) {
+      throw new Error(`sides ${a}, ${b} and ${c} can't form a triangle`);
+    }
+
+    this.shape = 'triangle';
+  }
+
+  getArea(): number {
+    const { a, b, c } = this;
+    const p = (a + b + c) / 2;
```
olya-shyian, line 32: what is 'p'? please update the name

src/figuresClasses.ts
```diff
@@ -1,19 +1,80 @@
...
+      = Math.floor(Math.sqrt(p * (p - a) * (p - b) * (p - c)) * 100) / 100;
 
+    return area;
+  }
 }
 
 export class Circle implements Figure {
+  shape: Shape;
 
+  constructor(public color: Color, public radius: number) {
```
olya-shyian, line 43: for better readability move parameters to new lines

src/figuresClasses.ts
```diff
@@ -1,19 +1,80 @@
...
 export class Rectangle implements Figure {
+  shape: Shape;
+
+  constructor(
+    public color: Color,
+    public weidth: number,
+    public height: number,
+  ) {
+    if (weidth < 1 || height < 1) {
+      throw new Error('must be a size');
```
olya-shyian, line 65: this message is not very clear... it would be better to write something like this:
'All side lengths must be more than 0'

Check all your messages, maybe some texts can be improved

olya-shyian (changes requested):
Nice work! let's make some small changes)

PR #1711, opened 2023-11-29
ts_figures_classes

mberezhna:


emilionila (approved):
Well done!

src/figuresClasses.ts
```diff
@@ -1,19 +1,79 @@
...
+    if (a <= 0 || b <= 0 || c <= 0 || a + b <= c || b + c <= a || a + c <= b) {
+      throw new Error('Invalid triangle sides');
+    }
 
+    this.color = color;
+  }
+
+  getArea(): number {
+    const s = (this.a + this.b + this.c) / 2;
+    const area = (Math.sqrt(s * (s - this.a) * (s - this.b) * (s - this.c)))
```
BudnikOleksii, line 30: - Avoid one-letter names for variables.
- you get `this.a` and others twice, you can use destructing to avoid repeating `const { a, b, c } = this;

src/figuresClasses.ts
```diff
@@ -1,19 +1,79 @@
...
+    color: Color,
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    if (a <= 0 || b <= 0 || c <= 0 || a + b <= c || b + c <= a || a + c <= b) {
+      throw new Error('Invalid triangle sides');
+    }
 
+    this.color = color;
```
BudnikOleksii, line 25: Checklist:
- When you declare a constructor parameter with an access modifier TypeScript implicitly creates a property on the class with the same name as the parameter and assigns the parameter value to that property. You don't need to do it manually.
image

BudnikOleksii (changes requested):
Good work 👍 
Left a few comments to improve your solution

BudnikOleksii (approved):
Good work 👍

PR #1714, opened 2023-11-29
done task Figures

Anton-Kuchmasov:


BoblykNazar (approved):
Good job!

src/figuresClasses.ts
```diff
@@ -1,19 +1,75 @@
...
+    readonly color: Color,
+    readonly a: number,
+    readonly b: number,
+    readonly c: number,
+  ) {
+    if (a < 1 || b < 1 || c < 1) {
+      throw new Error('Bad input data!');
+    }
+
+    if (a + b <= c || a + c <= b || b + c <= a) {
```
ArtemTeslenko, line 24: try to avoid such long conditions inside if, you can create variable for it, for example: `isInvalidInput = a<1 || b<1 || c<1` and i think that we should have ability to get number with decimals, therefore `a<=0`, `b<=0`, `c<=0`
`a + b <= c || a + c <= b || b + c <= a` -> here you can get sum of all and subtract a max value, but optional for you, and put condition into variable

src/figuresClasses.ts
```diff
@@ -1,19 +1,75 @@
...
 }
 
 export class Circle implements Figure {
+  readonly shape = 'circle';
+
+  constructor(
+    readonly color: Color,
+    readonly radius: number,
+  ) {
+    if (radius < 0) {
```
ArtemTeslenko, line 45: if radius === 0? use `<=`

src/figuresClasses.ts
```diff
@@ -1,19 +1,75 @@
...
 
 export class Rectangle implements Figure {
+  readonly shape = 'rectangle';
+
+  constructor(
+    readonly color: Color,
+    readonly a: number,
+    readonly b: number,
+  ) {
+    if (a < 1 || b < 1) {
```
ArtemTeslenko, line 63: the same as in triangle, what if a or b will equal to 0.5?

ArtemTeslenko (approved):
Nice 👍 I am approving it, but see comments please)

PR #1718, opened 2023-11-29
add task solution

ViktorAfk:


src/figuresClasses.ts
```diff
@@ -1,19 +1,79 @@
...
+    public b:number,
+    public c:number,
+  ) {
+    this.shape = 'triangle';
+
+    if (this.a <= 0 || this.b <= 0 || this.c <= 0) {
+      throw new Error('You can\'t make a triangle with negative number(s)');
+    }
+
+    if ((this.a + this.b) <= this.c) {
```
ksyvvkaa, line 23: I think any side can be the longest, not only this.c, so you need to check all possiple cases

ksyvvkaa (approved):
Good job! 👍

src/figuresClasses.ts
```diff
@@ -1,19 +1,79 @@
+type Color = 'red' | 'blue' | 'green';
 export interface Figure {
-
+  shape: 'circle'| 'triangle' | 'rectangle';
```
ArtemTeslenko, line 3: you can set type as you did with color

src/figuresClasses.ts
ArtemTeslenko: right remark

src/figuresClasses.ts
```diff
@@ -1,19 +1,79 @@
...
+
+  constructor(
+    public color:Color,
+    public a:number,
+    public b:number,
+    public c:number,
+  ) {
+    this.shape = 'triangle';
+
+    if (this.a <= 0 || this.b <= 0 || this.c <= 0) {
```
ArtemTeslenko, line 19: move such long conditions into variable, it will improve readability

src/figuresClasses.ts
```diff
@@ -1,19 +1,79 @@
...
+    }
+
+    if ((this.a + this.b) <= this.c) {
+      throw new Error(`Sides ${a},
+       ${b} and ${c} can't form a triangle`);
+    }
+  }
+
+  public getArea():number {
+    const p:number = 0.5 * (this.a + this.b + this.c);
```
ArtemTeslenko, line 30: p - is not meaningful: `semiperimeter`, `halfPerimeter` etc.

src/figuresClasses.ts
```diff
@@ -1,19 +1,79 @@
...
+  constructor(public color:Color, public radius:number) {
+    this.shape = 'circle';
+
+    if (this.radius <= 0) {
+      throw new Error(`You can't make a circkle with this radius:${radius}`);
+    }
+  }
+
+  public getArea():number {
+    const pi:number = Math.PI;
```
ArtemTeslenko, line 50: no need to specify type here, ts will understand that it is a number

ArtemTeslenko (changes requested):
Good job 👍 See comments)

BudnikOleksii (approved):
Good work 👍

PR #1724, opened 2023-11-30
solution

YehorHeneralov:


src/figuresClasses.ts
```diff
@@ -1,19 +1,81 @@
...
+  constructor(
+    public color: Color,
+    private a: number,
+    private b: number,
+    private c: number,
+  ) {
+    if (a <= 0 || b <= 0 || c <= 0 || a + b <= c || b + c <= a || c + a <= b) {
+      throw new Error('side is not correct');
+    }
+  }
```
ViktorAfk, line 28: Better to make different conditions and create some variables for that, because it is not good for readability.

src/figuresClasses.ts
```diff
@@ -1,19 +1,81 @@
...
 
+  getArea(): number {
+    const { a, b, c } = this;
+    const p = (a + b + c) * (1 / 2);
```
ViktorAfk, line 32: Rename 'p' because, it's not understandable what is 'p'. Better to call it semiperimitr or halfperemitr. I had the same mistake.

ViktorAfk (approved):
Good job! Fix minor mistakes.

BudnikOleksii (approved):
Good work 👍

PR #1725, opened 2023-12-01
Solution

YuriySobakar:


olegnovikov10 (approved):
good joob

src/figuresClasses.ts
```diff
@@ -1,19 +1,104 @@
...
+    const arePositive: boolean = sides[0] > 0 && sides[1] > 0 && sides[2] > 0;
+    const isTriangle: boolean = longestSide < otherSides[0] + otherSides[1];
+
+    if (!arePositive || !isTriangle) {
+      throw new Error('Not a triangle');
+    }
+
+    this.a = a;
+    this.b = b;
+    this.c = c;
```
BudnikOleksii, line 41: [CODE KNOWLEDGE] - When you declare a constructor parameter with an access modifier TypeScript implicitly creates a property on the class with the same name as the parameter and assigns the parameter value to that property. You don't need to do it manually.
image

The same as for `this.color = color;`

src/figuresClasses.ts
```diff
@@ -1,19 +1,104 @@
...
+  }
 
+  getArea(): number {
+    const p = (this.a + this.b + this.c) / 2;
+    const area = Math.sqrt(p * (p - this.a) * (p - this.b) * (p - this.c));
```
BudnikOleksii, line 46: - Avoid one-letter names such as `p`.
- You do `this.a` and other twice, you can use destructing. `const { a, b, c } = this;`

src/figuresClasses.ts
```diff
@@ -1,19 +1,104 @@
...
+  radius: number;
+
+  shape: Shape = 'circle';
+
+  constructor(
+    public color: Color,
+    radius: number,
+  ) {
+    this.color = color;
+    this.radius = radius;
```
BudnikOleksii, line 62: The same as I mentioned previously, fix in all places

BudnikOleksii (changes requested):
Good job 👍 
Just a few fixes are required, take a look on my suggestions

src/figuresClasses.ts
```diff
@@ -1,19 +1,92 @@
...
+  getArea(): number {
+    const { aSide, bSide, cSide } = this;
+    const halfPerimetr = (aSide + bSide + cSide) / 2;
+    const area = Math.sqrt(halfPerimetr
+      * (halfPerimetr - aSide)
+      * (halfPerimetr - bSide)
+      * (halfPerimetr - cSide));
+
+    return +area.toFixed(2);
+  }
```
roman-mirzoian, line 41: Good job!

PR #1735, opened 2023-12-05
solution 1

mickhakk:


src/figuresClasses.ts
```diff
@@ -1,19 +1,70 @@
 export interface Figure {
-
+  shape: string;
+  color: string;
```
roman-mirzoian, line 3: According to the task, we have defined certain types of shapes that can be, and by specifying a general type of `string`, I can write anything there and it will work. The same for color:
image

src/figuresClasses.ts
```diff
@@ -1,19 +1,70 @@
...
+
+  color: string;
+
+  constructor(
+    color: string,
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    if (a <= 0 || b <= 0 || c <= 0 || a + b <= c || a + c <= b || b + c <= a) {
```
roman-mirzoian, line 18: The condition is very difficult to read and understand, it is better to break it into several parts

src/figuresClasses.ts
```diff
@@ -1,19 +1,70 @@
...
+      throw new Error('Invalid triangle sides');
+    }
+    this.color = color;
+  }
 
+  getArea(): number {
+    const s = (this.a + this.b + this.c) / 2;
```
roman-mirzoian, line 25: Destructuring can be used for `a`, `b`, `c`, and the variable name should not be abbreviated to unreadable

BudnikOleksii (approved):
Good work 👍

PR #1742, opened 2023-12-11
Done

MarkiMark3:


src/figuresClasses.ts
```diff
@@ -1,19 +1,96 @@
 export interface Figure {
-
+  shape: string;
+  color: string;
```
Esceype, line 3: Create a union type for shape and color because you have a limited list of possible properties

src/figuresClasses.ts
```diff
@@ -1,19 +1,96 @@
...
+  constructor(color: string, width: number, height: number) {
+    if (width <= 0 || height <= 0) {
+      throw new Error('Width and height must be greater than 0.');
+    }
+
+    this.shape = 'rectangle';
+    this.color = color;
+    this.width = width;
+    this.height = height;
+  }
```
Esceype, line 85:
```suggestion
 shape: Shape = 'rectangle';

  constructor(public color: Color, private width: number, private height: number) {
    if (width <= 0 || height <= 0) {
      throw new Error('Width and height must be greater than 0');
    }
  }
```

As you remember you can set variables for constructor like this, this looks less complicated, do the same of all the similar cases above

Esceype (changes requested):
GJ!
Read the following comments.

VitaliyBodnar (approved):

PR #1745, opened 2023-12-12
Solution

Donovanoff:


src/figuresClasses.ts
```diff
@@ -1,19 +1,74 @@
...
+      throw new Error('Each side length must be greater than zero.');
+    }
+
+    if (a >= b + c || b >= a + c || c >= a + b) {
+      throw new Error('Wrong length sides of triangle');
+    }
+  }
 
+  getArea(): number {
+    const s = (this.a + this.b + this.c) / 2;
```
VitaliyBodnar, line 29: you can destructure this
it will make you code shorter 
```suggestion
    const { a, b, c } = this;
    const s = (a + b + c) / 2;
```

VitaliyBodnar (changes requested):

OleziO (approved):

PR #1749, opened 2023-12-14
Develop

VladimirKolesnikov:


src/figuresClasses.ts
```diff
@@ -1,19 +1,105 @@
...
+
+  constructor(
+    color: Color,
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    checkCorrectLengthsOfTriangleSides(a, b, c);
+    this.color = color;
+    this.shape = 'triangle';
```
etojeDenys, line 52:
```suggestion
  shape: Shape;

  constructor(
    public color: Color,
    public a: number,
    public b: number,
    public c: number,
  ) {
    checkCorrectLengthsOfTriangleSides(a, b, c);

    this.shape = 'triangle';
```

etojeDenys: https://dev.to/satansdeer/typescript-constructor-shorthand-3ibd

etojeDenys (changes requested):
do the same with color property everywhere

src/figuresClasses.ts
VladimirKolesnikov: autor's nickname is "satansdeer" )))))

volodymyr-soltys97 (approved):
Great work 🔥

PR #1757, opened 2023-12-17
solution

MykhailoLoniak:


src/figuresClasses.ts
```diff
@@ -1,19 +1,83 @@
...
+  shape: 'triangle' | 'circle' | 'rectangle';
+  color: 'red' | 'green' | 'blue';
+  getArea(): number;
 }
 
 export class Triangle implements Figure {
+  public shape: 'triangle' = 'triangle';
+
+  constructor(
+    public color: 'red' | 'green' | 'blue',
```
anastasiiavorobiova, line 11:
```suggestion
    public color: Color,
```
Consider creating the type for the color instead of duplicating values further in the code

src/figuresClasses.ts
```diff
@@ -1,19 +1,83 @@
 export interface Figure {
-
+  shape: 'triangle' | 'circle' | 'rectangle';
+  color: 'red' | 'green' | 'blue';
+  getArea(): number;
 }
 
 export class Triangle implements Figure {
+  public shape: 'triangle' = 'triangle';
```
anastasiiavorobiova, line 8: Consider creating an enum for the shape 
```suggestion
  public shape: Shape = Shape.triangle;
```

anastasiiavorobiova (changes requested):
Great work! Almost done!

anastasiiavorobiova (approved):
Awesome work!

PR #1771, opened 2023-12-26
task solution

TenetaJ:


src/figuresClasses.ts
```diff
@@ -1,19 +1,81 @@
...
+    if (radius < 0) {
+      throw new Error('the radius of the circle must be greater than zero');
+    }
+    this.shape = 'circle';
+    this.color = color;
+    this.radius = radius;
+  }
+
+  getArea(): number {
+    return Math.PI * (this.radius ** 2);
```
witflash, line 50: Here you have to use Math.floor to round the final value to 2 digits after point.

witflash (changes requested):
Pay attention to my comments, please. Try to fix these problems and pass all the tests

src/figuresClasses.ts
```diff
@@ -1,19 +1,81 @@
...
 export class Triangle implements Figure {
+  shape: string;
+
+  constructor(
+    public color: string,
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    if (a < 0 || b < 0 || c < 0) {
```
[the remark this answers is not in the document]
TenetaJ: done

src/figuresClasses.ts
TenetaJ: done

src/figuresClasses.ts
```diff
@@ -1,19 +1,84 @@
...
+
+  constructor(
+    public color: string,
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    if (a <= 0 || b <= 0 || c <= 0 || a + b <= c || a + c <= b || b + c <= a) {
+      throw new Error('Error');
+    }
```
SanyaBratashchuk, line 18:
```suggestion
  const sides = [a, b, c];
  const perimeter = sides.reduce((sum, side) => sum + side);
  const biggestSide = Math.max(...sides);
  const hasZeroLengthSide = sides.some(side => side <= 0);
  const hasTooLongSide = perimeter - biggestSide <= biggestSide;
    if (hasZeroLengthSide || hasTooLongSide) {
      throw new Error('Error');
    }
```

SanyaBratashchuk (approved):

PR #1772, opened 2023-12-27
add solution

ptbit:


src/figuresClasses.ts
```diff
@@ -1,19 +1,71 @@
 export interface Figure {
-
+  shape: string;
```
anastasiiavorobiova, line 2: Consider defining an enum for the shape and a type for the color and using them further in the code

anastasiiavorobiova (changes requested):
Great work! Almost done!

PR #1780, opened 2023-12-30
Implemented figure inteface and its implementations

DmytroV95:


sarakhmen (approved):
Good job! You can fix style/indentation issues if you want ;)

PR #1783, opened 2024-01-02
add task solution

heorhiig:


src/figuresClasses.ts
```diff
@@ -1,19 +1,79 @@
...
+    if (width <= 0 || height <= 0) {
+      throw new Error('error');
+    }
+    this.color = color;
+    this.width = width;
+    this.height = height;
+  }
+
+  getArea(): number {
+    return this.width * this.height;
```
l4st1m0za, line 71: The area of the rectangle should be rounded down to hundredths according to the task. You should use Math.floor() function to achieve this.

l4st1m0za (changes requested):
Great job, but one moment to fix

anastasiia-tilikina (changes requested):
seems like you have errors in your tests. pls fix them.

in order to do that you need first calculate area and then return rounded value:

```typescript
const area = //...

return Math.floor(area //...
```

Viktor-Kost (approved):

PR #1792, opened 2024-01-10
ts-figures

BodyaRespect:
Can you show me how to implement function `checkSide` into class Triangle? And if only i must do it.


lerastarynets (approved):
lgtm 🚀

PR #1795, opened 2024-01-10
Develop

sasholk:


src/figuresClasses.ts
```diff
@@ -1,19 +1,76 @@
...
+    public color: string,
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    if (this.a <= 0 || this.b <= 0 || this.c <= 0) {
+      throw new Error('Resulting value should be bigger than 0');
+    }
+
+    if (this.a + this.b <= this.c) {
```
ssaviak, line 20: There is a possible mistake. How do you know that hypotenuse always will be the last parameter? Recommend using Math.max to find the longest side.

ssaviak (approved):
well done 🫡

src/figuresClasses.ts
sasholk: If I correctly remmember in algebra the third side by name `c` its allways hypotenuse, so it's allways has longest side, so it wasn't misstake.

src/figuresClasses.ts
```diff
@@ -1,19 +1,76 @@
...
 export class Triangle implements Figure {
+  shape = 'triangle';
+
+  constructor(
+    public color: string,
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    if (this.a <= 0 || this.b <= 0 || this.c <= 0) {
```
DanilWeda, line 16: as alternative*
```suggestion
    if ([this.a, this.b, this.b].every( side => side > 0) {
```

DanilWeda (approved):
Great!
Approved.

src/figuresClasses.ts
sasholk: wow, didn't know about that... thanks 😊

PR #1798, opened 2024-01-11
add solution

Xilston:


src/figuresClasses.ts
```diff
@@ -1,19 +1,79 @@
...
 export class Triangle implements Figure {
+  shape: Shape = 'triangle';
+
+  constructor(
+    public color:Color,
+    public a:number,
+    public b:number,
+    public c:number,
+  ) {
+    if (a <= 0 || b <= 0 || c <= 0) {
```
DanilWeda, line 19: as alternative*
```suggestion
    if ([a, b, c].every( side => side > 0) {
```

DanilWeda (approved):
Great! 
Approved!
Thanks!

PR #1799, opened 2024-01-11
add solution

ValeriaMyronets:


src/figuresClasses.ts
```diff
@@ -1,19 +1,72 @@
 export interface Figure {
-
+  shape: string;
+  color: string;
```
nabokok, line 3: I think shape and color should have string literal types. We use only 3 colors and shapes

src/figuresClasses.ts
```diff
@@ -1,19 +1,72 @@
...
+    public color: string,
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    this.color = color;
+
+    if (a + b <= c) {
+      throw new Error();
+    }
```
nabokok, line 20: this check will work only if the biggest side is passed as the last argument

src/figuresClasses.ts
```diff
@@ -1,19 +1,72 @@
...
+  ) {
+    this.color = color;
+
+    if (width <= 0 || height <= 0) {
+      throw new Error();
+    }
+  }
 
+  getArea(): number {
+    return Math.round((this.width * this.height) * 100) / 100;
```
nabokok, line 66: why do you use different methods(Math.floor, Math.round) in getArea for different figures?

nabokok (changes requested):
Good job! Look at my comments below

DanilWeda (approved):
Great! 
Approved.

PR #1801, opened 2024-01-11
add solution

nabokok:


src/figuresClasses.ts
```diff
@@ -1,19 +1,76 @@
 export interface Figure {
-
+  shape: 'triangle' | 'circle' | 'rectangle';
+  color: 'red' | 'blue' | 'green';
```
artem8746, line 3: you can create type for `shape` and `color`

`type Shape = 'triangle' | 'circle' | 'rectangle'` and just use everywhere `Shape` instead of `Figure['shape']`

src/figuresClasses.ts
```diff
@@ -1,19 +1,76 @@
...
+  constructor(
+    public color: Figure['color'],
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    const biggestSide = Math.max(a, b, c);
+    const isIncorrectSides = biggestSide >= a + b + c - biggestSide;
+
+    if (a <= 0 || b <= 0 || c <= 0 || isIncorrectSides) {
```
artem8746, line 19: you can create function to check if each side greater than 0 and then just use it in other functions

artem8746 (approved):
Excellent job!!
Check a few comments how to improve code

DanilWeda (approved):
Great! 
Approved.

PR #1803, opened 2024-01-12
add solution

MaxKalalashnyk02:


src/figuresClasses.ts
```diff
@@ -1,19 +1,90 @@
...
+  color: string;
+  getArea(): number;
 }
 
 export class Triangle implements Figure {
+  shape = 'triangle';
+
+  color: string;
+
+  sides: number[];
```
kostrubitskii, line 12: I think better write this without spaces
```suggestion
  shape = 'triangle';
  color: string;
  sides: number[];
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,90 @@
...
+
+  constructor(color: string, a: number, b: number, c: number) {
```
kostrubitskii, line 14: You can rewrite your constructor to make it easier to read
```suggestion
  constructor(
    color: string, 
    a: number, 
    b: number, 
    c: number,
  ) {
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,90 @@
...
+    if (a >= b + c || b >= a + c || c >= a + b) {
+      throw new Error('Invalid triangle sides');
+    }
 
+    this.color = color;
+    this.sides = [a, b, c];
+  }
+
+  getArea(): number {
+    const s = (this.sides[0] + this.sides[1] + this.sides[2]) / 2;
```
kostrubitskii, line 28: What is "s"?
Better name it "Square" or "Area"

src/figuresClasses.ts
```diff
@@ -1,19 +1,90 @@
...
+    return Math.floor(area * 100) / 100;
+  }
 }
 
 export class Circle implements Figure {
+  shape = 'circle';
+
+  color: string;
+
+  radius: number;
```
kostrubitskii, line 42: Here you need to fix spaces too

kostrubitskii (approved):
Good job! Fix some comments

src/figuresClasses.ts
DanilWeda, line 28: Fix naming

DanilWeda (changes requested):
Good job! 
Please, fix the naming and recommended linting comments. 
Thanks!

PR #1804, opened 2024-01-12
solution

kostrubitskii:


src/figuresClasses.ts
```diff
@@ -1,19 +1,76 @@
...
+    public c: number,
+  ) {
+    if ([a, b, c].every((side) => side <= 0)) {
+      throw new Error('Sides must be bigger then 0');
+    }
+
+    if (a + b <= c || a + b <= b || b + c <= a) {
+      throw new Error('The sides provided do not form a valid triangle');
+    }
+  }
```
vpashko, line 23: you can sort you array and add only 1 check
```ts
const sides = [];

    sides.push(this.a, this.b, this.c);

    sides.sort((side1, side2) => side2 - side1);

    if (sides[0] >= sides[1] + sides[2]) {
      throw new Error(
        'error message',
      );
    }
```
but yout code shorter, so it`s ok

src/figuresClasses.ts
```diff
@@ -1,19 +1,76 @@
...
 
 export class Rectangle implements Figure {
+  shape = 'rectangle';
+
+  constructor(
+    public color: string,
+    public width: number,
+    public height: number,
+  ) {
+    if (width <= 0 || height <= 0) {
```
vpashko, line 62:
```suggestion
    if ([width, height].some((x) => x <= 0)) {
```

the same

vpashko (approved):
Great job, keep it up! &#128077;

PR #1805, opened 2024-01-12
calc area

vpashko:


src/figuresClasses.ts
```diff
@@ -1,19 +1,99 @@
 export interface Figure {
-
+  shape: string,
+  color: string,
```
vSainiuk, line 3: I think it had better to add also literal types for these properties.

vSainiuk (approved):

vSainiuk:
Well done!

src/figuresClasses.ts
```diff
@@ -1,19 +1,99 @@
 export interface Figure {
-
+  shape: string,
```
DanilWeda, line 2: Add the color and shape types that you use in the task.

DanilWeda (changes requested):
Good job! 
Check small type comments! 
Thanks!

DanilWeda (changes requested):
comment for Request changes*

DanilWeda (approved):
Good job! 
Approved! 
Many thanks.

PR #1808, opened 2024-01-13
Solution

Yur1yBubl1k:


rtkachuk-moon (approved):
👍

src/figuresClasses.ts
```diff
@@ -1,19 +1,78 @@
...
 export class Triangle implements Figure {
+  shape: string = 'triangle';
 
+  constructor(
+    public color: string,
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    if (a <= 0 || b <= 0 || c <= 0) {
```
DanilWeda, line 17: as alternative
```suggestion
    if ([a, b, c].every(side => side > 0)) {
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,78 @@
 export interface Figure {
+  shape: string;
+  color: string;
```
DanilWeda, line 3: Describe all shapes and colors used in this task.

DanilWeda (changes requested):
Good job!
Please, check minor type comments!
Many thanks!

src/figuresClasses.ts
Yur1yBubl1k: why?

olya-shyian (changes requested):
Good work!🌼

PR #1811, opened 2024-01-14
add task solution

avakiel:


src/figuresClasses.ts
```diff
@@ -1,19 +1,91 @@
...
+}
 
+export interface Figure {
+  shape: Shapes;
+  color: string;
+  getArea: () => number;
 }
 
 export class Triangle implements Figure {
+  public shape: Shapes;
```
YevhenOstrovskiy, line 14: shape = Shapes.Triangle;
  You can use your enum in that way!)

YevhenOstrovskiy (approved):
Good job!!! Check the comment.

src/figuresClasses.ts
```diff
@@ -1,19 +1,91 @@
-export interface Figure {
+enum Shapes {
+  Triangle = 'triangle',
+  Circle = 'circle',
+  Rectangle = 'rectangle',
+}
 
+export interface Figure {
+  shape: Shapes;
+  color: string;
```
DanilWeda, line 9: You can describe 3 colors which use in task ( not critical )

src/figuresClasses.ts
DanilWeda: correct

DanilWeda (approved):
Good job! 
Approved with small comments. 
Many thanks!

PR #1818, opened 2024-01-15
Solution

yepolotn1ak:


src/figuresClasses.ts
```diff
@@ -1,19 +1,69 @@
 export interface Figure {
-
+  shape: 'triangle' | 'circle' | 'rectangle';
```
anastasiiavorobiova, line 2: Consider declaring a separate type or an enum to avoid repeating code

src/figuresClasses.ts
```diff
@@ -1,19 +1,69 @@
...
+  color: 'red' | 'green' | 'blue';
```
anastasiiavorobiova, line 3: The same issue here

anastasiiavorobiova (changes requested):
Great work! Almost done!

anastasiia-tilikina (approved):
Well done!

PR #1820, opened 2024-01-15
task solution

BeginerAlex:


src/figuresClasses.ts
```diff
@@ -1,19 +1,80 @@
 export interface Figure {
+  shape: string;
+  color: string;
 
+  getArea(): number;
 }
 
 export class Triangle implements Figure {
+  shape: string = 'triangle';
```
anastasiia-tilikina, line 9: let's create a type for shape - either union or enum and use it instead of string to restrict expected values.

anastasiia-tilikina (changes requested):
Wery well done!
Just a small suggestion)

src/figuresClasses.ts
BeginerAlex: OK, thanks for your tip)

anastasiia-tilikina (approved):
Great!

PR #1821, opened 2024-01-15
solution

Dmytrosp:


src/figuresClasses.ts
```diff
@@ -1,19 +1,72 @@
...
-
+  shape: string;
+  color: string;
+  getArea: () => number;
 }
 
 export class Triangle implements Figure {
+  shape = 'triangle';
+
+  color: string;
```
etojeDenys, line 10: looks redundant

src/figuresClasses.ts
```diff
@@ -1,19 +1,72 @@
...
+  constructor(
+    color: string,
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    if (a <= 0 || b <= 0 || c <= 0 || a + b <= c || a + c <= b || b + c <= a) {
+      throw new Error('Invalid triangle sides');
+    }
+    this.color = color;
```
etojeDenys, line 21:
```suggestion
  constructor(
    public color: string,
    public a: number,
    public b: number,
    public c: number,
  ) {
    if (a <= 0 || b <= 0 || c <= 0 || a + b <= c || a + c <= b || b + c <= a) {
      throw new Error('Invalid triangle sides');
    }
```
looks like you can simplify it

etojeDenys (changes requested):

etojeDenys (changes requested):

src/figuresClasses.ts
```diff
@@ -1,19 +1,72 @@
...
+
+  constructor(
+    public color: string,
+    public width: number,
+    public height: number,
+  ) {
+    if (width <= 0 || height <= 0) {
+      throw new Error('Invalid rectangle sides');
+    }
+    this.color = color;
```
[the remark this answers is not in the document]
Dmytrosp: But Class "Rectangle" › should check whether all sides are positive numbers. Same with circle

lerastarynets (approved):
lgtm 🚀

PR #1827, opened 2024-01-17
add task solution

DaniilBarilotti:


src/figuresClasses.ts
```diff
@@ -1,19 +1,75 @@
...
+  getArea(): number;
 }
 
 export class Triangle implements Figure {
+  shape: string = 'triangle';
+
+  color: string;
+
+  constructor(
+    color: string,
```
etojeDenys, line 13:
```suggestion
  constructor(
   public color: string,
```
you can simplify it

etojeDenys (changes requested):

etojeDenys (approved):
lgtm

PR #1840, opened 2024-01-23
Finished the task

julwer1k:


src/figuresClasses.ts
```diff
@@ -1,19 +1,82 @@
 export interface Figure {
+  shape: string,
```

src/figuresClasses.ts
```diff
@@ -1,19 +1,82 @@
...
+  color: string,
```

anastasiiavorobiova (commented):
Great work! Almost done!

PR #1843, opened 2024-01-24
task solution

Diliradon:
I wonted use my function like that 
  // const roundedNumber = +figure.getArea().toFixed(2);
  // const roundedNumber = Math.floor(figure.getArea() * 100) / 100;

  return `A ${figure.color} ${figure.shape} - ${roundedNumber }`;
}      and don`t write Math.floor( ... * 100) / 100 for ever classes . but I could pass one test.

3d4d28ad-02df-4171-be44-46e57405a8b2


anastasiiavorobiova (approved):
Great work!

PR #1847, opened 2024-01-26
Solution

Krzychu92:


src/figuresClasses.ts
```diff
@@ -1,19 +1,70 @@
 export interface Figure {
-
+  shape: string;
+  color: string;
```

volodymyr-soltys97 (approved):
Well done 🚀

PR #1848, opened 2024-01-27
Solution

ivankovbohdan:


src/figuresClasses.ts
```diff
@@ -1,19 +1,64 @@
 export interface Figure {
-
+  shape: 'triangle' | 'circle' | 'rectangle';
+  color: 'red' | 'green' | 'blue' | string;
```
VitaliyBondarenko1982, line 3: make types or enum for shape and color and use it everywhere.
for example.
```
enum Shape {
   tirangle = 'triangle',
   circle = 'circle',
   rectangle = 'rectangle',
}
```

VitaliyBondarenko1982 (changes requested):
Good job.
Please check commment.

src/figuresClasses.ts
ivankovbohdan: Really good solution - I couldn't figure out how to write it correctly before!🙄 Thanks!!!🤝🔥

src/figuresClasses.ts
```diff
@@ -1,19 +1,81 @@
+enum Shape {
+  triangle = 'triangle',
+  circle = 'circle',
+  rectangle = 'rectangle',
+}
+
+enum Color {
+  triangle = 'triangle',
```
maxim2310, line 8: i think its not color)

maxim2310 (changes requested):
Almost done

etojeDenys (approved):
lgtm

PR #1857, opened 2024-01-31
add task solution

Roman-Bilichenko:


lerastarynets (approved):
You're doing amazing, keep it up!

PR #1865, opened 2024-02-03
Solution

RichardNixon17:


src/figuresClasses.ts
```diff
@@ -1,19 +1,70 @@
 export interface Figure {
-
+  shape: string,
```
etojeDenys, line 2:
```suggestion
  shape: Shape,
```
`type Shape = 'triangle' | 'circle' | 'rectangle'`

etojeDenys (approved):
good job

PR #1875, opened 2024-02-07
solution

robsza1122:


src/figuresClasses.ts
```diff
@@ -1,19 +1,79 @@
...
 export class Triangle implements Figure {
+  shape: Shape = 'triangle';
+
+  constructor(public color : Color, public a : number,
+    public b : number, public c : number) {
+    if (a <= 0 || b <= 0 || c <= 0) {
+      throw new Error('One side cannot be equal or longer than sum of both sides');
+    }
+
+    if (a >= b + c) {
```
yevhenii-pyl, line 21: Can't we combine these if statements into one? 

`a >= b + c || b >= a + c || c >= a + b`

PR #1881, opened 2024-02-12
add task solution

AgnieszkaWw:


yevhenii-pyl (approved):
Good job, nice and clean. 
However, you can make it even more readable by changing variable names: 
 `a` => `sideA`, `r` => `radius`, and so on, as you did it for Rectangle class with `width` and `height`.
It may become harder and harder to dig in through shorthands as the codebase grows.

PR #1893, opened 2024-02-15
solution

yurovych:


src/figuresClasses.ts
```diff
@@ -1,19 +1,100 @@
...
+
+  getArea(): number | string {
+    const result: number = (this.width * this.height);
+
+    if (result % 1 === 0) {
+      return result;
+    }
 
+    return Number((result.toString()
+      .slice(0, result.toString().indexOf('.') + 3)));
```
etojeDenys, line 94: Math.floor(num * 100) / 100

you can just round the number like that

etojeDenys (changes requested):

etojeDenys (approved):
awesome

PR #1900, opened 2024-02-21
add task solution

AnastasiaChaikovskaya:
add task solution for ts_figures_classes


src/figuresClasses.ts
```diff
@@ -1,19 +1,75 @@
 export interface Figure {
-
+  shape: string;
```
etojeDenys, line 2: better to add some specific type here

etojeDenys (approved):
good job

PR #1910, opened 2024-02-26
Add solution

Vasya564:


src/figuresClasses.ts
```diff
@@ -1,19 +1,69 @@
...
 export class Triangle implements Figure {
+  shape: Shape = 'triangle';
+
+  constructor(
+    public color: Color,
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    if (a <= 0 || b <= 0 || c <= 0 || a + b <= c || a + c <= b || b + c <= a) {
```
polosanya, line 19: I think it's better to divide it into a few checks and provide more informative error texts

polosanya (changes requested):
Almost done!

PR #1925, opened 2024-03-04
add task solution

pawelnowicki87:


src/figuresClasses.ts
```diff
@@ -1,19 +1,81 @@
...
 }
 
 export class Triangle implements Figure {
+  shape: string = 'triangle';
+
+  color: string;
+
+  constructor(
+    color: string,
+    public a: number,
```
yevhenii-pyl, line 14: Let's avoid calling variables like `a, b, c, st, pt, bd, smthng` => a name should be meaningful. In this case, let's call them `sideA`, etc

src/figuresClasses.ts
```diff
@@ -1,19 +1,81 @@
...
+    return Math.floor(area * 100) / 100;
+  }
 }
 
 export class Circle implements Figure {
+  shape: string = 'circle';
+
+  color: string;
+
+  constructor(color: string, public r: number) {
```
yevhenii-pyl, line 42: Why `color` is called `color` and `radius` is `r`? :)

yevhenii-pyl (changes requested):
Good job, but variable naming is also important.
Could you adjust it? :)

yevhenii-pyl (approved):
LGTM 🥇

PR #1931, opened 2024-03-06
Add solution

ViktoriiaRepo:


src/figuresClasses.ts
```diff
@@ -1,11 +1,68 @@
...
+    public c: number,
+  ) {
+    if (!a || !b || !c || a + b <= c) {
+      throw new Error();
+    }
+  }
+
+  getArea(): number {
+    const s = (this.a + this.b + this.c) / 2;
+    const S = Math.sqrt(s * (s - this.a) * (s - this.b) * (s - this.c));
```
BudnikOleksii, line 25: What is `s` and the big `S`?

src/figuresClasses.ts
```diff
@@ -1,11 +1,68 @@
...
+  shape: Shape = 'triangle';
+
+  constructor(
+    public color: Color,
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    if (!a || !b || !c || a + b <= c) {
+      throw new Error();
```
BudnikOleksii, line 19: - [CODE STYLE] - If you throw an error try to provide its detailed description in error message.
image
- And the condition is really messy, I have no idea what you check here

BudnikOleksii (changes requested):
Looks like your solution works, but let's do it more readable so other developers can understand what is happening

stassokolov1 (approved):
Great job, I'm approving it🔥

PR #1947, opened 2024-03-12
taskComplete

TideTwo:


TideTwo:
In the error log it says that prettier needs node v14
again works on my end
image

choeqq (approved):

PR #1951, opened 2024-03-13
Solution

Davwidos:


OlesPartyka (approved):

src/figuresClasses.ts
```diff
@@ -1,11 +1,98 @@
...
+
+  b: number;
+
+  c: number;
+
+  constructor(
+    public color: Color,
+    a: number,
+    b: number,
+    c: number,
```
nataliaklonowska, line 23: just add public to a, b, c in constructor as you did with color

src/figuresClasses.ts
```diff
@@ -1,11 +1,98 @@
...
+      throw new Error('All sides shuld be grater than 0');
+    }
+
+    if (a >= b + c || b >= a + c || c >= a + b) {
+      throw new Error(`sides ${a}, ${b}, ${c} can't form triangle`);
+    }
+
+    this.a = a;
+    this.b = b;
+    this.c = c;
```
nataliaklonowska, line 35: When you declare a constructor parameter with an access modifier TypeScript implicitly creates a property on the class with the same name as the parameter and assigns the parameter value to that property. You don't need to do it manually

nataliaklonowska (changes requested):

nataliaklonowska (approved):

PR #1956, opened 2024-03-13
develop

yaGodya:
# Figures
We have 3 types of figures: triangles, circles and rectangles. 

has:
- a method `getArea` that returns the area of the figure rounded down to 
hundredths.

In addition to a `color` constructors should accept required data:
- sides `a`, `b` and `c` for a triangle;
- a `radius` for a circle;
- a `width` and a `height` for a rectangle.

The constructors should `throw new Error('your error message')` if:
- any length is <= 0
- the longest side of a triangle is >= than a sum of two others

Hints:
- use `Math.PI` for calculating a circle square
- use [Heron's formula](https://en.wikipedia.org/wiki/Heron%27s_formula) for triangles

Example:
```typescript
new Rectangle('blue', 2, 0) // throws an error
new Triangle('red', 1, 2, 3) // throws an error: sides 1, 2 and 3 can't form a triangle
```

Also create a function `getInfo` that takes a figure and returns a string in the
next format:
```typescript
const redRectangle = new Rectangle('red', 3, 5);
getInfo(redRectangle) === 'A red rectangle - 15';

const greenCircle = new Circle('green', 1);
getInfo(greenCircle) === 'A green circle - 3.14';
```

### DIFF
--- a/src/figuresClasses.ts
+++ b/src/figuresClasses.ts
@@ -1,11 +1,77 @@
-export interface Figure {}
+export interface Figure {
+  shape: string;
+  color: string;
+  getArea(): number;
+}
 
-export class Triangle implements Figure {}
+export class Triangle implements Figure {
+  shape: string = 'triangle';
 
-export class Circle implements Figure {}
+  constructor(
+    public color: string,
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    const NotPositive = a <= 0 || b <= 0 || c <= 0;
+    const isNotTriangle = a + b <= c || a + c <= b || b + c <= a;
 
-export class Rectangle implements Figure {}
+    if (NotPositive || isNotTriangle) {
+      throw new Error('incorrect value');
+    }
+    this.color = color;
+  }
 
-export function getInfo(figure): string {
-  return typeof figure;
+  getArea(): number {
+    const s = (this.a + this.b + this.c) / 2;
+    const area = Math.sqrt(s * (s - this.a) * (s - this.b) * (s - this.c));
+
+    return Math.floor(area * 100) / 100;
+  }
+}
+
+
+export class Circle implements Figure {
+  public shape = 'circle';
+
+  constructor(
+    public color: string,
+    public radius: number,
+  ) {
+    if (radius <= 0) {
+      throw new Error('incorrect value');
+    }
+    this.color = color;
+  }
+
+  getArea(): number {
+    const area = Math.PI * this.radius * this.radius;
+
+    return Math.floor(area * 100) / 100;
+  }
+}
+
+export class Rectangle implements Figure {
+  public shape = 'rectangle';
+
+  constructor(
+    public color: string,
+    public width: number,
+    public height: number,
+  ) {
+    if (width <= 0 || height <= 0) {
+      throw new Error('incorrect value');
+    }
+    this.color = color;
+  }
+
+  getArea(): number {
+    const area = this.width * this.height;
+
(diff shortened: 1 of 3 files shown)


src/figuresClasses.ts
```diff
@@ -1,11 +1,77 @@
...
+    public b: number,
+    public c: number,
+  ) {
+    const NotPositive = a <= 0 || b <= 0 || c <= 0;
+    const isNotTriangle = a + b <= c || a + c <= b || b + c <= a;
+
+    if (NotPositive || isNotTriangle) {
+      throw new Error('incorrect value');
+    }
+    this.color = color;
```
DenNice-cloud, line 22: As I know, we don't need to create here `this.color = color;` because in 11line, TypeScript do it instead of us
```suggestion

```
It's repeated again in the code belowe👀👀

src/figuresClasses.ts
```diff
@@ -1,11 +1,77 @@
...
+  }
+
+  getArea(): number {
+    const s = (this.a + this.b + this.c) / 2;
```
DenNice-cloud, line 26: We need more specific name for values, not just "s"
```suggestion
    const semiperimeter = (this.a + this.b + this.c) / 2;
```

src/figuresClasses.ts
```diff
@@ -1,11 +1,77 @@
...
+  shape: string;
+  color: string;
+  getArea(): number;
+}
+
+export class Triangle implements Figure {
+  shape: string = 'triangle';
+
+  constructor(
+    public color: string,
```
DenNice-cloud, line 11: > Write an interface Figure and 3 classes implementing it so that every figure has:
> - a shape (triangle, circle or rectangle);
> - a color (red, green or blue);

```suggestion
  shape: Shape= 'triangle';

  constructor(
    public color: Color,
```
You need to create Color and Shape **type** and give tham key values (keywords)🤔

DenNice-cloud (approved):
The basic idea for the task was completed and looks great, keep up the good work! 😎😎😎

src/figuresClasses.ts
```diff
@@ -1,11 +1,83 @@
...
+    public color: Color,
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    const NotPositive = a <= 0 || b <= 0 || c <= 0;
+    const isNotTriangle = a + b <= c || a + c <= b || b + c <= a;
+
+    if (NotPositive || isNotTriangle) {
+      throw new Error('incorrect value');
```
BudnikOleksii, line 23: [CODE STYLE] - If you throw an error try to provide its detailed description in error message.
image

src/figuresClasses.ts
BudnikOleksii: > As I know, we don't need to create here `this.color = color;` because in 11line, TypeScript do it instead of us
> 
> It's repeated again in the code belowe👀👀

Good point

src/figuresClasses.ts
```diff
@@ -1,11 +1,83 @@
...
+
+  constructor(
+    public color: Color,
+    public width: number,
+    public height: number,
+  ) {
+    if (width <= 0 || height <= 0) {
+      throw new Error('incorrect value');
+    }
+    this.color = color;
```
BudnikOleksii, line 71: The same

BudnikOleksii (changes requested):
Good work 👍 
Just a few moments should be fixed

BudnikOleksii (approved):
Looks good 🚀

PR #1960, opened 2024-03-13
Solution

VitaliiVashkevych:


olduma (approved):
Good job! 👍

src/figuresClasses.ts
```diff
@@ -1,11 +1,77 @@
-export interface Figure {}
+export interface Figure {
+  shape: string;
+  color: string;
```
BudnikOleksii, line 3: - shape is not a string, available values "triangle, circle or rectangle"
- color too, "red, green or blue"
image

src/figuresClasses.ts
```diff
@@ -1,11 +1,77 @@
...
+  getArea(): number;
+}
+
+export class Triangle implements Figure {
+  shape: string;
```
BudnikOleksii, line 8:
```suggestion
  shape = 'triangle';
```
- The type for `shape` is already set as this class implements Figure /
- You can set value here, as you don't get it from arguments in constructor

Fix it in all places

src/figuresClasses.ts
```diff
@@ -1,11 +1,77 @@
...
+  constructor(
+    public color: string,
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    this.shape = 'triangle';
+
+    if (a <= 0 || b <= 0 || c <= 0) {
+      throw new Error('Invalid input');
```
BudnikOleksii, line 19: [CODE STYLE] - If you throw an error try to provide its detailed description in error message.
image

Fix it in all places

src/figuresClasses.ts
```diff
@@ -1,11 +1,77 @@
...
+    }
+
+    if (a + b <= c || a + c <= b || b + c <= a) {
+      throw new Error('Invalid input');
+    }
+  }
+
+  getArea(): number {
+    const p = (this.a + this.b + this.c) / 2;
```
BudnikOleksii, line 28: What is `p`? Avoid one-letter variable names

BudnikOleksii (changes requested):
Almost, done 👍 
Let's fix types and a few more things that I mentioned in comments

BudnikOleksii (changes requested):
.

StasSokolov1 (approved):
Looks good to me, I'm approving it🔥

PR #1962, opened 2024-03-13
Solution

kubsen:


src/figuresClasses.ts
```diff
@@ -1,11 +1,90 @@
...
+    this.sides = [a, b, c];
+  }
+
+  getArea(): number {
+    const s = (this.sides[0] + this.sides[1] + this.sides[2]) / 2;
+
+    return Math.sqrt(
+      s * (s - this.sides[0]) * (s - this.sides[1]) * (s - this.sides[2]),
+    );
+  }
```
Davwidos, line 34: return of method `getArea()` should be rounded down to hundredths.

choeqq (changes requested):
@kubsen please make sure to adjust your solution to comments above 😉

PR #1965, opened 2024-03-13
solution

IRONH4MMER:


src/figuresClasses.ts
```diff
@@ -1,11 +1,85 @@
...
+    public c: number,
+  ) {
+    if (a <= 0 || b <= 0 || c <= 0) {
+      throw new Error('All sides must be greater than 0');
+    }
+
+    if (a >= b + c || b >= a + c || c >= a + b) {
+      throw new Error('Invalid triangle sides: cannot form a triangle');
+    }
+    this.color = color;
```
tetlisna, line 25: you dont need to do this.color = color

src/figuresClasses.ts
```diff
@@ -1,11 +1,85 @@
...
+    public color: `red` | `green` | `blue`,
+    public radius: number,
+  ) {
+    if (radius <= 0) {
+      throw new Error('Radius must be greater than 0');
+    }
 
-export class Circle implements Figure {}
+    this.color = color;
+    this.radius = radius;
```
tetlisna, line 49: use it inside constructor only    constructor( private radius: number)

tetlisna (changes requested):
Hi, great work, needs tiny improvement mentioned in checklist 
4. [CODE KNOWLEDGE] - When you declare a constructor parameter with an access modifier TypeScript implicitly creates a property on the class with the same name as the parameter and assigns the parameter value to that property. You don't need to do it manually.

BAD EXAMPLE: 
```
constructor(a: number, b: number, c: number) {
    this.a = a;
    this.b = b;
    this.c = c;
}
```

GOOD EXAMPLE: 
```
constructor(a: number, b: number, c: number) {

}
```

src/figuresClasses.ts
```diff
@@ -1,11 +1,79 @@
-export interface Figure {}
+interface Figure {
+  shape: `triangle` | `circle` | `rectangle`;
+  color: `red` | `green` | `blue`;
```
Moroz-Dmytro, line 3: Create separate types `Color` and `Shape`

src/figuresClasses.ts
```diff
@@ -1,11 +1,79 @@
...
+  getArea(): number;
+}
+
+export class Triangle implements Figure {
+  shape: 'triangle' = 'triangle';
```
Moroz-Dmytro, line 8:
```suggestion
  shape: Shape = 'triangle';
```

Use type that you will create here

src/figuresClasses.ts
```diff
@@ -1,11 +1,79 @@
...
+
+  sides: number[];
```
Moroz-Dmytro, line 10:
```suggestion
```

You don't need this. You can use your sides like `this.a` `this.b`...

src/figuresClasses.ts
```diff
@@ -1,11 +1,79 @@
...
+    public c: number,
+  ) {
+    if (a <= 0 || b <= 0 || c <= 0) {
+      throw new Error('All sides must be greater than 0');
+    }
+
+    if (a >= b + c || b >= a + c || c >= a + b) {
+      throw new Error('Invalid triangle sides: cannot form a triangle');
+    }
+    this.sides = [a, b, c];
```
Moroz-Dmytro, line 25:
```suggestion
```

src/figuresClasses.ts
```diff
@@ -1,11 +1,79 @@
...
+  }
+
+  getArea(): number {
+    const [a, b, c] = this.sides;
+    const s = (a + b + c) / 2;
```
Moroz-Dmytro, line 30: 1 letter variable name is a bad idea

src/figuresClasses.ts
```diff
@@ -1,11 +1,79 @@
...
+  getArea(): number;
+}
+
+export class Triangle implements Figure {
+  shape: 'triangle' = 'triangle';
+
+  sides: number[];
+
+  constructor(
+    public color: `red` | `green` | `blue`,
```
```suggestion
    public color: Color,
```

src/figuresClasses.ts
```diff
@@ -1,11 +1,79 @@
...
+    private radius: number,
+  ) {
+    if (radius <= 0) {
+      throw new Error('Radius must be greater than 0');
+    }
+  }
 
-export class Rectangle implements Figure {}
+  getArea(): number {
+    return Math.floor(Math.PI * this.radius ** 2 * 100) / 100;
```
Moroz-Dmytro, line 49: Create variable for `area` value. And then do `Math.floor(area * 100) / 100`

src/figuresClasses.ts
```diff
@@ -1,11 +1,79 @@
...
+    const s = width * height;
+
+    return s;
+  }
+}
 
-export function getInfo(figure): string {
-  return typeof figure;
+export function getInfo(figure: Figure): string {
+  return `A ${figure.color} ${figure.shape} - ${figure.getArea()}`;
```
Moroz-Dmytro, line 78: Destructuring

src/figuresClasses.ts
```diff
@@ -1,11 +1,79 @@
...
+  ) {
+    if (width <= 0 || height <= 0) {
+      throw new Error('All sides must be greater than 0');
+    }
+    this.sides = [width, height];
+  }
+
+  getArea(): number {
+    const [width, height] = this.sides;
+    const s = width * height;
```
Moroz-Dmytro, line 71: Bad variable name

src/figuresClasses.ts
```diff
@@ -1,11 +1,79 @@
...
+  }
+}
+
+export class Rectangle implements Figure {
+  shape: 'rectangle' = 'rectangle';
+
+  sides: number[];
+
+  constructor(
+    public color: `red` | `green` | `blue`,
```
Moroz-Dmytro, line 59: Same comments as above

src/figuresClasses.ts
```diff
@@ -1,11 +1,79 @@
...
+  }
+}
 
-export class Triangle implements Figure {}
+export class Circle implements Figure {
+  shape: 'circle' = 'circle';
 
-export class Circle implements Figure {}
+  constructor(
+    public color: `red` | `green` | `blue`,
```
Moroz-Dmytro, line 40: Same comments as above

Moroz-Dmytro (changes requested):
Good progress 👍 
But let's improve your code

src/figuresClasses.ts
BudnikOleksii: > You don't need this. You can use your sides like `this.a` `this.b`...

Not fixed

src/figuresClasses.ts
```diff
@@ -1,11 +1,89 @@
...
+    }
+
+    if (a >= b + c || b >= a + c || c >= a + b) {
+      throw new Error('Invalid triangle sides: cannot form a triangle');
+    }
+    this.sides = [this.a, this.b, this.c];
+  }
+
+  getArea(): number {
+    const [a, b, c] = this.sides;
```
BudnikOleksii, line 33: You still use `this.sides`. The problem is that you don't have one source of truth as you have stored values `a, b, c` and in the array sides. So if any value changes you have to update in two places

src/figuresClasses.ts
```diff
@@ -1,11 +1,89 @@
...
+    const area = Math.PI * this.radius ** 2;
+
+    return Math.floor(area * 100) / 100;
+  }
+}
+
+export class Rectangle implements Figure {
+  shape: Shape = 'rectangle';
+
+  sides: number[];
```
BudnikOleksii, line 62: The same

BudnikOleksii (changes requested):
Some of the previous reviewer comments are not fixes 🤷

StasSokolov1 (approved):
Looks good to me, I'm approving it🔥

PR #1968, opened 2024-03-14
add solution

illya-onyshchuk:


src/figuresClasses.ts
```diff
@@ -1,11 +1,84 @@
-export interface Figure {}
+export interface Figure {
+  shape: string;
+  color: string;
 
-export class Triangle implements Figure {}
+  getArea: () => number;
+}
```
Math8585, line 6: you should specify the type of color and shape

has:

Math8585 (changes requested):
Good job, just a little bit more to improve.

BudnikOleksii (approved):
Good work 👍

PR #1969, opened 2024-03-14
add solution

Graf6612:


src/figuresClasses.ts
```diff
@@ -1,11 +1,87 @@
-export interface Figure {}
+export interface Figure {
+  shape: 'triangle' | 'circle' | 'rectangle';
```
anastasia1383, line 2: better to create a type for it

src/figuresClasses.ts
```diff
@@ -1,11 +1,87 @@
...
+  ) {
+    if (a <= 0 || b <= 0 || c <= 0) {
+      throw new Error('All sides must be greater than 0');
+    }
+
+    if (a >= b + c || b >= a + c || c >= a + b) {
+      throw new Error('Invalid triangle sides: cannot form a triangle');
+    }
+    this.color = color;
+    this.sides = [a, b, c];
```
anastasia1383, line 26: you do not need an array here (i got this from a mentor)

src/figuresClasses.ts
```diff
@@ -1,11 +1,87 @@
...
+  }
+
+  getArea(): number {
+    const [a, b, c] = this.sides;
+    const s = (a + b + c) / 2;
```
anastasia1383, line 31: s is a bad name for a variable

src/figuresClasses.ts
```diff
@@ -1,11 +1,87 @@
...
+  constructor(
+    public color: 'red' | 'green' | 'blue',
+    public radius: number,
+  ) {
+    if (radius <= 0) {
+      throw new Error('Radius must be greater than 0');
+    }
 
-export class Circle implements Figure {}
+    this.color = color;
```
anastasia1383, line 48: you dont need to use this here, when you have a public property

anastasia1383 (commented):
Good job! Your code needs a little bit of refactoring.

dima8528:
I don't why I'm here. Everything is checked already )
But talking about that variable name, I don't agree with you.
image
You see, the name of that variable is present in the formula above. It totally explains all the context and subjectively looks okay. I know, usually we should give more context to vars' names, but I suppose this is the same case when we can call it an exception

src/figuresClasses.ts
okarnaukh: I completely agree with this point

okarnaukh (changes requested):

BudnikOleksii (approved):
Good job 👍

PR #1971, opened 2024-03-14
add task solution

vladarskyi:


src/figuresClasses.ts
```diff
@@ -1,21 +1,22 @@
...
   constructor(
-    public color: 'red' | 'green' | 'blue',
+    public color: Color,
     public a: number,
     public b: number,
     public c: number,
   ) {
-    this.shape = 'triangle';
-
     if (a + b <= c || a + c <= b || b + c <= a) {
```
yaGodya, line 20: you need to check in case one of the sides = 0
image

yaGodya (approved):
Good job, but need to fix 1 moment

src/figuresClasses.ts
```diff
@@ -1,11 +1,81 @@
...
+    ) {
+      throw new Error(
+        'The longest side of a triangle is not >= than a sum of two others',
+      );
+    }
+  }
+
+  public getArea(): number {
+    const { a, b, c } = this;
+    const s = (a + b + c) / 2;
```
BudnikOleksii, line 36: Avoid one-letter variables

src/figuresClasses.ts
```diff
@@ -1,11 +1,81 @@
...
+  constructor(
+    public color: Color,
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    if (
+      a === 0 ||
+      b === 0 ||
+      c === 0 ||
```
BudnikOleksii, line 23: What if any length is less than 0?

BudnikOleksii (changes requested):
Good work 👍 
Just don't forget to check for negative values as well

BudnikOleksii (approved):
👍

PR #1974, opened 2024-03-14
add task solution

Math8585:


src/figuresClasses.ts
```diff
@@ -1,11 +1,73 @@
...
+      throw new Error('Sides should be bigger than 0.');
+    }
+
+    if (a >= b + c || b >= a + c || c >= a + b) {
+      throw new Error('This sides cannot form a triangle');
+    }
+  }
+
+  getArea(): number {
+    const s = (this.a + this.b + this.c) / 2;
```
Dmitrel, line 26: What is s? Avoid one-letter variable names

Dmitrel (approved):
good job!

src/figuresClasses.ts
```diff
@@ -1,11 +1,73 @@
...
+
+    return (
+      Math.floor(
+        Math.sqrt(s * (s - this.a) * (s - this.b) * (s - this.c)) * 100,
```
Dmitrel, line 30: please move it to a separate function because you use this formula on every class and use here

src/figuresClasses.ts
```diff
@@ -1,11 +1,73 @@
...
+    public radius: number,
+  ) {
+    if (radius <= 0) {
+      throw new Error('Radius should be bigger than 0.');
+    }
+  }
 
-export class Triangle implements Figure {}
+  getArea(): number {
+    return Math.floor(Math.PI * this.radius ** 2 * 100) / 100;
```
Dmitrel, line 49: please move it to a separate function because you use this formula on every class and use here

src/figuresClasses.ts
```diff
@@ -1,11 +1,73 @@
...
+    public height: number,
+  ) {
+    if (width <= 0 || height <= 0) {
+      throw new Error('Width and height should be bigger than 0.');
+    }
+  }
 
-export class Rectangle implements Figure {}
+  getArea(): number {
+    return Math.floor((this.width * this.height * 100) / 100);
```
Dmitrel, line 67: please move it to a separate function because you use this formula on every class and use here

src/figuresClasses.ts
```diff
@@ -1,11 +1,73 @@
-export interface Figure {}
+export interface Figure {
+  shape: 'triangle' | 'circle' | 'rectangle';
+  color: 'red' | 'green' | 'blue';
```
BudnikOleksii, line 3: Create types for `shape` and `color` as you also use them below

src/figuresClasses.ts
```diff
@@ -1,11 +1,73 @@
...
+  getArea(): number;
+}
+
+export class Triangle implements Figure {
+  shape: 'triangle' = 'triangle';
+
+  constructor(
+    public color: 'red' | 'green' | 'blue',
```
BudnikOleksii, line 11:
```suggestion
    public color: Color,
```
And so on

src/figuresClasses.ts
BudnikOleksii: > What is s? Avoid one-letter variable names

Good point

src/figuresClasses.ts
BudnikOleksii: > please move it to a separate function because you use this formula on every class and use here later

Good point

BudnikOleksii (changes requested):
Good job 👍 
Let's improve your solution slightly check my and the previous reviewer comments

BudnikOleksii (approved):
Looks good 🔥

PR #1975, opened 2024-03-14
solution

marinatea:


src/figuresClasses.ts
```diff
@@ -1,11 +1,109 @@
...
+  getArea(): number {
+    return this.width * this.height;
+  }
+}
 
-export class Circle implements Figure {}
+export function getInfo(figure: Figure): string {
+  const area = figure.getArea();
+  const areaIntegered = Number.isInteger(area);
+  let areaString: string;
```
choeqq, line 100: why area would ever be a string?

choeqq (changes requested):

asakevych (approved):
Good job! :rocket:

PR #1976, opened 2024-03-14
task done

Oleksandr2108:


src/figuresClasses.ts
```diff
@@ -1,11 +1,104 @@
...
+export interface Figure {
+  shape: string;
+  color: string;
+  getArea(): number;
+}
+
+export class Triangle implements Figure {
+  shape: string;
+
+  color: string;
```
dima8528, line 10: Why do you use a blank line here?

src/figuresClasses.ts
```diff
@@ -1,11 +1,104 @@
...
+
+  private sideA: number;
+
+  private sideB: number;
+
+  private sideC: number;
```
dima8528, line 16: Same here. There is the same logic, you don't need to separate them. Fix this in other places as well pls

dima8528 (changes requested):
Looks good to me. Fix these issues and everything will be ok )

src/figuresClasses.ts
```diff
@@ -1,11 +1,106 @@
...
+  shape: Shape = 'triangle';
+  private sideA: number;
+  private sideB: number;
+  private sideC: number;
+
+  constructor(
+    public color: Color,
+    sideA: number,
+    sideB: number,
+    sideC: number,
```
BudnikOleksii, line 21:
```suggestion
  constructor(
    public color: Color,
    private sideA: number,
    private sideB: number,
    private sideC: number,
```
You can do it inside constructor

src/figuresClasses.ts
```diff
@@ -1,11 +1,106 @@
...
+  ) {
+    if (sideA <= 0 || sideB <= 0 || sideC <= 0) {
+      throw new Error('your error message');
```
BudnikOleksii, line 24: [CODE STYLE] - If you throw an error try to provide its detailed description in error message.
image

Fix it in all places

src/figuresClasses.ts
```diff
@@ -1,11 +1,106 @@
...
+      sideA + sideC <= sideB ||
+      sideC + sideB <= sideA
+    ) {
+      throw new Error('your error message');
+    }
+
+    this.color = color;
+    this.sideA = sideA;
+    this.sideB = sideB;
+    this.sideC = sideC;
```
BudnikOleksii, line 38: If you declare in a constructor as I mentioned typescript will do it for you, so these are redundant here

Fix in all places

BudnikOleksii (changes requested):
Good work 👍 
But could be simplified, so let's do it

src/figuresClasses.ts
```diff
@@ -1,11 +1,93 @@
...
+    if (radius <= 0) {
+      throw new Error(
+        'Incorrect input data. Negative numbers or zero are not allowed',
+      );
+    }
+  }
 
-export class Rectangle implements Figure {}
+  getArea(): number {
+    const area = Math.PI * this.radius * this.radius;
```
Moroz-Dmytro, line 63: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/pow

Moroz-Dmytro (changes requested):
Good work 👍 
Some comments below

BudnikOleksii (approved):
👍

PR #1977, opened 2024-03-14
add task solution

dima8528:


dima8528:
image
Locally all the tests have passed. Prettier...

dima8528:
image
Talking about the var's name ('s'), I've already said my position here

BudnikOleksii:
> image Talking about the var's name ('s'), I've already said my position here

In this case, it's ok as the method gets the area of a triangle. But if the function is more complicated you have to track all these things so to be better descriptive the name is, as much easier than to understand what it does.
It's not math, it's web development,

BudnikOleksii (approved):
Nicely done 🔥

PR #1978, opened 2024-03-14
solution

pasha1932:


src/figuresClasses.ts
```diff
@@ -1,11 +1,81 @@
-export interface Figure {}
+export interface Figure {
+  shape: string;
+  color: string;
```
```suggestion
  shape: 'triangle' | 'circle' | 'rectangle';
  color: 'red' | 'green' | 'blue';
```
better to specify it more precisely

etojeDenys (approved):

PR #1980, opened 2024-03-15
solution

andriy-yurkiv:


src/figuresClasses.ts
```diff
@@ -1,11 +1,81 @@
...
+  public color: string;
+
+  constructor(color: string, width: number, height: number) {
+    if (width <= 0 || height <= 0) {
+      throw new Error('Invalid rectangle dimensions');
+    }
+
+    this.height = height;
+    this.width = width;
+    this.color = color;
```
etojeDenys, line 71: simplify it using the [Typescript Constructor Shorthand](https://dev.to/satansdeer/typescript-constructor-shorthand-3ibd)

src/figuresClasses.ts
```diff
@@ -1,11 +1,81 @@
-export interface Figure {}
+export interface Figure {
+  shape: string;
+  color: string;
```
```suggestion
type Shape = 'triangle' | 'circle' | 'rectangle';
type Color = 'red' | 'green' | 'blue';

export interface Figure {
  shape: Shape;
  color: Color;
```
write types more precisely

etojeDenys (changes requested):

PR #1985, opened 2024-03-17
add task solution

hattoriKanto:


khromakeiqt (approved):
LGTM 🫡

src/figuresClasses.ts
```diff
@@ -1,11 +1,81 @@
-export interface Figure {}
+export interface Figure {
+  shape: string;
+  color: string;
```
Moroz-Dmytro, line 3: Create types for these values, they can't be any string

src/figuresClasses.ts
```diff
@@ -1,11 +1,81 @@
...
 
-export class Rectangle implements Figure {}
+    return Math.floor(area * 100) / 100;
+  }
+}
 
-export function getInfo(figure): string {
-  return typeof figure;
+export function getInfo(figure: Figure): string {
+  return `A ${figure.color} ${figure.shape} - ${figure.getArea()}`;
```
Moroz-Dmytro, line 80: Destrcutring

src/figuresClasses.ts
```diff
@@ -1,11 +1,81 @@
...
+  getArea(): number {
+    const semiperimeter = (this.a + this.b + this.c) / 2;
+    const area = Math.sqrt(
+      semiperimeter *
+        (semiperimeter - this.a) *
+        (semiperimeter - this.b) *
+        (semiperimeter - this.c),
+    );
+
+    return Math.floor(area * 100) / 100;
```
Moroz-Dmytro, line 38: Would be good to create separate function for rounding(you use `Math.floor(...* 100) / 100` at least 3 times
Remember about DRY rule

Moroz-Dmytro (changes requested):
Good progress 👍 
Some comments below

src/figuresClasses.ts
```diff
@@ -1,11 +1,91 @@
-export interface Figure {}
+export interface Figure {
+  shape: 'triangle' | 'circle' | 'rectangle';
```
Moroz-Dmytro, line 2: Create separate type  `type Shape = 'triangle' | 'circle' | 'rectangle'` and use it here

src/figuresClasses.ts
```diff
@@ -1,11 +1,91 @@
...
+  color: 'red' | 'blue' | 'green';
```
Moroz-Dmytro, line 3: Same comment as above

src/figuresClasses.ts
```diff
@@ -1,11 +1,91 @@
...
+
+export function roundArea(area: number): number {
+  return Math.floor(area * 100) / 100;
+}
+
+export class Triangle implements Figure {
+  readonly shape = 'triangle';
+
+  constructor(
+    readonly color: 'red' | 'blue' | 'green',
```
Moroz-Dmytro, line 15: Use type that you will create here and in other places

Moroz-Dmytro (changes requested):
Good work 👍 
But still some fixes need to be done

BudnikOleksii (approved):
Good work 👍

PR #1987, opened 2024-03-18
Solution

przwojwwp:


src/figuresClasses.ts
```diff
@@ -1,11 +1,76 @@
...
 
-export function getInfo(figure): string {
-  return typeof figure;
+  getArea(): number;
+}
+
+export class Triangle implements Figure {
+  constructor(
+    public color: Colors,
+    public a: number,
```
yevhenii-pyl, line 17: Let's avoid naming variables as 'a | b | c | p' - it's not an alphabet, each variable actually means something. 

If shape can be named as 'shape', why can't we use 'sideA' instead of just 'a'? :)

yevhenii-pyl (approved):
Good job! 👍 
However, always consider meaningful variable names - this may play the crucial difference between working and bugged code sometimes. :)

src/figuresClasses.ts
przwojwwp: Good question 😅
I renamed them

PR #1990, opened 2024-03-19
ts Figures Classes

Abdrakhmanov-Renat:


src/figuresClasses.ts
```diff
@@ -1,11 +1,79 @@
-export interface Figure {}
+export interface Figure {
+  shape: string;
+  color: string;
```
```suggestion
type Shapes = 'triangle' | 'circle' | 'rectangle';

type Colors = 'red' | 'green' | 'blue';

export interface Figure {
  shape: Shapes;
  color: Colors;
```
better to add types more precisely

etojeDenys (changes requested):

etojeDenys (changes requested):
.

etojeDenys (approved):
lgtm

PR #1991, opened 2024-03-19
added task solution

Stako21:


src/figuresClasses.ts
```diff
@@ -1,11 +1,73 @@
-export interface Figure {}
+export interface Figure {
+  shape: string;
+  color: string;
```
```suggestion
type Shapes = 'triangle' | 'circle' | 'rectangle';
type Colors = 'red' | 'green' | 'blue';

export interface Figure {
  shape: Shapes;
  color: Colors;
```
set types more precisely

etojeDenys (changes requested):

Stako21:
the Rectangle class starts at line 56 and ends at line 72

etojeDenys (approved):
good job

PR #1994, opened 2024-03-21
Solution1

Salmaralex:


src/figuresClasses.ts
```diff
@@ -1,11 +1,84 @@
-export interface Figure {}
+export interface Figure {
+  shape: string;
+  color: string;
```
```suggestion
type Shape = 'triangle' | 'circle' | 'rectangle';
type Color = 'red' | 'green' | 'blue';

interface Figure {
  color: Color;
  shape: Shape;
```
better to specify types more precisely

etojeDenys (changes requested):

etojeDenys (approved):
lgtm

PR #1997, opened 2024-03-21
solution

ant-tur:


src/figuresClasses.ts
```diff
@@ -1,11 +1,90 @@
...
+    public color: Color,
+    public radius: number,
+  ) {
+    if (this.radius <= 0) {
+      throw new Error("This circle doesn't exist");
+    }
+  }
+
+  getArea(): number {
+    const { radius } = this;
```
anastasiia-tilikina, line 55: i think it is not a good idea destructuring this keyword. It is non obvious
Especially here when you need to get only one property

src/figuresClasses.ts
```diff
@@ -1,11 +1,90 @@
...
+  ) {
+    if (this.height <= 0 || this.width <= 0) {
+      throw new Error('The sides must be greater than 0');
+    }
+  }
+
+  getArea(): number {
+    const { width, height } = this;
+
+    const s = Math.floor(width * height * 100) / 100;
```
anastasiia-tilikina, line 79: You could move rounding to a helper

anastasiia-tilikina (approved):
GJ

consider my comments for the future

PR #2009, opened 2024-03-25
Solution

Romanov7788:


src/figuresClasses.ts
```diff
@@ -1,11 +1,90 @@
...
+
+  color: string;
+
+  a: number;
+
+  b: number;
+
+  c: number;
+
+  constructor(color: string, a: number, b: number, c: number) {
```
etojeDenys, line 18: try to simplify it using the [Typescript Constructor Shorthand](https://dev.to/satansdeer/typescript-constructor-shorthand-3ibd)

etojeDenys (changes requested):

lerastarynets (approved):
You're on the right track!

PR #2010, opened 2024-03-26
Add solution

olafchuszno:


src/figuresClasses.ts
```diff
@@ -1,11 +1,71 @@
...
+  color: Color;
+  getArea: GetArea;
+}
+
+export class Triangle implements Figure {
+  shape: Shape = 'triangle';
+
+  constructor(
+    public color: Color,
+    public a: number,
```
yevhenii-pyl, line 18: Let's use more explicit naming here. The first var is color, so why second is just 'a'? 
Let's call it a 'sideA' and so on :)

src/figuresClasses.ts
```diff
@@ -1,11 +1,71 @@
...
+    public b: number,
+    public c: number,
+  ) {
+    if (a + b <= c) {
+      throw new Error();
+    }
+  }
+
+  getArea(): number {
+    const s: number = (1 / 2) * (this.a + this.b + this.c);
```
yevhenii-pyl, line 28: Just to avoid typing `this` each time, let's extract properties at the top level. 

```
  getArea(): number {
    { ... } = this;
```

yevhenii-pyl (changes requested):
Good job! Just a minor fix from a code style perspective and we are good to go.

yevhenii-pyl (approved):
🥇

PR #2015, opened 2024-03-27
Solution

Galers:


src/figuresClasses.ts
```diff
@@ -1,11 +1,78 @@
...
+export class Rectangle implements Figure {
+  shape = 'rectangle';
+
+  constructor(
+    public color: string,
+    public width: number,
+    public height: number,
+  ) {
+    if (width <= 0 || height <= 0) {
+      throw new Error('Method not implemented.');
```
TarasHoliuk, line 65: Error message is misleading - write the correct error

lerastarynets (approved):
Keep up the great work!

PR #2017, opened 2024-03-28
farmed

true-binar-White-Rabbit:


src/figuresClasses.ts
```diff
@@ -1,11 +1,81 @@
-export interface Figure {}
+export interface Figure {
+  shape: string;
+  color: string;
```
Esceype, line 3:
```suggestion
  shape: 'triangle' | 'circle' | 'rectangle';
  color: 'red' | 'green' | 'blue';
```

src/figuresClasses.ts
```diff
@@ -1,11 +1,81 @@
...
+  }
+
+  getArea(): number {
+
+
+    const p = (this.a + this.b + this.c) / 2;
+    const area = Math.sqrt(p * (p - this.a) * (p - this.b) * (p - this.c));
 
-export class Triangle implements Figure {}
+    return Math.round(area * 100) / 100;
```
Esceype, line 36:
```suggestion
    return Math.floor(area * 100) / 100;
```

In this case as you see, you use same like of code in two places, move it to some function and reuse

Esceype (changes requested):
GJ!
Read the following comments

src/figuresClasses.ts
```diff
@@ -1,11 +1,85 @@
-export interface Figure {}
+export interface Figure {
+  shape: 'triangle' | 'circle' | 'rectangle';
+  color: 'red' | 'green' | 'blue';
```
maxim2310, line 3: would be better create separate union type for these types, and use it everywhere below

maxim2310 (changes requested):

src/figuresClasses.ts
```diff
@@ -1,11 +1,87 @@
-export interface Figure {}
+type Shape = 'triangle' | 'circle' | 'rectangle';
 
-export class Triangle implements Figure {}
+export interface Figure {
+  shape: Shape;
+  color: 'red' | 'green' | 'blue';
```
maxim2310, line 5: need to do color in same way

maxim2310 (changes requested):

etojeDenys (approved):
lgtm

PR #2019, opened 2024-03-29
ts_figures_classes

ksstassy:


src/figuresClasses.ts
```diff
@@ -1,11 +1,75 @@
-export interface Figure {}
+export interface Figure {
+  shape: string;
+  color: string;
+  getArea(): number;
+}
+
+export class Triangle implements Figure {
+  shape: string = 'triangle';
```
anastasiiavorobiova, line 8: Consider using an enum for the shape
```suggestion
  shape: Shape = 'triangle';
```

src/figuresClasses.ts
```diff
@@ -1,11 +1,75 @@
...
+
+  constructor(
+    public color: string,
```
anastasiiavorobiova, line 11:
```suggestion
    public color: Color,
```
Consider using a type for the color

anastasiiavorobiova (changes requested):
Great work! Almost done!

src/figuresClasses.ts
```diff
@@ -1,11 +1,82 @@
...
+  Triangle = 'triangle',
+  Rectangle = 'rectangle',
+  Circle = 'circle',
+}
+
+type Color = 'red' | 'green' | 'blue';
+
+export interface Figure {
+  shape: string;
+  color: string;
```
etojeDenys, line 11:
```suggestion
enum Shape {
  Triangle = 'triangle',
  Rectangle = 'rectangle',
  Circle = 'circle',
}

type Color = 'red' | 'green' | 'blue';

export interface Figure {
  shape: Shape;
  color: Color;
```

etojeDenys (changes requested):
small comment

PR #2027, opened 2024-04-08
Solution

rwlp:
Hi!

### DIFF
--- a/src/figuresClasses.ts
+++ b/src/figuresClasses.ts
@@ -1,11 +1,88 @@
-export interface Figure {}
+export interface Figure {
+  shape: string;
+  color: string;
+  getArea: Function;
+}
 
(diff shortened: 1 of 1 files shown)


src/figuresClasses.ts
```diff
@@ -1,11 +1,88 @@
...
+    );
+  }
+}
+
+export class Circle implements Figure {
+  shape = 'circle';
+
+  constructor(
+    public color: string,
+    public rad: number,
```
DarkMistyRoom, line 48:
```suggestion
    public radius: number,
```

DarkMistyRoom: better use a full name of variables

DarkMistyRoom (approved):
Good job!

PR #2031, opened 2024-04-15
Solution

aMtiIV:


yevhenii-pyl (approved):
Good job!
But don't forget to provide meaningful names to your variables, try to be as explicit as possible naming something - it may play a huge(!) role as your codebase grows.

PR #2035, opened 2024-04-16
task done

Reaffith:


Reaffith:
I honestly dont know why i keep getting 1 test wrong.  `parseFloat((Math.PI * this.radius * this.radius).toFixed(2))` shouldnt round the number but it does :/

src/figuresClasses.ts
```diff
@@ -1,11 +1,88 @@
-export interface Figure {}
+export interface Figure {
+  shape: 'triangle' | 'circle' | 'rectangle';
+  color: 'red' | 'green' | 'blue';
+  getArea: () => number;
+}
```
```suggestion
type Shape = 'triangle' | 'circle' | 'rectangle'

export interface Figure {
  shape: Shape;
  color: 'red' | 'green' | 'blue';
  getArea: () => number;
}
```

src/figuresClasses.ts
```diff
@@ -1,11 +1,88 @@
...
+
+export class Triangle implements Figure {
+  constructor(
+    public color: 'red' | 'green' | 'blue',
```
etojeDenys, line 9:
```suggestion
    public color: Shape,
```

etojeDenys: do the same with color

src/figuresClasses.ts
```diff
@@ -1,11 +1,88 @@
...
+  getArea(): number {
+    const halfPerimeter: number = (this.a + this.b + this.c) / 2;
+
+    return parseFloat(
+      Math.sqrt(
+        halfPerimeter *
+          (halfPerimeter - this.a) *
+          (halfPerimeter - this.b) *
+          (halfPerimeter - this.c),
+      ).toFixed(2),
```
etojeDenys, line 29: try to use `Math.floor(num * 100) / 100` instead of toFixed and parseFloat

etojeDenys (changes requested):

src/figuresClasses.ts
Reaffith: didn't help :(

etojeDenys (approved):
lgtm

PR #2042, opened 2024-04-18
task solution

kinqbert:


src/figuresClasses.ts
```diff
@@ -1,11 +1,97 @@
-export interface Figure {}
+type Shape = 'triangle' | 'circle' | 'rectangle';
+type Color = 'red' | 'green' | 'blue';
 
-export class Triangle implements Figure {}
+export interface Figure {
+  shape: Shape;
+  color: Color;
+  getArea: object;
```
alexander-ignatow, line 7: I guess, you should use some function type here?

alexander-ignatow, line 7:
```suggestion
  getArea: () => number;
```

alexander-ignatow (approved):
Looks good, just mind the type of `getArea`

PR #2044, opened 2024-04-19
solution

manch0ffline:


src/figuresClasses.ts
```diff
@@ -1,11 +1,86 @@
...
+
+  color: Colors;
+
+  constructor(
+    color: Colors,
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    if (a <= 0 || b <= 0 || b <= 0) {
```
l4st1m0za, line 21: On this line you are checking if 'b' is less than or equal to 0 twice instead of checking 'a', 'b' and 'c'. It should be 'a <= 0 || b <= 0 || c <= 0.'

l4st1m0za (approved):
Good job!

PR #2046, opened 2024-04-19
task solution

MatheusStefan:


jv-aquino (changes requested):
Great code Matheus!! I would just ask you, as a side quest, to add "Figure" as a proto for every other shape so it's easier to reuse its code if we wanted to add common features or methods like (not in this case bc it's not a method) the getInfo()

src/figuresClasses.ts
```diff
@@ -1,11 +1,92 @@
...
 
-export class Rectangle implements Figure {}
+export function getInfo(figure: Figure): string {
+  const area: number = figure.getArea();
 
-export function getInfo(figure): string {
-  return typeof figure;
+  if (Number.isInteger(area)) {
+    return `A ${figure.color} ${figure.shape} - ${area}`;
+  } else {
```
pedro-ruas, line 89: Else is not needed here, because the **if** contains a return. So, when the if runs, nothing else will run after.
In this way, you could move the return inside the **else** to the function itself as a default return.

Another way possible would be to simply return the string and use a **ternary operator** for the last parameter inside of it. like so:
```
 return `A ${figure.color} ${figure.shape} - ${Number.isInteger(area) ? area : area.toFixed(2)}`; 
```
Or you could always use the method .toFixed(2) like so: (if the number is an integer, it will return the number with two zeros as decimal cases)
```
 return `A ${figure.color} ${figure.shape} - ${area.toFixed(2)}`; 
```

And last, but not least, *getInfo* could be a method of the *Shape* class directly and access all needed values using *this*. But no problem having it here.

pedro-ruas (changes requested):
Great job overall! Minimal improvement.
Great use of extension and implementation of classes/interfaces!

src/figuresClasses.ts
MatheusStefan: That approach is way cleaner, thank's!

I tried to set _getInfo_ as a method of _Shape_ , but this change fails in some tests, so I'm keeping the function apart from the class if that's okay..

As what I can understand, I can't set this method inside the class, 'cause I need to call it and pass the object as argument, is that right? Or am I able to do it but and I'm missing something?

"Also create a function `getInfo` that takes a figure and returns a string in the
next format:"
```typescript
const redRectangle = new Rectangle('red', 3, 5);
getInfo(redRectangle) === 'A red rectangle - 15';

const greenCircle = new Circle('green', 1);
getInfo(greenCircle) === 'A green circle - 3.14';
```

EdPirro (approved):
Great job, @MatheusStefan! Keep it up! 🚀

PR #2053, opened 2024-04-23
solution done

emberlast1:


src/figuresClasses.ts
```diff
@@ -1,11 +1,78 @@
-export interface Figure {}
+export interface Figure {
+  shape: string;
+  color: string;
+  getArea(): number;
+}
```
```suggestion
type Color = 'red' | 'blue' | 'green';
type Shape = 'triangle' | 'rectangle' | 'circle';

export interface Figure {
  shape: Shape;
  color: Color;
  getArea(): number;
}
```
specify it more precisely

etojeDenys (changes requested):

anastasiia-tilikina (approved):
GJ!

PR #2056, opened 2024-04-25
Add task solution

eduvinagre:


src/figuresClasses.ts
```diff
@@ -1,11 +1,106 @@
...
+
+export class Triangle implements Figure {
+  readonly shape: 'triangle';
+
+  readonly color: 'red' | 'green' | 'blue';
+
+  a: number;
+
+  b: number;
+
```
jv-aquino, line 16: these blank lines may be unnecessary

jv-aquino (changes requested):
Hey Edu, great job, nice use of TS!
In terms of JavaScript / OOP, we can make this code more concise, don't you think? 

side quest: look at the repetition of declarations like this.shape and stuff and think about a way of reusing common code for all of them

src/figuresClasses.ts
eduvinagre: I totally agree with you Boss! Very nice observations, thank you very much. As for the spaces, unfortunately they were not done by me  they were made by mate's Lint script when testing my code. I'll get into those improvements right away.

src/figuresClasses.ts
```diff
@@ -1,11 +1,106 @@
...
+export class Triangle extends Shape {
+  a: number;
+
+  b: number;
+
+  c: number;
+
+  constructor(
+    color: 'red' | 'green' | 'blue',
```
pedro-ruas, line 29: You could reference the Figure type here like you did for Shape

src/figuresClasses.ts
```diff
@@ -1,11 +1,106 @@
...
 
-export class Circle implements Figure {}
+    return Math.floor(area * 100) / 100;
+  }
+}
+
+export class Circle extends Shape {
+  radius: number;
+
+  constructor(color: 'red' | 'green' | 'blue', radius: number) {
```
pedro-ruas, line 66: Same thing here

src/figuresClasses.ts
```diff
@@ -1,11 +1,106 @@
...
+  }
+}
 
-export class Rectangle implements Figure {}
+export class Rectangle extends Shape {
+  width: number;
+
+  height: number;
+
+  constructor(color: 'red' | 'green' | 'blue', width: number, height: number) {
```
pedro-ruas, line 87: Same here.
By declaring the prop inside the Figure Interface and later on passing strings for testing instead of referencing the Interface itself you create the issue:
- You are creating a value verification in each class instead of using a single source of truth (Figure Interface), in case the Interface changes, allowing for more or less colors, the classes will not allow it.

pedro-ruas (changes requested):

src/figuresClasses.ts
eduvinagre: Very good observations. Sorry for my lack of understanding here, must be the flue and medicine I'm taking. Cheers!

src/figuresClasses.ts
jv-aquino: kkkkkkkkkkkk boa edu

jv-aquino (approved):
Great job Edu!

PR #2057, opened 2024-04-25
add task solution

HigorPedroso:


jv-aquino (changes requested):
Great job Higor! Your typescript code is good but I have a side quest for you to improve your JavaScript and OOP principles!

side quest: make your code more concise to avoid rewriting the same code like you are doing in the constructor (reflect a bit on OOP principles)!

jv-aquino (approved):
Great job Higor! Nice way of doing it, I was more thinking about prototypes and class inheritance but you also did a really great job with only ts!

PR #2058, opened 2024-04-25
Solution

odunchyk:


src/figuresClasses.ts
```diff
@@ -1,11 +1,93 @@
...
+
+export interface Figure {
+  shape: Shape;
+  color: Color;
+  getArea(): number;
+}
 
-export class Triangle implements Figure {}
+export class Triangle implements Figure {
+  readonly shape: Shape = Shape.Triangle;
```
alexander-ignatow, line 20: Here and in other places - you may skip some types, as TypeScript is smart enough to guess them

```suggestion
  readonly shape = Shape.Triangle;
```

alexander-ignatow (approved):
👍

PR #2059, opened 2024-04-25
wip: add

mfcastilho:


src/figuresClasses.ts
```diff
@@ -1,11 +1,95 @@
-export interface Figure {}
+export interface Figure {
+  shape: string;
+  color: string;
```
jv-aquino, line 3: I think you can restrict the types a bit more

jv-aquino (changes requested):
Nice Mario, but I think we can improve this code a bit - try to avoid rewriting the same code - think about prototypes and also restrict your types a bit!

jv-aquino (approved):
Great job Mario! nice use of type variables!

PR #2061, opened 2024-04-25
add task solution

JohnyJDX:


src/figuresClasses.ts
```diff
@@ -1,11 +1,79 @@
-export interface Figure {}
+export interface Figure {
+  shape: string;
```
alexander-ignatow, line 2: You need to limit possible shapes (say using the union type)
```suggestion
  shape:'triagle' | 'circle' | 'rectangle';
```

src/figuresClasses.ts
```diff
@@ -1,11 +1,79 @@
...
+  ) {
+    if (radius <= 0) {
+      throw new Error('Radius must be greater than 0');
+    }
+  }
+
+  getArea(): number {
+    if (this.radius === 6) {
+      return 113.09;
+    }
```
alexander-ignatow, line 53: ?

alexander-ignatow (changes requested):
Good, job, but you need to improve some issues

alexander-ignatow (approved):
👍

PR #2062, opened 2024-04-25
task solved

ekateriina7:


src/figuresClasses.ts
```diff
@@ -1,11 +1,76 @@
-export interface Figure {}
+export interface Figure {
+  shape: 'triangle' | 'circle' | 'rectangle';
+  color: 'red' | 'green' | 'blue';
```
maxim2310, line 3: create separate type for these strings and use it below. 
for shape even would be  better to create enum

vadiimvooo (approved):
Well done!)

PR #2063, opened 2024-04-25
Solution

MakhmudKerimov:


src/figuresClasses.ts
```diff
@@ -1,11 +1,89 @@
-export interface Figure {}
+type Color = 'red' | 'blue' | 'green';
+type Shape = 'triangle' | 'rectangle' | 'circle';
```
maxim2310, line 2: you could use enum here

maxim2310 (approved):
Good work

PR #2064, opened 2024-04-25
Solution

ropanefarias:


jv-aquino (changes requested):
Nice Rodrigo, but there changes that we can make to create a more concise and reusable code! Try to make more reusability of the common functionalities of the three shapes - remember about prototypes

Also try restricting the types more to ensure type safety - for example, color has only 3 possible values

EdPirro (approved):
Great job, @ropanefarias! Keep it up!

PR #2066, opened 2024-04-26
Solution

MarchAlbion:


src/figuresClasses.ts
```diff
@@ -1,11 +1,93 @@
...
+  shape: 'triangle' | `circle` | `rectangle`;
+  color: 'red' | 'green' | 'blue';
+  getArea(): number;
+}
+
+export class Triangle implements Figure {
+  readonly shape: Figure['shape'] = 'triangle';
+
+  constructor(
+    public color: 'red' | 'green' | 'blue',
```
alexander-ignatow, line 11:
```suggestion
    public color: Figure['color']
```

Please, check other places as well

src/figuresClasses.ts
```diff
@@ -1,11 +1,93 @@
...
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    const sorted: number[] = [a, b, c].sort((el1, el2) => el2 - el1);
+
+    if (this.a <= 0 || this.b <= 0 || this.c <= 0) {
+      throw new Error('all sides should be positive number');
+    }
```
alexander-ignatow, line 20: Makes sense to put this check, before any other

src/figuresClasses.ts
```diff
@@ -1,11 +1,93 @@
...
+export class Triangle implements Figure {
+  readonly shape: Figure['shape'] = 'triangle';
+
+  constructor(
+    public color: 'red' | 'green' | 'blue',
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    const sorted: number[] = [a, b, c].sort((el1, el2) => el2 - el1);
```
alexander-ignatow, line 16: Not required - but I would not use `sort` for such case

I would suggest looking at `Math.max`

src/figuresClasses.ts
```diff
@@ -1,11 +1,93 @@
...
+    }
+
+    this.color = color;
+    this.a = a;
+    this.b = b;
+    this.c = c;
+  }
+
+  getArea(): number {
+    const semiparam = (this.a + this.b + this.c) / 2;
```
alexander-ignatow, line 35: `semiparam` sounds bit weird :)

I would suggest `halfSum` or something like it

alexander-ignatow (changes requested):
Looks good, but need to fix some issues

mrkovaliv (approved):
👍 👍 👍

PR #2067, opened 2024-04-26
figuresClasses Solition

sevenlays:


src/figuresClasses.ts
```diff
@@ -1,11 +1,81 @@
-export interface Figure {}
+export interface Figure {
+  shape: string;
```
alexander-ignatow, line 2: Use union type for the shapes

src/figuresClasses.ts
```diff
@@ -1,11 +1,81 @@
...
+  color: string;
```
alexander-ignatow, line 3: Per task:
> a color (red, green or blue);

please use union type here

src/figuresClasses.ts
```diff
@@ -1,11 +1,81 @@
...
+export class Triangle implements Figure {
+  shape = 'triangle';
+
+  constructor(
+    public color: string,
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    const sides = [this.a, this.b, this.c].sort((x, y) => y - x);
```
alexander-ignatow, line 16: You may skip it - but I would avoid `sort` here

You may look at `Math.max` instead

src/figuresClasses.ts
```diff
@@ -1,11 +1,81 @@
...
+    public width: number,
+    public height: number,
+  ) {
+    if (width <= 0 || height <= 0) {
+      throw new Error('Any side of Rectangle cant be less than 0');
+    }
+  }
+
+  getArea(): number {
+    const area = (this.width * this.height * 100) / 100;
```
alexander-ignatow, line 73: Minor - but in all three figure you are using three different ways to round the area

What is the reason to do this?
Can you unify it if possible

alexander-ignatow (changes requested):
Looks good, but some fixes are needed

src/figuresClasses.ts
sevenlays: Can show me how to do that with Math.max, please)

sevenlays: I want to see your solution with Math.max because in my a lot of lines

sevenlays: I chose sort because this way I have access not only to the largest side but also to all the others

PR #2071, opened 2024-04-27
add task solution

Voloshyn-Oleg:


src/figuresClasses.test.ts
```diff
@@ -116,7 +116,7 @@ describe('Method "getArea" of class "Circle"', () => {
     const circle = new Circle('red', 6);
 
     expect(circle.getArea())
-      .toBe(113.09);
```
alexander-ignatow, line 119: Mathematically you are right (113,0973354) - but for the future - please report such cases to the chat first

In real projects changing the tests assets can happen in different ways, and often just re-writing them is not what you should do

src/figuresClasses.ts
```diff
@@ -1,11 +1,70 @@
-export interface Figure {}
+export interface Figure {
+  shape: string;
+  color: string;
```
alexander-ignatow, line 3: per task only limited set of the values is allowed here - please consider using union types

alexander-ignatow (changes requested):
Looks good, but some minor issues are needed to be fixed

src/figuresClasses.ts
```diff
@@ -1,11 +1,74 @@
-export interface Figure {}
+export interface Figure {
+  shape: 'triangle' | 'circle' | 'rectangle';
+  color: 'red' | 'green' | 'blue';
```
mrkovaliv, line 3: the same

PR #2073, opened 2024-04-27
add solution

iAlfreX:


src/figuresClasses.ts
```diff
@@ -1,11 +1,124 @@
...
+        'Invalid side length for triangle: b must be greater than 0.',
+      );
+    }
 
-export class Triangle implements Figure {}
+    if (c <= 0) {
+      throw new Error(
+        'Invalid side length for triangle: c must be greater than 0.',
+      );
+    }
```
alexander-ignatow, line 54: Not critical, but you can probably unite it like

````
if ((a <= 0) || (b <= 0) || (c <= 0)) {
 throw new Error(
        'Negative sides are not allowed',
      );
}
````

File: src/figuresClasses.ts [hunk 2]
```diff
@@ -1,11 +1,124 @@
...
+
+function isValidTriangle(
+  firstSide: number,
+  secondSide: number,
+  thirdSide: number,
+): boolean {
+  const maxSide = Math.max(firstSide, secondSide, thirdSide);
+
+  return maxSide < firstSide + secondSide + thirdSide - maxSide;
+}
```
alexander-ignatow, line 27: This should be a static method of the `Triangle` class

alexander-ignatow (changes requested):
Good, but some changes are needed

iAlfreX:
Done

mrkovaliv (approved):
👍 👍 👍

PR #2075, opened 2024-04-28
Solution

pKaterynaa:


src/figuresClasses.ts
```diff
@@ -1,11 +1,91 @@
...
+
+  b: number;
+
+  c: number;
+
+  constructor(color: Color, a: number, b: number, c: number) {
+    this.color = color;
+    this.a = a;
+    this.b = b;
+    this.c = c;
```
vadiimvooo, line 25: You can simplify it. Read the [article ](https://dev.to/satansdeer/typescript-constructor-shorthand-3ibd) how you can do it

vadiimvooo: The same is for other classes

src/figuresClasses.ts
```diff
@@ -1,11 +1,91 @@
-export interface Figure {}
+type Shape = 'triangle' | 'circle' | 'rectangle';
+type Color = 'red' | 'green' | 'blue';
```
vadiimvooo, line 2: you could use enum here

vadiimvooo (changes requested):
Good job. Just a few fixes.

vadiimvooo (approved):
Well done!)

PR #2078, opened 2024-04-28
solution

Olek-91:


src/figuresClasses.ts
```diff
@@ -1,11 +1,94 @@
-export interface Figure {}
+type Shape = `triangle` | `circle` | `rectangle`;
+type Color = `red` | `green` | `blue`;
+// enum Parameters {
+//   a: number,
+//   b: number,
+//   c: number,
+//   radius: number,
+//   color: color,
+// }
```
alexander-ignatow, line 9: No need to keep this

src/figuresClasses.ts
```diff
@@ -1,11 +1,94 @@
...
 
-export class Triangle implements Figure {}
+export interface Figure {
+  shape: Shape;
+  color: Color;
+  getArea: () => number;
+}
+
+function longestSide(a: number, b: number, c: number): boolean {
```
alexander-ignatow, line 17: 1. From the function name does not clear what it is doing. As of now, it sounds like it just picks and returns the longest side.
2. This function should be a static method inside `Triangle` class

src/figuresClasses.ts
```diff
@@ -1,11 +1,94 @@
...
+  shape: Shape = 'triangle';
+
+  constructor(
+    public color: Color,
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    if (longestSide(a, b, c) || a <= 0 || b <= 0 || c <= 0) {
+      throw new Error('your error message');
```
alexander-ignatow, line 37: Please add valid error message

src/figuresClasses.ts
```diff
@@ -1,11 +1,94 @@
...
+
+export class Circle implements Figure {
+  shape: Shape = 'circle';
+
+  constructor(
+    public color: Color,
+    public radius: number,
+  ) {
+    if (this.radius <= 0) {
+      throw new Error('your error message');
```
alexander-ignatow, line 61: This and other error messages are also needed to be fixed

src/figuresClasses.ts
```diff
@@ -1,11 +1,94 @@
...
+    }
+  }
+
+  getArea(): number {
+    if (this.radius === 6) {
+      return 113.09;
+    }
```
alexander-ignatow, line 68: ?

alexander-ignatow (changes requested):
Good start, but some fixes are needed

alexander-ignatow (approved):
👍

PR #2079, opened 2024-04-29
add task solution

Esther-Guimaraes:


src/figuresClasses.ts
```diff
@@ -1,11 +1,92 @@
-export interface Figure {}
+export interface Figure {
+  shape: string;
+  color: string;
```
jv-aquino, line 3: here

src/figuresClasses.ts
```diff
@@ -1,11 +1,92 @@
...
+  getArea(): number;
+}
+
+export class Triangle implements Figure {
+  color: string;
+
+  shape: string;
```
jv-aquino, line 10: here

jv-aquino (changes requested):
Nice job Esther, but there are some corrections that will make your TS fabulous!

- First, try always ensuring the type safety to the max possible - example: the colors can only be 3 and the shapes are also restricted, so you should use TS to avoid that.
- Second, try making code more reusable and concise (without repeating things a lot) - remember about prototypes (just focus on TS in this case, u don't necessarily need protos)

jv-aquino (approved):
Nice Esther, you really got it!! Amazing TS!

PR #2080, opened 2024-04-29
add solution

stepanyukvova200:


src/figuresClasses.ts
```diff
@@ -1,11 +1,81 @@
-export interface Figure {}
+export interface Figure {
+  shape: string;
+  color: string;
```
alexander-ignatow, line 3: Per task:
> a shape (triangle, circle or rectangle);
> a color (red, green or blue);

You should use different type here

alexander-ignatow (changes requested):
Looks good, but minor fix of the typings is needed

src/figuresClasses.ts
```diff
@@ -1,11 +1,81 @@
-export interface Figure {}
+export interface Figure {
+  shape: 'triangle' | 'circle' | 'rectangle';
+  color: 'red' | 'green' | 'blue';
```
mrkovaliv, line 3: move on this to special Type to reuse everywhere

ColorType = 'red' | 'green' | 'blue'

src/figuresClasses.ts
```diff
@@ -1,11 +1,81 @@
-export interface Figure {}
+export interface Figure {
+  shape: 'triangle' | 'circle' | 'rectangle';
```
mrkovaliv, line 2: the same

mrkovaliv (approved):
👍 👍 👍 👍

PR #2083, opened 2024-04-29
solution

AYupko:


src/figuresClasses.ts
```diff
@@ -1,11 +1,78 @@
...
+  ) {
+    if (width <= 0 || height <= 0) {
+      throw new Error('Width and height cannot be negative or equal to 0');
+    }
+  }
+
+  getArea(): number {
+    const { width, height } = this;
+
+    return +(width * height).toFixed(2);
```
DanilWeda, line 69:
```suggestion
    return Number(width * height).toFixed(2);
```
as alternative*

DanilWeda (approved):
Great. 
Approved.
Thanks)

PR #2086, opened 2024-04-29
add solving

ordila:


src/figuresClasses.ts
```diff
@@ -1,11 +1,68 @@
-export interface Figure {}
+export interface Figure {
+  shape: string;
+  color: string;
```
DanilWeda, line 3: You need to describe all cases for shape and color, for example `'red' | 'green' | 'blue'`

DanilWeda (changes requested):
Good job. 
Check a coment for improvement. 
Thanks)

mrkovaliv (approved):
👍 👍 👍

PR #2087, opened 2024-04-29
class figure implemented

thisIvanRepo:
All tests work:)

### DIFF
--- a/src/figuresClasses.ts
+++ b/src/figuresClasses.ts
@@ -6,8 +6,14 @@ export interface Figure {
   getArea(): number;
 }
 
+enum FigureType {
+  triangle = 'triangle',
+  circle = 'circle',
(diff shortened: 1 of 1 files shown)


src/figuresClasses.ts
```diff
@@ -1,11 +1,74 @@
-export interface Figure {}
+type Color = 'red' | 'blue' | 'green';
 
-export class Triangle implements Figure {}
+export interface Figure {
+  shape: string;
```
anastasiiavorobiova, line 4: Consider creating an enum for the shape

anastasiiavorobiova (changes requested):
Great work! Almost done!

anastasiiavorobiova (approved):
Awesome!

PR #2089, opened 2024-04-30
add task solution

ArtemDemeniev:


src/figuresClasses.ts
```diff
@@ -1,11 +1,86 @@
-export interface Figure {}
+export interface Figure {
+  shape: 'triangle' | 'circle' | 'rectangle';
+  color: 'red' | 'green' | 'blue';
+  getArea: () => number;
+}
+
+export class Triangle implements Figure {
+  shape: 'triangle' = 'triangle';
```
alexander-ignatow, line 8:
```suggestion
  shape = 'triangle';
```

src/figuresClasses.ts
```diff
@@ -1,11 +1,86 @@
...
+
+  getArea: () => number;
+
+  constructor(
+    public color: 'red' | 'green' | 'blue',
```
alexander-ignatow, line 13: Notice, you repeating this `'red' | 'green' | 'blue'` in several places
That is not ok - imagine you would need to add support for another color

You need to either create separate `Color` type and reference it, or simply use `Figure[color]`

alexander-ignatow (changes requested):
Good, just some minor changes are needed

PR #2090, opened 2024-04-30
add task solution

indigo04:


src/figuresClasses.ts
```diff
@@ -1,11 +1,75 @@
-export interface Figure {}
+export interface Figure {
+  shape: 'triangle' | 'circle' | 'rectangle';
+  color: 'red' | 'blue' | 'green';
```
maxim2310, line 3: need to create separate type for these

src/figuresClasses.ts
```diff
@@ -1,11 +1,75 @@
-export interface Figure {}
+export interface Figure {
+  shape: 'triangle' | 'circle' | 'rectangle';
```
maxim2310, line 2: for shape would be even better to create enum

src/figuresClasses.ts
```diff
@@ -1,11 +1,75 @@
...
+  }
+}
 
-export class Circle implements Figure {}
+export class Circle implements Figure {
+  constructor(
+    public color: 'red' | 'green' | 'blue',
+    public radius: number,
+    public shape: 'circle' = 'circle',
+  ) {
```
maxim2310, line 41:
```suggestion
export class Circle implements Figure {
public shape = ShapeEnum.circle,
  constructor(
    public color: 'red' | 'green' | 'blue',
    public radius: number,
  ) {
```

maxim2310 (changes requested):
Good work, but lets make some improvement

src/figuresClasses.ts
```diff
@@ -1,11 +1,86 @@
-export interface Figure {}
+export interface Figure {
+  shape: 'triangle' | 'circle' | 'rectangle';
+  color: 'red' | 'blue' | 'green';
```
```suggestion
export interface Figure {
  shape: Shape;
  color: Сolor;
```

etojeDenys (changes requested):
small comment

anastasiiavorobiova (approved):
Awesome work!

PR #2091, opened 2024-04-30
add task solution

SlawikH:


src/figuresClasses.ts
```diff
@@ -1,11 +1,89 @@
-export interface Figure {}
+export interface Figure {
+  shape: string;
+  color: string;
```
alexander-ignatow, line 3: per task - these properties have limited possible values
use something like union types or enums here

src/figuresClasses.ts
```diff
@@ -1,11 +1,89 @@
...
+  getArea(): number;
+}
+
+export class Triangle implements Figure {
+  shape: string = 'triangle';
```
alexander-ignatow, line 8:
```suggestion
  shape = 'triangle';
```

alexander-ignatow (changes requested):
Good, but some changes are needed

alexander-ignatow (changes requested):
`shape` and `color` in the `Figure` still not fixed

src/figuresClasses.ts
```diff
@@ -1,11 +1,98 @@
...
+  ) {
+    if (radius <= 0) {
+      throw new Error('Radius must be a positive numbers');
+    }
+  }
+
+  getArea(): number {
+    const area = Math.PI * this.radius ** 2;
+
+    return Math.floor(area * 100) / 100;
```
undeletable, line 74: [Not critical] Handling of area (`Math.floor(area * 100) / 10`) occurs in multiple places, so it makes sense to define it in a separate method, in order to avoid self-duplicating.

undeletable (approved):
Looks fine, approving

PR #2092, opened 2024-04-30
Figure Solution

ilyasolo881:
Have a nice day :)

### DIFF
--- a/src/figuresClasses.ts
+++ b/src/figuresClasses.ts
@@ -1,11 +1,82 @@
-export interface Figure {}
+type Color = 'red' | 'green' | 'blue';
+type Shape = 'triangle' | 'circle' | 'rectangle';
 
-export class Triangle implements Figure {}
+export interface Figure {
+  shape: Shape;
+  color: Color;
+  getArea: () => number;
+}
+export class Triangle implements Figure {
+  shape: Shape = 'triangle';
 
-export class Circle implements Figure {}
+  constructor(
+    public color: Color,
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    if (!(a + b > c && b + c > a && a + c > b) || a <= 0 || b <= 0 || c <= 0) {
+      throw new Error('Error');
+    }
+  }
 
-export class Rectangle implements Figure {}
+  getArea = (): number => {
+    const a = this.a;
+    const b = this.b;
+    const c = this.c;
 
-export function getInfo(figure): string {
-  return typeof figure;
+    const s = 0.5 * (a + b + c);
+
+    const area = Math.sqrt(s * (s - a) * (s - b) * (s - c));
+
+    return Math.floor(area * 100) / 100;
+  };
+}
+
+export class Circle implements Figure {
+  shape: Shape = 'circle';
+
+  constructor(
+    public color: Color,
+    public radius: number,
+  ) {
+    if (radius <= 0) {
+      throw new Error('Error');
+    }
+  }
+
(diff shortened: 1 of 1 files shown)


src/figuresClasses.ts
```diff
@@ -1,11 +1,82 @@
...
+type Shape = 'triangle' | 'circle' | 'rectangle';
 
-export class Triangle implements Figure {}
+export interface Figure {
+  shape: Shape;
+  color: Color;
+  getArea: () => number;
+}
+export class Triangle implements Figure {
+  shape: Shape = 'triangle';
```
alexander-ignatow, line 10: it is ok just to 
```suggestion
  shape = 'triangle';
```

src/figuresClasses.ts
```diff
@@ -1,11 +1,82 @@
...
+
+  constructor(
+    public color: Color,
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    if (!(a + b > c && b + c > a && a + c > b) || a <= 0 || b <= 0 || c <= 0) {
+      throw new Error('Error');
+    }
```
alexander-ignatow, line 20: You need to be more specific with the error messages

src/figuresClasses.ts
```diff
@@ -1,11 +1,82 @@
...
+  }
+
+  getArea = (): number => {
+    const a = this.a;
+    const b = this.b;
+    const c = this.c;
```
alexander-ignatow, line 26: Why do you copy these variables?

src/figuresClasses.ts
```diff
@@ -1,11 +1,82 @@
...
+
+export class Circle implements Figure {
+  shape: Shape = 'circle';
+
+  constructor(
+    public color: Color,
+    public radius: number,
+  ) {
+    if (radius <= 0) {
+      throw new Error('Error');
```
alexander-ignatow, line 44: Fix the error message

src/figuresClasses.ts
```diff
@@ -1,11 +1,82 @@
...
+export class Rectangle implements Figure {
+  shape: Shape = 'rectangle';
+
+  constructor(
+    public color: Color,
+    public width: number,
+    public height: number,
+  ) {
+    if (width <= 0 || height <= 0) {
+      throw new Error('Error');
```
alexander-ignatow, line 66: fix the error message

src/figuresClasses.ts
```diff
@@ -1,11 +1,82 @@
...
+    }
+  }
+
+  getArea = (): number => {
+    const width = this.width;
+    const height = this.height;
```
alexander-ignatow, line 72: why copy?

src/figuresClasses.ts
```diff
@@ -1,11 +1,82 @@
...
+    public radius: number,
+  ) {
+    if (radius <= 0) {
+      throw new Error('Error');
+    }
+  }
 
-export class Circle implements Figure {}
+  getArea = (): number => {
+    const radius = this.radius;
```
alexander-ignatow, line 49: why copy?

alexander-ignatow (changes requested):
Good, but you need to fix some issues

ilyasolo881:
Fixed all error message, and copys. But when write like: shape = 'triangle';

i take error:
Property 'shape' in type 'Triangle' is not assignable to the same property in base type 'Figure'. Type 'string' is not assignable to type 'Shape'.

mrkovaliv (approved):
👍 👍 👍

PR #2093, opened 2024-04-30
solution

tarassamchenko:


src/figuresClasses.ts
```diff
@@ -1,11 +1,96 @@
...
 
-export class Triangle implements Figure {}
+export interface Figure {
+  shape: Shape;
+  color: Color;
+  getArea: () => number;
+}
+
+export class Triangle implements Figure {
+  shape: Shape;
```
alexander-ignatow, line 11: You can just have 

```suggestion
  shape = 'triangle';
```
So you don't need to set it in the constructor
This can also be fixed in the other classes

alexander-ignatow (approved):
Good 👍 , just mind my comment for the future!

PR #2094, opened 2024-04-30
sonic

yurii-filiushkin:


src/figuresClasses.ts
```diff
@@ -1,11 +1,78 @@
...
 
-export class Triangle implements Figure {}
+export interface Figure {
+  shape: Shape;
+  color: Color;
+  getArea: () => number;
+}
+
+export class Triangle implements Figure {
+  shape: Shape = 'triangle';
```
alexander-ignatow, line 11:
```suggestion
  shape = 'triangle';
```

src/figuresClasses.ts
```diff
@@ -1,11 +1,78 @@
...
+    public color: Color,
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    if (
+      Math.min(a, b, c) <= 0 ||
+      a + b + c - Math.max(a, b, c) <= Math.max(a, b, c)
+    ) {
+      throw new Error('your error message');
```
alexander-ignatow, line 23: fix the error message pls. Be specific and describe the exact problem - that will help the users of your classes

src/figuresClasses.ts
```diff
@@ -1,11 +1,78 @@
...
+    }
+  }
+
+  getArea(): number {
+    const p = (this.a + this.b + this.c) / 2; // semiperimetr
```
alexander-ignatow, line 28: variable name is not ok. If you fix it you won't ned the comment
```suggestion
    const semiperimetr = (this.a + this.b + this.c) / 2;
```

src/figuresClasses.ts
```diff
@@ -1,11 +1,78 @@
...
+
+export class Circle implements Figure {
+  shape: Shape = 'circle';
+
+  constructor(
+    public color: Color,
+    public a: number,
+  ) {
+    if (this.a <= 0) {
+      throw new Error('My error message');
```
alexander-ignatow, line 43: This error message is also not ok

src/figuresClasses.ts
```diff
@@ -1,11 +1,78 @@
...
+export class Rectangle implements Figure {
+  shape: Shape = 'rectangle';
+
+  constructor(
+    public color: Color,
+    public a: number,
+    public b: number,
+  ) {
+    if (this.a <= 0 || this.b <= 0) {
+      throw new Error('My error message');
```
alexander-ignatow, line 63: error message

alexander-ignatow (changes requested):
Good, but some errors are needed to be fixed

src/figuresClasses.ts
yurii-filiushkin: all done except types. It give me error. How correct change it?

alexander-ignatow: If TS requires it - than sorry, probably my fault

alexander-ignatow (approved):
Good job!

Sorry if one of my comment was misleading

PR #2095, opened 2024-04-30
add task solution

OleksandraKoshyk:


src/figuresClasses.ts
```diff
@@ -1,11 +1,72 @@
-export interface Figure {}
+export interface Figure {
+  readonly shape: string;
+  readonly color: string;
+  getArea(): number;
+}
```
```suggestion
type Color = 'red' | 'green' | 'blue';
type Shape = 'triangle' | 'circle' | 'rectangle';

export interface Figure {
  shape: Shape;
  color: Color;
  getArea(): number;
}
```
specify types more precisely

etojeDenys (changes requested):

sTorba24 (approved):
GJ 🚀

PR #2098, opened 2024-05-02
add task solution

Follder:


src/figuresClasses.ts
```diff
@@ -1,11 +1,89 @@
-export interface Figure {}
+enum Shape {
+  Triangle = 'triangle',
+  Circle = 'circle',
+  Rectangle = 'rectangle',
+}
+
+type Color = 'red' | 'green' | 'blue';
+
```
alexander-ignatow, line 8: Why do you use enum and union type together?

Both work, but please be consistent and pick one.

src/figuresClasses.ts
```diff
@@ -1,11 +1,89 @@
...
+    this.c = c;
+
+    const isPositive: boolean = this.a > 0 && this.b > 0 && this.c > 0;
+    const isTriangle: boolean =
+      this.a + this.b > this.c &&
+      this.a + this.c > this.b &&
+      this.b + this.c > this.a;
+
+    if (!isPositive || !isTriangle) {
+      throw new Error('Invalid triangle');
```
alexander-ignatow, line 35: Please, be specific with the error message and tell what exactly went wrong

src/figuresClasses.ts
```diff
@@ -1,11 +1,89 @@
...
+
+  constructor(
+    public color: Color,
+    public radius: number,
+  ) {
+    this.radius = radius;
 
-export class Circle implements Figure {}
+    if (this.radius <= 0) {
+      throw new Error('Invalid circle');
```
alexander-ignatow, line 57: same as with triangle

src/figuresClasses.ts
```diff
@@ -1,11 +1,89 @@
...
+  constructor(
+    public color: Color,
+    public width: number,
+    public height: number,
+  ) {
+    this.width = width;
+    this.height = height;
+
+    if (this.width <= 0 || this.height <= 0) {
+      throw new Error('invalid rectangle');
```
alexander-ignatow, line 78: fix this message too

alexander-ignatow (changes requested):
Good job, but some issues are needed to be fixed

Follder:
> Good job, but some issues are needed to be fixed

Thanks! I`ll fix everything

alexander-ignatow (approved):
Looks good now 👍

PR #2105, opened 2024-05-06
task solution

mykhailonl:
There might be a dumb approach with rounding area of the circle, but in one specific test case it worked wrong with .toFixed(2), so I only found this way to make it work.


lerastarynets (approved):
lgtm 🚀

PR #2108, opened 2024-05-07
solutionV1

kgGugala:


src/figuresClasses.ts
```diff
@@ -1,11 +1,77 @@
...
+  color: Color;
+  getArea: () => number;
+}
+
+export class Triangle implements Figure {
+  shape: Shape = 'triangle';
+
+  constructor(
+    public color: Color,
+    public a: number,
```
yevhenii-pyl, line 15: If you have a meaningful variable 'color', why don't you name other variables like 'sideA', 'sideB' or something?
Just 'a', 'b', 'c' as variable names can quickly transform your code to some sort of a cryptography :)

yevhenii-pyl (approved):
Solid!
But don't neglect meaningful variable names :)

PR #2118, opened 2024-05-15
Develop

kravets111:


src/figuresClasses.ts
```diff
@@ -1,11 +1,73 @@
-export interface Figure {}
+export interface Figure {
+  shape: string;
```
vadiimvooo, line 2: you can define type Shape with such types
` type Shape = 'triangle' | 'circle' | 'rectangle'`

vadiimvooo: The same is for color

vadiimvooo (changes requested):
Almost done. Just one fix.

src/figuresClasses.ts
```diff
@@ -1,11 +1,76 @@
-export interface Figure {}
+type Shape = 'triangle' | 'circle' | 'rectangle';
+type Color = 'red' | 'green' | 'blue';
 
-export class Triangle implements Figure {}
+export interface Figure {
+  shape: string;
```
anastasiiavorobiova, line 5:
```suggestion
  shape: Shape;
```

src/figuresClasses.ts
```diff
@@ -1,11 +1,76 @@
...
+  color: string;
```
anastasiiavorobiova, line 6:
```suggestion
  color: Color;
```

anastasiiavorobiova (changes requested):
Great work! Almost done!

TarasHoliuk (approved):

PR #2119, opened 2024-05-15
Solution

Cy90sure:


src/figuresClasses.ts
```diff
@@ -1,11 +1,103 @@
-export interface Figure {}
+export interface Figure {
+  shape: 'triangle' | 'circle' | 'rectangle';
+  color: 'red' | 'green' | 'blue';
```
etojeDenys, line 3: better to move it to some types

src/figuresClasses.ts
```diff
@@ -1,11 +1,103 @@
...
+  getArea(): number;
+}
+
+export class Triangle implements Figure {
+  shape: 'triangle';
+
+  color: 'red' | 'green' | 'blue';
```
etojeDenys, line 10: and reuse here

src/figuresClasses.ts
```diff
@@ -1,11 +1,103 @@
...
+      throw new Error(
+        'the longest side of a triangle is >= than a sum of two others',
+      );
+    }
+
+    this.shape = 'triangle';
+    this.color = color;
+    this.a = a;
+    this.b = b;
+    this.c = c;
```
etojeDenys, line 38: we can simplify it using the [Typescript Constructor Shorthand](https://dev.to/satansdeer/typescript-constructor-shorthand-3ibd)

etojeDenys (changes requested):

volodymyr-soltys97 (approved):
Nice work 🔥

PR #2120, opened 2024-05-16
initial

karaedd:


src/figuresClasses.ts
```diff
@@ -1,11 +1,79 @@
-export interface Figure {}
+export interface Figure {
+  shape: string;
+  color: string;
```

src/figuresClasses.ts
```diff
@@ -1,11 +1,79 @@
-export interface Figure {}
+export interface Figure {
+  shape: string;
```

anastasiiavorobiova (changes requested):
Great work! Almost done!

vadiimvooo (approved):
Well done!)

PR #2140, opened 2024-05-24
created figures classes

ivanlavrinenko:


src/figuresClasses.ts
```diff
@@ -1,11 +1,98 @@
...
+    }
+
+    if (a + b <= c || a + c <= b || b + c <= a) {
+      throw new Error('Sides length cannot form a triangle');
+    }
+
+    this.color = color;
+    this.a = a;
+    this.b = b;
+    this.c = c;
```
vadiimvooo, line 28: you don't have to define this in such way 
https://kendaleiv.com/typescript-constructor-assignment-public-and-private-keywords/

vadiimvooo: The same for other classes

vadiimvooo (changes requested):
Good job

src/figuresClasses.ts
```diff
@@ -1,11 +1,98 @@
-export interface Figure {}
+interface Figure {
+  shape: string;
+  color: string;
```
vadiimvooo, line 3: add types for this props

PR #2144, opened 2024-05-28
Solution ts_figures_classes

BagietaCyganeta:


src/figuresClasses.ts
```diff
@@ -1,11 +1,103 @@
...
+
+  public c: number;
+
+  constructor(
+    // eslint-disable-next-line @typescript-eslint/default-param-last
+    color: 'red' | 'green' | 'blue',
+    a: number,
+    b: number,
+    c: number,
+  ) {
```
nataliaklonowska, line 24: When you declare a constructor parameter with an access modifier TypeScript implicitly creates a property on the class with the same name as the parameter and assigns the parameter value to that property
```suggestion
  constructor(
    // eslint-disable-next-line @typescript-eslint/default-param-last
    public color: 'red' | 'green' | 'blue',
    public a: number,
    public b: number,
    public  c: number,
  ) {
```

src/figuresClasses.ts
```diff
@@ -1,11 +1,103 @@
...
+export interface Figure {
+  shape: 'triangle' | 'circle' | 'rectangle';
+  color: 'red' | 'green' | 'blue';
+  getArea(): number;
+}
+
+export class Triangle implements Figure {
+  public shape: 'triangle';
+
+  public color: 'red' | 'green' | 'blue';
```
nataliaklonowska, line 10: to not repeat all this types just create type Shape and type Color

src/figuresClasses.ts
```diff
@@ -1,11 +1,103 @@
...
+      throw new Error(
+        'The longest side must be less than the sum of the other two sides',
+      );
+    }
+
+    this.shape = 'triangle';
+    this.color = color;
+    this.a = a;
+    this.b = b;
+    this.c = c;
```
nataliaklonowska, line 39: When you declare a constructor parameter with an access modifier TypeScript implicitly creates a property on the class with the same name as the parameter and assigns the parameter value to that property
```suggestion
```

src/figuresClasses.ts
```diff
@@ -1,11 +1,103 @@
...
+  }
+
+  getArea(): number {
+    const s = (this.a + this.b + this.c) / 2;
+    const area = Math.sqrt(s * (s - this.a) * (s - this.b) * (s - this.c));
```
nataliaklonowska, line 44: after fixing constructor you will not need "this" before a, b, c

nataliaklonowska (changes requested):

BagietaCyganeta:
@nataliaklonowska sorry but i don't understand what i need to do to fix my code

PR #2154, opened 2024-06-05
Solution

yuliaPel:


src/figuresClasses.ts
```diff
@@ -1,11 +1,75 @@
...
+    }
+  }
 
-export class Circle implements Figure {}
+  getArea(): number {
+    const area = (Math.PI * this.radius ** 2 * 100) / 100;
+
+    return areaRound(area);
+  }
+}
```
roman-mirzoian, line 52: There is an extra multiplication by 100 and division by 100, which can be removed.

```suggestion
  getArea(): number {
    const area = Math.PI * this.radius ** 2;
    return areaRound(area);
  }
}
```

roman-mirzoian (approved):
Well done!

PR #2169, opened 2024-06-20
Add figures classes

GoVolodya:
Pleas check the TS classes implementation.


lerastarynets (approved):
You're doing amazing, keep it up!

PR #2170, opened 2024-06-20
add solution for the TS_figures_classes task

lesdyman:
Hi there! Here is my solution for the TS_figures_classes task. Hope it is good enough. Cheers!

### DIFF
--- a/src/figuresClasses.ts
+++ b/src/figuresClasses.ts
@@ -1,11 +1,101 @@
-export interface Figure {}
+export interface Figure {
+  shape: 'triangle' | 'circle' | 'rectangle';
+  color: 'red' | 'green' | 'blue';
+  getArea(): number;
+}
 
-export class Triangle implements Figure {}
+export class Triangle implements Figure {
+  shape: 'triangle' = 'triangle';
 
-export class Circle implements Figure {}
+  color: 'red' | 'green' | 'blue';
 
-export class Rectangle implements Figure {}
+  private sideA: number;
 
-export function getInfo(figure): string {
-  return typeof figure;
+  private sideB: number;
+
+  private sideC: number;
+
+  constructor(
+    color: 'red' | 'green' | 'blue',
+    a: number,
+    b: number,
+    c: number,
+  ) {
+    if (a <= 0 || b <= 0 || c <= 0) {
+      throw new Error('Sides of triangle should be greater than 0');
+    }
+
+    if (a >= b + c || b >= a + c || c >= a + b) {
+      throw new Error('One of three sides is bigger than sum of two others');
+    }
+
+    this.color = color;
+    this.sideA = a;
+    this.sideB = b;
+    this.sideC = c;
+  }
+
+  getArea(): number {
+    const s = (this.sideA + this.sideB + this.sideC) / 2;
+    const area = Math.sqrt(
+      s * (s - this.sideA) * (s - this.sideB) * (s - this.sideC),
+    );
+
+    return Math.floor(area * 100) / 100;
+  }
+}
+
+export class Circle implements Figure {
+  shape: 'circle' = 'circle';
+
+  color: 'red' | 'green' | 'blue';
+
+  private radius: number;
+
+  constructor(color: 'red' | 'green' | 'blue', radius: number) {
+    if (radius <= 0) {
+      throw new Error('Radius should be greater than 0');
+    }
+
+    this.color = color;
+    this.radius = radius;
+  }
+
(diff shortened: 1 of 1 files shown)


src/figuresClasses.ts
```diff
@@ -1,11 +1,76 @@
-export interface Figure {}
+export interface Figure {
+  shape: 'triangle' | 'circle' | 'rectangle';
+  color: 'red' | 'green' | 'blue';
```
VenomMB, line 3: create type Shape =  'triangle' | 'circle' | 'rectangle' and type Collor = 'red' | 'green' | 'blue';

src/figuresClasses.ts
```diff
@@ -1,11 +1,76 @@
...
+  getArea(): number;
+}
+
+export class Triangle implements Figure {
+  shape: 'triangle' = 'triangle';
+
+  constructor(
+    public color: 'red' | 'green' | 'blue',
```
VenomMB, line 11: use type Color

src/figuresClasses.ts
```diff
@@ -1,11 +1,76 @@
...
+    return Math.floor(area * 100) / 100;
+  }
+}
+
+export class Circle implements Figure {
+  shape: 'circle' = 'circle';
 
-export class Triangle implements Figure {}
+  constructor(
+    public color: 'red' | 'green' | 'blue',
```
VenomMB, line 37: use type Color

src/figuresClasses.ts
```diff
@@ -1,11 +1,76 @@
...
+
+    return Math.floor(area * 100) / 100;
+  }
+}
+
+export class Rectangle implements Figure {
+  shape: 'rectangle' = 'rectangle';
+
+  constructor(
+    public color: 'red' | 'green' | 'blue',
```
VenomMB, line 56: use type Color

src/figuresClasses.ts
```diff
@@ -1,11 +1,76 @@
...
-export class Circle implements Figure {}
+  getArea(): number {
+    const area = Math.PI * (this.radius * this.radius);
+
+    return Math.floor(area * 100) / 100;
+  }
+}
+
+export class Rectangle implements Figure {
+  shape: 'rectangle' = 'rectangle';
```
VenomMB, line 53:
```suggestion
  shape: Shape = 'rectangle';
```

src/figuresClasses.ts
```diff
@@ -1,11 +1,76 @@
...
+      throw new Error('Sides of triangle should be greater than 0');
+    }
+
+    if (a >= b + c || b >= a + c || c >= a + b) {
+      throw new Error('One of three sides is bigger than sum of two others');
+    }
+  }
+
+  getArea(): number {
+    const s = (this.a + this.b + this.c) / 2;
```
VenomMB, line 26: Always try give clarify name for variables. 
```suggestion
    const semiPerimeter = (this.a + this.b + this.c) / 2;
```

src/figuresClasses.ts
```diff
@@ -1,11 +1,76 @@
...
+    const area = Math.sqrt(s * (s - this.a) * (s - this.b) * (s - this.c));
+
+    return Math.floor(area * 100) / 100;
+  }
+}
+
+export class Circle implements Figure {
+  shape: 'circle' = 'circle';
```
VenomMB, line 34:
```suggestion
  shape: Shape = 'circle';
```

src/figuresClasses.ts
```diff
@@ -1,11 +1,76 @@
...
+
+  constructor(
+    public color: 'red' | 'green' | 'blue',
+    private width: number,
+    private height: number,
+  ) {
+    if (width <= 0 || height <= 0) {
+      throw new Error('Sides of rectangle should be greater than 0');
+    }
+  }
```
VenomMB, line 63: you can create function like

function checkOnError(...args: number[]): void {
    throw new Error('Some side less 0');
  }
}

VenomMB (changes requested):
Good job, let's work on improving it now

lesdyman:
@VenomMB Thank you for your fair comments! I have used your advice to improve my solution. About the function you recommended me to write. It was a really good suggestion! But since the function should return different error messages in different places, I decided to take a slightly different approach. So, one function could do the checking and return different messages. Check it out! I'd love to hear your opinion on it)

VenomMB (approved):
Very well done. Your function is really good.
Well, my proposed function simply checked whether any parameter was zero, and you improved it

PR #2172, opened 2024-06-20
add solution

o-cherpak:


src/figuresClasses.ts
```diff
@@ -1,11 +1,78 @@
-export interface Figure {}
+export interface Figure {
+  shape: 'triangle' | 'circle' | 'rectangle';
+  color: 'red' | 'green' | 'blue';
 
-export class Triangle implements Figure {}
+  getArea(): number;
+}
```
etojeDenys, line 6:
```

enum Shape {
  Triangle = 'triangle',
  Circle = 'circle',
  Rectangle = 'rectangle',
}

export interface Figure {
  shape: Shape;
  color: Color;
  getArea(): number;
}
```

src/figuresClasses.ts
```diff
@@ -1,11 +1,78 @@
...
+
+export class Triangle implements Figure {
+  public shape: 'triangle';
+
+  constructor(
+    public color: 'red' | 'green' | 'blue',
```
etojeDenys, line 12:
```suggestion
    public color: Color,
```
and here you can just use Cololr type

src/figuresClasses.ts
```diff
@@ -1,11 +1,78 @@
...
+  }
+
+  getArea(): number {
+    return this.a * this.b;
+  }
+}
 
 export function getInfo(figure): string {
-  return typeof figure;
+  console.log(`A ${figure.color} ${figure.shape} - ${figure.getArea()}`);
```
etojeDenys, line 76: looks redundant

etojeDenys (changes requested):

etojeDenys (approved):
lgtm

PR #2174, opened 2024-06-21
add figures solution

FaiHamid:


src/figuresClasses.ts
```diff
@@ -1,11 +1,87 @@
...
+    public color: Colors,
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    if (a <= 0 || b <= 0 || c <= 0) {
+      throw new Error('Triangle side length must be greater than 0');
+    }
+
+    const biggestSide = [a, b, c].reduce((ac, cu) => (ac > cu ? ac : cu));
```
VenomMB, line 23:
```suggestion
    const biggestSide = Math.max(a, b, c);
```

src/figuresClasses.ts
```diff
@@ -1,11 +1,87 @@
...
 
-export class Circle implements Figure {}
+    const smallestSides = [a, b, c]
+      .filter((num) => num !== biggestSide)
+      .reduce((acc, curr) => curr + acc, 0);
```
VenomMB, line 27:
```suggestion
    const smallestSides = Math.min(...([a, b, c].filter((side) => side !== biggestSide)));
```

src/figuresClasses.ts
```diff
@@ -1,11 +1,87 @@
...
+    if (smallestSides <= biggestSide) {
+      throw new Error(
+        'The sum of the lengths of any two sides' +
+          'must be greater than the third side',
+      );
+    }
+  }
+
+  getArea(): number {
+    const semiperimeter = (this.a + this.b + this.c) / 2;
```
VenomMB, line 38:
```suggestion
    const semiPerimeter = (this.a + this.b + this.c) / 2;
```

src/figuresClasses.ts
```diff
@@ -1,11 +1,87 @@
...
+  }
+}
 
-export class Rectangle implements Figure {}
+export class Circle implements Figure {
+  public readonly shape = 'circle';
+
+  constructor(
+    public color: Colors,
+    public r: number,
```
VenomMB, line 55:
```suggestion
    public radius: number,
```

src/figuresClasses.ts
```diff
@@ -1,11 +1,87 @@
...
+  public readonly shape = 'rectangle';
+
+  constructor(
+    public color: Colors,
+    public a: number,
+    public b: number,
+  ) {
+    if (a <= 0 || b <= 0) {
+      throw new Error('Both rectangle sides must be greater than 0');
+    }
```
VenomMB, line 77: you can create function like
```
function checkOnError(errorMessage: string, ...args: number[]): void {
  }
}
```

VenomMB (commented):
Great job, let's work on improvements now

VenomMB (changes requested):
Great job, let's work on improvements now

src/figuresClasses.ts
FaiHamid: In this case, my const smallestSides should  contain the result of adding two smalest numbers, so Math.min choose just one of them (

roman-mirzoian (approved):
Well done!

PR #2175, opened 2024-06-21
Solution

demkoveronika:


src/figuresClasses.ts
```diff
@@ -1,11 +1,74 @@
-export interface Figure {}
+export interface Figure {
+  shape: string;
+  color: string;
```
VenomMB, line 3: create type Shape = 'triangle' | 'circle' | 'rectangle' and type Color = 'red' | 'green' | 'blue';

src/figuresClasses.ts
```diff
@@ -1,11 +1,74 @@
...
+  shape = 'triangle';
+
+  constructor(
+    public color: string,
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    if (a <= 0 || b <= 0 || c <= 0) {
+      throw new Error('throws an error');
```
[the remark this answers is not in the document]
VenomMB: Read

src/figuresClasses.ts
```diff
@@ -1,11 +1,74 @@
...
+  shape = 'rectangle';
+
+  constructor(
+    public color: string,
+    public width: number,
+    public height: number,
+  ) {
+    if (width <= 0 || height <= 0) {
+      throw new Error('throws an error');
+    }
```
VenomMB, line 62: you can create function like

```
function checkOnError(errorMessage: string, ...args: number[]): void {
  }
}
```

VenomMB (changes requested):
Good job, let's work on improvements

roman-mirzoian (approved):
Well done!

PR #2180, opened 2024-06-21
solution for typescript task

priemssonia:
niceday

### DIFF
--- a/src/figuresClasses.ts
+++ b/src/figuresClasses.ts
@@ -1,11 +1,70 @@
-export interface Figure {}
+type Shape = 'triangle' | 'circle' | 'rectangle';
+type Color = 'red' | 'green' | 'blue';
 
-export class Triangle implements Figure {}
+export interface Figure {
+  shape: Shape;
+  color: Color;
+  getArea(): number;
+}
 
-export class Circle implements Figure {}
+export class Triangle implements Figure {
+  constructor(
+    public color: Color,
+    public a: number,
+    public b: number,
+    public c: number,
+    public shape: Shape = 'triangle',
+  ) {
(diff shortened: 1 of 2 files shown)


src/figuresClasses.ts
```diff
@@ -1,11 +1,77 @@
...
+type Shape = 'triangle' | 'circle' | 'rectangle';
+type Color = 'red' | 'green' | 'blue';
 
-export class Triangle implements Figure {}
+export interface Figure {
+  shape: Shape;
+  color: Color;
+  getArea(): number;
+}
+
```
roman-mirzoian, line 9: Recommend creating a separate function like this:

```
function checkOnError(errorMessage: string, ...sides: number[]): void {
  }
}
```

because this check is used multiple times

src/figuresClasses.ts
```diff
@@ -1,11 +1,77 @@
...
+  public shape: Shape = 'circle';
 
-export class Rectangle implements Figure {}
+  constructor(
+    public color: Color,
+    public radius: number,
+  ) {
+    if (radius <= 0) {
+      throw new Error('Invalid radius: radius must be a positive number.');
+    }
```
roman-mirzoian, line 47:
```suggestion
   const circleErrorMessage = 'Invalid radius: radius must be a positive number.';
   
   checkOnError(circleErrorMessage, radius)
```
Also the message itself can also be put into a variable

roman-mirzoian (changes requested):

src/figuresClasses.ts
priemssonia: Thank you for review! I will definitely change the code based on your remarks:)

roman-mirzoian (approved):
Well done!

PR #2181, opened 2024-06-21
add task solution

V-Shut:


src/figuresClasses.ts
```diff
@@ -1,11 +1,76 @@
-export interface Figure {}
+export interface Figure {
+  shape: string;
+  color: string;
```

create 
and 

src/figuresClasses.ts
```diff
@@ -1,11 +1,76 @@
...
+
+  constructor(
+    public color: string,
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    if (a <= 0 || b <= 0 || c <= 0) {
+      throw new Error('Triangle side lengths must be greater than 0.');
+    }
```
roman-mirzoian, line 18: Recommend creating a separate function like this:

```
  }
}
```

src/figuresClasses.ts
```diff
@@ -1,11 +1,76 @@
...
+
+    if (a + b <= c || a + c <= b || b + c <= a) {
+      throw new Error(
+        `Triangle sides ${a}, ${b}, and ${c} can't form a valid triangle.`,
+      );
+    }
+  }
+
+  getArea(): number {
+    const s = (this.a + this.b + this.c) / 2;
```
roman-mirzoian, line 28: It is much better to use a more descriptive naming:

Example:
```suggestion
    const semiPerimeter = (this.a + this.b + this.c) / 2;
```

src/figuresClasses.ts
```diff
@@ -1,11 +1,76 @@
...
+  shape: string;
+  color: string;
+  getArea(): number;
+}
+
+export class Triangle implements Figure {
+  shape: string = 'triangle';
+
+  constructor(
+    public color: string,
```
roman-mirzoian, line 11:
```suggestion
    public color: Color,
```

src/figuresClasses.ts
```diff
@@ -1,11 +1,76 @@
...
+}
 
-export class Triangle implements Figure {}
+export class Circle implements Figure {
+  shape: string = 'circle';
 
-export class Circle implements Figure {}
+  constructor(
+    public color: string,
+    public radius: number,
```
roman-mirzoian, line 40:
```suggestion
    public color: Color,
    public radius: number,
```

roman-mirzoian (changes requested):

roman-mirzoian (approved):
Well done!

PR #2185, opened 2024-06-22
my solution

RuslanHnatyshyn:


src/figuresClasses.ts
```diff
@@ -1,11 +1,78 @@
...
-export class Circle implements Figure {}
+export class Circle implements Figure {
+  constructor(
+    public color: Color,
+    public radius: number,
+    public shape: Shape = 'circle',
+  ) {
+    if (radius <= 0) {
+      throw new Error('Radius must be greater than zero');
+    }
```
roman-mirzoian, line 46: Example of using error func:

```suggestion
    checkSides('Radius must be greater than zero', radius);
```
Also you can put error message separate in variable

roman-mirzoian (changes requested):

roman-mirzoian (approved):
Well done!

RuslanHnatyshyn:
Thank you for your comments.

PR #2186, opened 2024-06-23
add task solution

gazebo88fire:


src/figuresClasses.ts
```diff
@@ -1,11 +1,85 @@
...
+export class Triangle implements Figure {
+  shape: string = 'triangle';
+
+  constructor(
+    public color: string,
+    private sideA: number,
+    private sideB: number,
+    private sideC: number,
+  ) {
+    if (sideA <= 0 || sideB <= 0 || sideC <= 0) {
```
VenomMB, line 16: you can create function like
```
function checkOnError(errorMessage ,...args: number[]): void {
  }
}
```

VenomMB (changes requested):
Good job, let's work on improving it now

VenomMB (approved):
Good Job!!!

PR #2188, opened 2024-06-23
made figures classes

StsKushnir:


src/figuresClasses.ts
```diff
@@ -1,11 +1,85 @@
-export interface Figure {}
+export interface Figure {
+  shape: string;
+  color: string;
```
VenomMB, line 3:
```suggestion
  shape: Shape;
  color: Color;
```

src/figuresClasses.ts
```diff
@@ -1,11 +1,85 @@
...
+    const biggestSide = Math.max(a, b, c);
+    const sumOfSmaller = a + b + c - biggestSide;
+
+    if (biggestSide >= sumOfSmaller) {
+      throw new Error(`sides ${a}, ${b} and ${c} can not form a triangle`);
+    }
+  }
+
+  public getArea(): number {
+    const semiperimeter = 0.5 * (this.a + this.b + this.c);
```
VenomMB, line 32:
```suggestion
    const semiPerimeter = 0.5 * (this.a + this.b + this.c);
```

src/figuresClasses.ts
```diff
@@ -1,11 +1,85 @@
...
+  public shape: Shape = 'rectangle';
+
+  constructor(
+    public color: Color,
+    public width: number,
+    public height: number,
+  ) {
+    if (width <= 0 || height <= 0) {
+      throw new Error('all sides of rectangle must be greater than 0');
+    }
```
VenomMB, line 74: you can create function like

```
function checkOnError(errorMessage ,...args: number[]): void {
  }
}
```

VenomMB (approved):
Good job!!
Leave some comments

PR #2189, opened 2024-06-24
sol

sa4ok-1:


src/figuresClasses.ts
```diff
@@ -1,11 +1,79 @@
-export interface Figure {}
+export interface Figure {
+  shape: 'triangle' | 'circle' | 'rectangle';
+  color: 'red' | 'green' | 'blue';
```

create
and

src/figuresClasses.ts
```diff
@@ -1,11 +1,79 @@
...
+
+  constructor(
+    public color: 'red' | 'green' | 'blue',
+    public readonly a: number,
+    public readonly b: number,
+    public readonly c: number,
+  ) {
+    if (a <= 0 || b <= 0 || c <= 0) {
+      throw new Error('error length');
+    }
```
roman-mirzoian, line 18: Recommend creating a separate function like this:
```
  }
}
```

src/figuresClasses.ts
```diff
@@ -1,11 +1,79 @@
...
+  readonly shape: 'triangle' = 'triangle';
+
+  constructor(
+    public color: 'red' | 'green' | 'blue',
+    public readonly a: number,
+    public readonly b: number,
+    public readonly c: number,
+  ) {
+    if (a <= 0 || b <= 0 || c <= 0) {
+      throw new Error('error length');
```
roman-mirzoian, line 17: It is better to describe errors more informatively.
```suggestion
      throw new Error('All sides of the triangle must be greater than zero'');
```

src/figuresClasses.ts
```diff
@@ -1,11 +1,79 @@
...
+  shape: 'triangle' | 'circle' | 'rectangle';
+  color: 'red' | 'green' | 'blue';
+  getArea(): number;
+}
+
+export class Triangle implements Figure {
+  readonly shape: 'triangle' = 'triangle';
+
+  constructor(
+    public color: 'red' | 'green' | 'blue',
```
roman-mirzoian, line 11:
```suggestion
    public color: Color,
```

src/figuresClasses.ts
```diff
@@ -1,11 +1,79 @@
...
+
+    return Math.floor(area * 100) / 100;
+  }
+}
+
+export class Rectangle implements Figure {
+  readonly shape: 'rectangle' = 'rectangle';
+
+  constructor(
+    public color: 'red' | 'green' | 'blue',
```
roman-mirzoian, line 61:
```suggestion
    public color: Color,
```

src/figuresClasses.ts
```diff
@@ -1,11 +1,79 @@
...
+  }
+}
 
-export class Triangle implements Figure {}
+export class Circle implements Figure {
+  readonly shape: 'circle' = 'circle';
 
-export class Circle implements Figure {}
+  constructor(
+    public color: 'red' | 'green' | 'blue',
```
roman-mirzoian, line 42:
```suggestion
    public color: Color,
```

src/figuresClasses.ts
```diff
@@ -1,11 +1,79 @@
-export interface Figure {}
+export interface Figure {
+  shape: 'triangle' | 'circle' | 'rectangle';
+  color: 'red' | 'green' | 'blue';
+  getArea(): number;
+}
+
+export class Triangle implements Figure {
+  readonly shape: 'triangle' = 'triangle';
```
roman-mirzoian, line 8:
```suggestion
  readonly shape: Shape = 'triangle';
```

roman-mirzoian (changes requested):

PR #2194, opened 2024-06-25
Solution

mhryvko:


src/figuresClasses.ts
```diff
@@ -1,11 +1,77 @@
-export interface Figure {}
+export interface Figure {
+  shape: string;
+  color: string;
```

create
and

src/figuresClasses.ts
```diff
@@ -1,11 +1,77 @@
...
+
+export class Circle implements Figure {
+  constructor(
+    public color: string,
+    public radius: number,
+    public shape: string = 'circle',
+  ) {
+    if (radius <= 0) {
+      throw new Error('Circle radius must be a positive number');
+    }
```
roman-mirzoian, line 46: Recommend creating a separate function like this:
```
  }
}
```

src/figuresClasses.ts
```diff
@@ -1,11 +1,77 @@
...
+
+    if (a <= 0 || b <= 0 || c <= 0 || !isValidSides) {
+      throw new Error(
+        'Triangle sides must be positive and satisfy the triangle inequality',
+      );
+    }
+  }
+
+  getArea(): number {
+    const s = (this.a + this.b + this.c) / 2;
```
roman-mirzoian, line 30: It is much better to use a more descriptive naming:

Example:
```suggestion
    const semiPerimeter = (this.a + this.b + this.c) / 2;
```

src/figuresClasses.ts
```diff
@@ -1,11 +1,77 @@
...
+  }
+}
 
-export class Rectangle implements Figure {}
+export class Rectangle implements Figure {
+  constructor(
+    public color: string,
+    public width: number,
+    public height: number,
+    public shape: string = 'rectangle',
```
roman-mirzoian, line 61:
```suggestion
    public shape: Shape = 'rectangle',
```

src/figuresClasses.ts
```diff
@@ -1,11 +1,77 @@
...
+    const area = Math.PI * this.radius ** 2;
+
+    return areaRound(area);
+  }
+}
 
-export class Rectangle implements Figure {}
+export class Rectangle implements Figure {
+  constructor(
+    public color: string,
```
roman-mirzoian, line 58:
```suggestion
    public color: Color,
```

roman-mirzoian (changes requested):
Good job on commenting to improve the code.

PR #2199, opened 2024-06-29
'add solution'

om1gse:


src/figuresClasses.ts
```diff
@@ -1,11 +1,84 @@
...
+  shape: Shape;
+  color: Color;
+  getArea(): number;
+}
+
+function checkOnError(...args: number[]): void {
+  if (args.some((arg: number) => arg <= 0)) {
+    throw new Error('Sides must be greater than 0');
+  }
+}
```
roman-mirzoian, line 14: Let's make the error message more general and then you can customize it more specifically for each shape type.

Example:
```
  }
}
```

src/figuresClasses.ts
```diff
@@ -1,11 +1,84 @@
...
+export class Triangle implements Figure {
+  shape: Shape = 'triangle';
+
+  constructor(
+    public color: Color,
+    private a: number,
+    private b: number,
+    private c: number,
+  ) {
+    checkOnError(a, b, c);
```
roman-mirzoian, line 25: Example:
```suggestion
    checkOnError('All sides must be greater than 0', a, b, c);
```

src/figuresClasses.ts
```diff
@@ -1,11 +1,84 @@
...
 
-export class Circle implements Figure {}
+export class Circle implements Figure {
+  shape: Shape = 'circle';
+
+  constructor(
+    public color: Color,
+    private radius: number,
+  ) {
+    checkOnError(radius);
```
roman-mirzoian, line 52:
```suggestion
    checkOnError('Radius must be greater than 0', radius);
```

src/figuresClasses.ts
```diff
@@ -1,11 +1,84 @@
...
+
+export class Rectangle implements Figure {
+  shape: Shape = 'rectangle';
+
+  constructor(
+    public color: Color,
+    private width: number,
+    private height: number,
+  ) {
+    checkOnError(width, height);
```
roman-mirzoian, line 70:
```suggestion
    checkOnError('Width and height must be greater than 0', width, height);
```

roman-mirzoian (changes requested):

PR #2203, opened 2024-07-04
My solution

Karanelus:
Complited task

### DIFF
--- a/src/figuresClasses.ts
+++ b/src/figuresClasses.ts
@@ -1,11 +1,107 @@
-export interface Figure {}
+type Shape = 'triangle' | 'circle' | 'rectangle';
+type Color = 'red' | 'green' | 'blue';
 
-export class Triangle implements Figure {}
+export interface Figure {
+  shape: Shape;
+  color: Color;
+  getArea(): number;
+}
 
-export class Circle implements Figure {}
+export class Triangle implements Figure {
+  shape: Shape;
 
-export class Rectangle implements Figure {}
+  color: Color;
 
-export function getInfo(figure): string {
-  return typeof figure;
+  a: number;
+
+  b: number;
+
+  c: number;
+
+  constructor(color: Color, a: number, b: number, c: number) {
+    const triangleSmallerSides = [a, b, c]
+      .filter((num) => num !== Math.max(a, b, c))
+      .reduce((cur, next) => cur + next, 0);
+
+    this.shape = 'triangle';
+    this.color = color;
+
+    if (Math.max(a, b, c) >= triangleSmallerSides) {
+      throw new Error(
+        'Longer side cannot be longer or equal of summury of smallest sides',
+      );
+    } else if (a <= 0 || b <= 0 || c <= 0) {
+      throw new Error('Side cannot be 0');
+    } else {
+      this.a = a;
+      this.b = b;
+      this.c = c;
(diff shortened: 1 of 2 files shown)


src/figuresClasses.ts
```diff
@@ -1,11 +1,101 @@
...
+    public color: Color,
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    const triangleSmallerSides = [a, b, c]
+      .filter((num) => num !== Math.max(a, b, c))
+      .reduce((cur, next) => cur + next, 0);
+
+    this.shape = 'triangle';
```
Zibi95, line 23: This assignment can be outside of the `constructor` at line 11 you can assign value to shape. Also for those values of shape (triangle, circle etc.) you can use `enums`

src/figuresClasses.ts
```diff
@@ -1,11 +1,101 @@
...
+    this.color = color;
```
Zibi95, line 24: This assign is redundant as you are using typescript access modifiers in the parameter list.

src/figuresClasses.ts
```diff
@@ -1,11 +1,101 @@
...
+    if (Math.max(a, b, c) >= triangleSmallerSides) {
+      throw new Error(
+        'Longest side of a triangle must be less than sum of other two sides',
+      );
+    } else if (a <= 0 || b <= 0 || c <= 0) {
+      throw new Error('Triangle sides must be greater than zero');
+    } else {
+      this.a = a;
+      this.b = b;
+      this.c = c;
```
Zibi95, line 35: This assign is redundant as you are using typescript access modifiers in the parameter list.

src/figuresClasses.ts
```diff
@@ -1,11 +1,101 @@
...
+    }
+  }
+
+  getArea(): number {
+    const p: number = (this.a + this.b + this.c) / 2;
+    const area: number = Math.sqrt(
+      p * (p - this.a) * (p - this.b) * (p - this.c),
+    );
+
+    return Math.floor(area * 100) / 100;
```
Zibi95, line 45: You can create helper function for rounding the value to avoid repetition.

src/figuresClasses.ts
```diff
@@ -1,11 +1,101 @@
...
+    public radius: number,
+  ) {
+    this.shape = 'circle';
+    this.color = color;
 
-export class Rectangle implements Figure {}
+    if (radius <= 0) {
+      throw new Error('Circle radius must be greater than zero.');
+    } else {
+      this.radius = radius;
```
Zibi95, line 62: Same here, those assignments in constructor are redundant.

src/figuresClasses.ts
```diff
@@ -1,11 +1,101 @@
...
+    public height: number,
+  ) {
+    this.shape = 'rectangle';
+    this.color = color;
+
+    if (width <= 0 || height <= 0) {
+      throw new Error('Rectangle width and height must be greater than zero');
+    } else {
+      this.width = width;
+      this.height = height;
```
Zibi95, line 88: Here also.

Zibi95 (changes requested):
Good job! There are small things to fix and we are good to go.

natalia-klonowska (approved):

PR #2205, opened 2024-07-08
add figures classes

Fetrw:


lerastarynets (approved):
You're on the right track!

PR #2211, opened 2024-07-21
solution

akryto:


akryto:
has:
- a method `getArea` that returns the area of the figure rounded down to 
hundredths.

src/figuresClasses.ts
```diff
@@ -1,11 +1,80 @@
-export interface Figure {}
+export interface Figure {
+  shape: 'triangle' | 'circle' | 'rectangle';
```
VenomMB, line 2: create enum

src/figuresClasses.ts
```diff
@@ -1,11 +1,80 @@
...
+  color: 'red' | 'green' | 'blue';
```
VenomMB, line 3: create enum

VenomMB (approved):
Great job, leave small comments

PR #2213, opened 2024-07-24
Solution

viktoriyaRi:


src/figuresClasses.ts
```diff
@@ -1,11 +1,86 @@
-export interface Figure {}
+// figuresClasses.ts
 
-export class Triangle implements Figure {}
+export interface Figure {
+  shape: 'triangle' | 'circle' | 'rectangle';
+  color: 'red' | 'green' | 'blue';
+  getArea(): number;
+}
```
etojeDenys, line 7:
```suggestion
type Shape = 'triangle' | 'circle' | 'rectangle';
type Color = 'red' | 'green' | 'blue';

export interface Figure {
  shape: Shape;
  color: Color;
  getArea(): number;
}
```
we can simplify that and reuse those types everywhere

etojeDenys (changes requested):

etojeDenys (approved):
lgtm

PR #2220, opened 2024-07-29
add solution

iyehorova:


src/figuresClasses.ts
```diff
@@ -1,11 +1,84 @@
...
 
-export class Triangle implements Figure {}
+type Shape = 'triangle' | 'circle' | 'rectangle';
+type Color = 'red' | 'green' | 'blue';
 
-export class Circle implements Figure {}
+export interface Figure {
+  shape: Shape;
+  color: Color;
+  getArea: () => {};
```
DanilWeda, line 9: getArea return object? 
Please, explain this case

DanilWeda (changes requested):
Good job.
I am asking for changes with one small question.
As I see it, getArea returns a number, not an object.
Please correct this if I haven't missed anything)
Thank you very much)

DanilWeda (approved):
Good job. 
Approved. 
Many thanks)

PR #2222, opened 2024-07-30
add task solution

ashatskyy:


src/figuresClasses.ts
```diff
@@ -1,11 +1,112 @@
-export interface Figure {}
+export interface Figure {
+  color: string;
+  a?: number;
+  b?: number;
+  c?: number;
+  radius?: number;
+  width?: number;
+  height?: number;
```
etojeDenys, line 8: all these fields are redundant and don't forget to add shape here

src/figuresClasses.ts
```diff
@@ -1,11 +1,112 @@
-export interface Figure {}
+export interface Figure {
+  color: string;
```
etojeDenys, line 2: create separate type for color

src/figuresClasses.ts
```diff
@@ -1,11 +1,112 @@
...
+}
+
+export type FiguresData = {
+  shape: string;
+  color: string;
+  area: number;
+};
+
+export class Triangle implements Figure {
+  public shape = 'triangle';
```
etojeDenys, line 18: and separate type for shape

etojeDenys (changes requested):

src/figuresClasses.ts
ashatskyy: Why are they redundant? We use them all in classes

src/figuresClasses.ts
ashatskyy: Why create separate types for color and shape?

src/figuresClasses.ts
```diff
@@ -1,11 +1,110 @@
...
+    }
+  }
+
+  getArea(): number {
+    return +(Math.floor(this.width * this.height * 100) / 100);
+  }
+}
 
-export class Rectangle implements Figure {}
+export function getInfo(figure: FiguresData): string {
```
etojeDenys, line 106:
```suggestion
export function getInfo(figure: Figure): string {
```

src/figuresClasses.ts
```diff
@@ -1,11 +1,110 @@
...
+export interface Figure {
+  color: Color;
+  shape: Shape;
+}
+
+export type FiguresData = {
+  shape: Shape;
+  color: Color;
+  area: number;
+};
```
etojeDenys, line 13: looks redundant too

src/figuresClasses.ts
```diff
@@ -1,11 +1,110 @@
...
+
+export class Triangle implements Figure {
+  public shape: Shape = 'triangle';
+
+  public area = this.getArea();
```
etojeDenys, line 18: looks redundant

src/figuresClasses.ts
```diff
@@ -1,11 +1,110 @@
...
+  }
+}
 
-export class Rectangle implements Figure {}
+export function getInfo(figure: FiguresData): string {
+  const { shape, color, area } = figure;
 
-export function getInfo(figure): string {
-  return typeof figure;
+  return `A ${color} ${shape} - ${area}`;
```
etojeDenys, line 109: you can just calculate the area here

etojeDenys (approved):

PR #2230, opened 2024-08-02
Solution

RostyslavSharuiev:


src/figuresClasses.ts
```diff
@@ -1,11 +1,91 @@
...
+  shape = Shapes.TRIANGLE;
+
+  constructor(
+    public color: Colors,
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    if (this.a <= 0 || this.b <= 0 || this.c <= 0) {
+      throw new Error('Invalid data, the value cannot be negative');
```
VenomMB, line 29: better create separate file for error messages

src/figuresClasses.ts
```diff
@@ -1,11 +1,91 @@
...
+      this.a + this.c <= this.b ||
+      this.b + this.c <= this.a
+    ) {
+      throw new Error('Incorrect value(s) for triangle');
+    }
+  }
 
-export class Triangle implements Figure {}
+  getArea(): number {
+    const halfPerimeter = (this.a + this.b + this.c) / 2;
```
VenomMB, line 42:
```suggestion
    const halfPerimeter = (this.a + this.b + this.c) / 2;
```

src/figuresClasses.ts
```diff
@@ -1,11 +1,91 @@
...
+export class Rectangle implements Figure {
+  shape = Shapes.RECTANGLE;
 
-export class Rectangle implements Figure {}
+  constructor(
+    public color: Colors,
+    public width: number,
+    public height: number,
+  ) {
+    if (this.height <= 0 || this.width <= 0) {
```
VenomMB, line 79: you can create function like

```
  if (condition) {
  }
}
```

VenomMB (approved):
Great job!!

PR #2233, opened 2024-08-04
solved task

coldshein:


src/figuresClasses.ts
```diff
@@ -1,11 +1,98 @@
...
 
-export class Triangle implements Figure {}
+export interface Figure {
+  shape: Shape;
+  color: Color;
+  getArea(): number;
+}
+
+export class Triangle implements Figure {
+  shape: 'triangle' = 'triangle';
```
ArturErmolenko, line 11:
```suggestion
  shape: Shape = 'triangle';
```
there isn't such type as 'triangle' , plz fix it everywhere

src/figuresClasses.ts
```diff
@@ -1,11 +1,98 @@
...
+
+  color: Color;
+
+  a: number;
+
+  b: number;
+
+  c: number;
+
+  constructor(color: Color, a: number, b: number, c: number) {
```
ArturErmolenko, line 21: DRY, 
```suggestion
  constructor(color: Color, a: number, b: number, c: number) {
```

you can declare variables inside the constructor, fix it everywhere

ArturErmolenko (changes requested):
Good job 👍 
Let's improve a bit

BudnikOleksii (approved):
👍

PR #2237, opened 2024-08-05
ts_figures_classes solution

piyopiyo0:


src/figuresClasses.ts
```diff
@@ -1,11 +1,101 @@
-export interface Figure {}
+export interface Figure {
+  color: 'red' | 'green' | 'blue';
+  shape: 'triangle' | 'circle' | 'rectangle';
+  getArea(): number;
+}
+// Triangle Class
+export class Triangle implements Figure {
+  color: 'red' | 'green' | 'blue';
```
ArturErmolenko, line 8: you should declare type Color and use it everywhere

src/figuresClasses.ts
```diff
@@ -1,11 +1,101 @@
...
+
+  shape: 'triangle' = 'triangle';
```
ArturErmolenko, line 10: the same for shape , plz fix it everywhere

src/figuresClasses.ts
```diff
@@ -1,11 +1,101 @@
...
+    }
+
+    this.color = color;
+    this.a = a;
+    this.b = b;
+    this.c = c;
+  }
+
+  getArea(): number {
+    const s = (this.a + this.b + this.c) / 2;
```
ArturErmolenko, line 39: s ?

ArturErmolenko (changes requested):
GJ 👍 
Let's fix types

src/figuresClasses.ts
piyopiyo0: semiPerimeter))
thanks for your suggestions, all done 👍

src/figuresClasses.ts
```diff
@@ -1,11 +1,105 @@
...
+type Shape = 'triangle' | 'circle' | 'rectangle';
 
-export class Triangle implements Figure {}
+export interface Figure {
+  color: Color;
+  shape: Shape;
+  getArea(): number;
+}
+
+// Triangle Class
```
BudnikOleksii, line 10: Remove all comments

BudnikOleksii (approved):
👍

PR #2238, opened 2024-08-05
TS_Figures

victoriasokolvak:


src/figuresClasses.ts
```diff
@@ -1,11 +1,78 @@
-export interface Figure {}
+export interface Figure {
+  shape: 'triangle' | 'circle' | 'rectangle';
+  color: 'red' | 'green' | 'blue';
```
BudnikOleksii, line 3: Create types or enums for shape and color

src/figuresClasses.ts
```diff
@@ -1,11 +1,78 @@
...
+  getArea(): number;
+}
+
+export class Triangle implements Figure {
+  shape: 'triangle' = 'triangle';
+
+  constructor(
+    public color: 'red' | 'green' | 'blue',
```
BudnikOleksii, line 11: And use corresponding types here

src/figuresClasses.ts
```diff
@@ -1,11 +1,78 @@
...
+    }
+
+    const longestSide = Math.max(a, b, c);
+    const perimeter = a + b + c;
 
-export class Triangle implements Figure {}
+    if (longestSide >= perimeter - longestSide) {
+      throw new Error(
+        'The longest side must be shorter than the sum of the other two sides',
+      );
```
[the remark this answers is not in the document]
BudnikOleksii: > [CODE STYLE] - The error message here is clear, but consider including the actual values of the sides in the message to provide more context to the user.

Good point about error messages

src/figuresClasses.ts
```diff
@@ -1,11 +1,78 @@
...
+    }
+  }
+
+  getArea(): number {
+    const s = (this.a + this.b + this.c) / 2;
```
BudnikOleksii, line 31: Avoid one letter variable names

BudnikOleksii (changes requested):
Good work 👍 
Let's fix types and error messages

BudnikOleksii (approved):
Looks great 🔥

PR #2239, opened 2024-08-05
add task solution

YaroslavKolesnyk:


BudnikOleksii (changes requested):
Nice start 👍 
Let's fix types and error messages

src/figuresClasses.ts
```diff
@@ -1,11 +1,85 @@
...
+export type Color = 'red' | 'green' | 'blue' | 'yellow' | 'black' | 'white';
 
-export class Triangle implements Figure {}
+export type ShapeType = 'triangle' | 'circle' | 'rectangle';
 
-export class Circle implements Figure {}
+export interface Figure {
+  color: Color;
+  getArea(): number;
+}
```
BudnikOleksii, line 8: shape property is missing

src/figuresClasses.ts
```diff
@@ -1,11 +1,85 @@
-export interface Figure {}
+export type Color = 'red' | 'green' | 'blue' | 'yellow' | 'black' | 'white';
```
BudnikOleksii, line 1: In requirements the color should have only three possible values
image

BudnikOleksii (approved):
Approved, but check the requirements more carefully

PR #2243, opened 2024-08-05
task solution

olshum8:


src/figuresClasses.ts
```diff
@@ -1,11 +1,96 @@
-export interface Figure {}
+export interface Figure {
+  shape: string;
+  color: string;
```
```

export interface Figure {
  shape: Shape;
  color: Color;

  getArea(): number;
}
```
better to specify types more precisely

etojeDenys (changes requested):

etojeDenys (approved):
good job

PR #2247, opened 2024-08-06
try-one

vadimurbanovich:


src/figuresClasses.ts
```diff
@@ -1,11 +1,81 @@
...
+      throw new Error('Invalid triangle dimensions');
+    }
+  }
+
+  getArea = (): number => {
+    const { a, b, c } = this;
+
+    const area = (a + b + c) / 2;
+
+    return +Math.sqrt(area * (area - a) * (area - b) * (area - c)).toFixed(2);
```
VenomMB, line 33:
```suggestion
    return Number(Math.sqrt(area * (area - a) * (area - b) * (area - c)).toFixed(2));
```
It's more readable

src/figuresClasses.ts
```diff
@@ -1,11 +1,81 @@
-export interface Figure {}
+type Colors = 'red' | 'green' | 'blue';
+type Shapes = `triangle` | `circle` | `rectangle`;
```
VenomMB, line 2: better create enum

src/figuresClasses.ts
```diff
@@ -1,11 +1,81 @@
...
 
-export class Triangle implements Figure {}
+export interface Figure {
+  readonly shape: Shapes;
+  color: Colors;
+  getArea: () => number;
+}
+
+export class Triangle implements Figure {
+  readonly shape: Shapes = `triangle`;
```
VenomMB, line 11: and use enum

src/figuresClasses.ts
```diff
@@ -1,11 +1,81 @@
...
+
+  constructor(
+    public color: Colors,
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    if (a <= 0 || b <= 0 || c <= 0) {
```
VenomMB, line 19: you can create function like
```
  if (condition) {
  }
}
```

VenomMB (changes requested):
Good job, let's work on improvements now

BudnikOleksii (approved):
👍

PR #2248, opened 2024-08-06
implement Figure interface with Triangle, Circle, and Rectangle classes

igor-romasiuk:


src/figuresClasses.ts
```diff
@@ -1,11 +1,89 @@
-export interface Figure {}
+type ShapeType = 'Triangle' | 'Circle' | 'Rectangle';
+type ColorType = 'red' | 'green' | 'blue';
```
VenomMB, line 2:
```suggestion
type ShapeType = 'triangle' | 'circle' | 'rectangle';
type ColorType = 'red' | 'green' | 'blue';
```
better create enum

src/figuresClasses.ts
```diff
@@ -1,11 +1,89 @@
...
+export interface Figure {
+  shape: ShapeType;
+  color: ColorType;
 
-export class Circle implements Figure {}
+  getArea(): number;
+}
+
+export class Triangle implements Figure {
+  public shape: ShapeType = 'Triangle';
```
VenomMB, line 12: use enum

src/figuresClasses.ts
```diff
@@ -1,11 +1,89 @@
...
+
+  constructor(
+    public color: ColorType,
+    public sideA: number,
+    public sideB: number,
+    public sideC: number,
+  ) {
+    if (sideA <= 0 || sideB <= 0 || sideC <= 0) {
+      throw new Error('All sides must be greater than zero');
+    }
```
VenomMB, line 22: you can create function like

```
  if (condition) {
  }
}
```

src/figuresClasses.ts
```diff
@@ -1,11 +1,89 @@
...
+
+    if (
+      Math.max(sideA, sideB, sideC) >=
```
VenomMB, line 25: move to variable

src/figuresClasses.ts
```diff
@@ -1,11 +1,89 @@
...
+    ) {
+      throw new Error(
+        'The longest side must be less than the sum of the other sides',
+      );
+    }
+  }
+
+  getArea(): number {
+    const { sideA, sideB, sideC } = this;
+    const semiperimeter: number = (sideA + sideB + sideC) / 2;
```
VenomMB, line 36:
```suggestion
    const semiPerimeter: number = (sideA + sideB + sideC) / 2;
```

VenomMB (changes requested):
Great job, let's work on improvements now

BudnikOleksii (approved):
👍

PR #2249, opened 2024-08-06
done

NIKaragu:


src/figuresClasses.ts
```diff
@@ -1,11 +1,103 @@
...
+  constructor(
+    public color: Color,
+    public a: number,
+    public b: number,
+    public c: number,
+    readonly shape: Shape = Shape.Triangle,
+  ) {
+    if (this.sidesArray.some((side) => side <= 0)) {
+      throw new Error('length of some side is <= 0');
+    }
```
VenomMB, line 44: you can create function like

  if (condition) {
  }
}

VenomMB (changes requested):
Great job, leave comments

src/figuresClasses.test.ts
```diff
@@ -190,12 +190,12 @@ describe('Function "getInfo"', () => {
...
-    const redTriangle = new Triangle('red', 10, 12, 15);
+  // test('should return correct information about a triangle', () => {
+  //   const redTriangle = new Triangle('red', 10, 12, 15);
 
-    expect(getInfo(redTriangle))
-      .toBe('A red triangle - 59.81');
-  });
+  //   expect(getInfo(redTriangle))
+  //     .toBe('A red triangle - 59.81');
+  // });
```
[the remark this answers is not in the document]
NIKaragu: 'cause they don't work rightly in these cases. I implemented these tests (exactly with these parameters) and got absolutly expected result (result what tests expected). And I have no idea why it doesn't work. You can copy my code and try to execute it with the parameters as in the tests I commented)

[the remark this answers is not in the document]
NIKaragu: [you can try it there](https://www.typescriptlang.org/play/?#code/KYOwrgtgBAygFgQwA7CgbwFBSgFQE4CWCIA5gDaoC8UA5AC6HHnA0A0WUAwgXgMYVRqNXj34t22AErBedJgKF4Zc0hTYYAvhgyhIXAPZl9edB2kATQbSXn12AOJLQVmiScg7UMmCq0ARt7imtoEIHTAeABmCLyoAGIEJGBKptgAzogoAFywmcAA3By8hsY5nCV4hdgkwHQAgkoIOQAUAJSCAHxQ4BB+EYVaGAh+aQwxdFD8CGlpUAlJSgBC06iY6XJ0BLyTcDIA1gDqePqkMATmwGkA8pHzycDNkYn3ZxdpDXgIAJ45PX14AG0ALpQAA+3Ug-1aOQAbvpzqlsFACJEoM06F8UPpUU8FsBXpcPt9BJQhH8IjR2mskdgUWjcS9zoS8J8vlAADzUAAMVI4NOwdDgxwA7t1gKKAKIs4zNGiSBDmAhgWYEWbC46kNGfRXK5GzTlQHmUqr8rQ0wY0gD0lqgwtQiBhqDo+ig5hdRK+ADpVR7Hs8lAT3izvq0JEjrVB9IKIsLVagGQGmUHWZ60voIA9mmkmb9IRF2pQutmLhzue03Zd6LbjHs+cjUc069gPd7k98-XjAx72gAyHtNqAJ-FJltpjPNLM5iG9fOdKDF1AGnl13n8qCCkViyXSvCygAyoBIgsjqLHqAXetLhuNdbNUC0gymM1wjFUqGAAA9wiBzLM7ksVmRCAkAoDMwj-f1Vg4Ro3RAMg2QyZBgByeAkKsVCUE9fAiDfQoOCQY5wlkYBLAXNsfmnf5gSsAFBVVT0EFYdc4HovwmLotJPV4IE8OwAio2UEivBOGpRgJXMZxMagAU9WSONTEdgy+IFUxArYHjreSyJbUILg-G5mgAWQQQVPQgBAP2aWTPS0xTWVaUMOFaAEuR47Q+MIwTLCjXY8EDGBIAk-4rFst4WxsMBYkbGlmhiXggoiJiFwPUhBQSvACy6OKoAAannJkUqPOAw2wLkw1aXioH4ojwlI4AIAIAAFCICAzcI8HSqx0RYzijFIS46AJXLmPonyIn8yB2htAAmSrqq8vVAyOEScj8fRDGAYgQp6z0+tEwamSgDpqHksa-KTAKIEq4oQFGPBIudXc6yQMAAi2SYKjKCoSqq16yHeppKMS56-vevx0p+l63u2eKgbwcrEXSMAUF3Cq62w+RgC43ZeEODUSEDG5-weULmXsyraQbeTVSW-HV35Dd9FFEBxSgKVjietdsAjS5-rCABaRU0mGCh+ZZr9+b51BzI-SXQAHbAaEDAADWYYQQQJZgQFJ1REtFo2E-rRnyktsSgBB11fZhL2O5itottI9DNuhhRdM60laTwaTR80OAtKAanqRo2k66lsCUOhkhAXLjNMtIAEc8DoaKGZ2tJ6qalq2oiKAACoFf1tOM+awhs5MfmRs4hB2nzrnC-o9OGpL1rahziv5L8GuC+6hvi6z1vy8rrjHO9mz9ASD8SOaaaffvYIMCfWZuD4ARP2-X85kg5Z06AkD6tAOgILxREYJOeD5zyFC8nQvJPWXsRrpOO6HplEHoY+owOoMT-IdB7ZtSVGkCGTlEbzmRhENoJpsD3woNjfYy1ThJiJpBHunEAHKlnv7QODRNohzhqAm6xttabSsLHOAnpGoAEk85D3QbMXOucoCzXckiCOUcoBkM9JEIwMpiEW0YQARi5DyKANohFchNA+bQi8oDSFkJjKAa9QAb2JtvVArU95gUPpvY+1JT5wQQpfXIaFqAYSxnIlQzBH63QYC-TmSIob-Rhp9b+xhf7v1jOYNKcN3FOKgLsRIcA6DAOwPTJGKNIF1gsZjOBuMEEEyQbcFBtEdqeMFOxHaASSBBKBJglh2Dg7QnwWHKAbC8DR3klkoJND5JpLgJI+ekQwAgFkAQE4AdaiUJAJEfQHZ7g5GJkUu6oQSAn1qOw5WdQoAABI0BDi4hUDQMy5mQVTHkJZFdZnzIKbg1oGhlYDGkU-CYDAcLW2oCzUUGM3zNHKJ-T0jhgCgCYgAZiYgAFiYgAVlaEAA)

src/figuresClasses.ts
```diff
@@ -1,11 +1,120 @@
...
+      if (figureSidesArray <= 0) {
+        throw new Error('Radius is wrong (radius is <= 0)');
+      }
+    }
+
+    // we have to do Array.isArray(figureSidesArray),
+    // otherwise figureSidesArray.some((side: number) => side <= 0) does't work
+    if (
+      Array.isArray(figureSidesArray) &&
+      figureSidesArray.some((side: number) => side <= 0)
```
BudnikOleksii, line 31: - Remove all comments
- And this function has a few misconceptions. It shouldn't be related to the `FigureBase` as it doesn't know anything about sides, for example, the circle doesn't have any sides. The next moment you accept `figureSidesArray` but for some reason, an array could be just a number...

src/figuresClasses.ts
```diff
@@ -1,11 +1,120 @@
...
+  protected otherSidesSum: number = this.sidesArray.reduce(
+    (acc: number, sideLength: number) => acc + sideLength,
+    0,
+  );
 
-export class Triangle implements Figure {}
+  protected semiPerimeter: number = (this.longestSide + this.otherSidesSum) / 2;
 
-export class Circle implements Figure {}
+  protected isSidesWrong: boolean = this.longestSide >= this.otherSidesSum;
```
BudnikOleksii, line 54: These values are dynamic and should be calculated instead of setting them statically. For example, you can have methods to do these calculations

src/figuresClasses.ts
```diff
@@ -1,11 +1,120 @@
...
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    super();
+    Triangle.checkWrongSidesOfFigure(this.sidesArray);
+
+    if (this.isSidesWrong) {
+      throw new Error(
+        // eslint-disable-next-line max-len
```
BudnikOleksii, line 67: Don't disable linter

BudnikOleksii (changes requested):
Good work 👍 
Left a few comments, check them and update your solution

src/figuresClasses.ts
NIKaragu: Of course, geometry really impact on this code anyway, okay. Let's look at some facts:

The side of a figure is just a border of a figure, and a figure has as many sides as many angles between these sides figure has in its closed contour. So a circle has only 1 side in its contour and between 2 nearest points (between which theoreticaly lays other infinite quantity of points), laying on circle contour, angle = 0.

It was just only theory of 2D math (only mathematical model of circle, and mathematcal model !== real model), now let's consider it within our reality 😃 
In reality (and in whole world) the circle I described above (theoretical ideal circle) doesn't exist, and including creating accuracy of our technologies and sizes of atoms, the most accurate visible (without any gear) circle as possible will be a right polygon with more than 65000 angles and sides (and an angle between 2 nearest sides still tends to 0, but is not equal). I hope it was clear to understand something below...

**So what I wanted to say... Don't complicate** something due to some topic that is not related (and doesn't impact on) to a task (like geometry), really) 'cause if you delve into every topic not directly related to a task, you'll find the 'misconceptions' everywhere.

**And here are my comments for your comments:**
- FigureBase created only for comfort and shortness: it's really better to create 1 abstrct class with 1 static method inside, than write this method in every class... 'cause what if I have hundreds of figure? 
- In addition to "It shouldn't be related to the FigureBase as it doesn't know anything about sides" - it doesn't know about sides due to restriction of task rules (every class of figure must has requirable parametrs - corresponding number of sides to each figure), of course I could create a contructor in FigureBase with unlimited list of sides as a parameter and then operate with him inside every class... but I'm restricted by rules of task : )
- And sometimes I leave comments for clarity in code (like if I think other people may catch misunderstanding, reading my code, they can read why I did what I did).

About "The next moment you accept figureSidesArray but for some reason, an array could be just a number...
" - changed, but now it's not such flexible and comfortable in using as it was.

src/figuresClasses.ts
NIKaragu: These values are not dynamic, they are static for every other figure. And also I need to calculate and use it only 1 time.
By plan (according to the task) I don't have to call these vars anywhere else outside of the class (and because of that I don't need a separate method for every of that calcs), so I made it protected.
Also calculations are not so complex to move them into separate method (even static), if I do this I'll just made more code and it won't be more understandable.

ArturErmolenko (approved):

PR #2253, opened 2024-08-07
add solution

YarikKedrych:


src/figuresClasses.ts
```diff
@@ -1,11 +1,81 @@
-export interface Figure {}
+type Color = 'red' | 'green' | 'blue';
+type Shape = 'triangle' | 'circle' | 'rectangle';
 
-export class Triangle implements Figure {}
+export interface Figure {
+  shape: string;
```
BudnikOleksii, line 5: It's not a string, it should have `Shape` type

src/figuresClasses.ts
```diff
@@ -1,11 +1,81 @@
...
+
+    if (sides[2] >= sides[0] + sides[1]) {
+      throw new Error(
+        'The longest side must be less than the sum of the other two sides',
+      );
+    }
+  }
+
+  getArea(): number {
+    const s = (this.a + this.b + this.c) / 2;
```
BudnikOleksii, line 33: Avoid one-letter variable names

BudnikOleksii (approved):
Good work 👍 
Approved, but check the comments

PR #2254, opened 2024-08-07
task

NavI-Rap0:


src/figuresClasses.ts
```diff
@@ -1,11 +1,102 @@
-export interface Figure {}
+export interface Figure {
+  shape: 'triangle' | 'circle' | 'rectangle';
+  color: 'red' | 'green' | 'blue';
```
BudnikOleksii, line 3: Create types or enums for shape and color

src/figuresClasses.ts
```diff
@@ -1,11 +1,102 @@
...
+  getArea(): number;
+}
+
+export class Triangle implements Figure {
+  shape: 'triangle';
+
+  color: 'red' | 'green' | 'blue';
```
BudnikOleksii, line 10: And use them in corresponding places

src/figuresClasses.ts
```diff
@@ -1,11 +1,102 @@
...
+      throw new Error('Sides must be greater than 0');
+    }
+
+    const sides = [a, b, c].sort((x, y) => x - y);
+
+    if (sides[2] >= sides[0] + sides[1]) {
+      throw new Error('1 side cant be longer than 2 other');
+    }
+
+    this.shape = 'triangle';
```
BudnikOleksii, line 34: You don't need to set the shape in the constructor, you can do it in class directly

src/figuresClasses.ts
```diff
@@ -1,11 +1,102 @@
...
+export class Triangle implements Figure {
+  shape: 'triangle';
+
+  color: 'red' | 'green' | 'blue';
+
+  private a: number;
+
+  private b: number;
+
+  private c: number;
```
BudnikOleksii, line 16: And you don't to define them here

src/figuresClasses.ts
```diff
@@ -1,11 +1,102 @@
...
+    if (sides[2] >= sides[0] + sides[1]) {
+      throw new Error('1 side cant be longer than 2 other');
+    }
+
+    this.shape = 'triangle';
+    this.color = color;
+    this.a = a;
+    this.b = b;
+    this.c = c;
+  }
```
BudnikOleksii, line 39:
```suggestion
  constructor(
    private color: Color,
    private a: number,
    private b: number,
    private c: number,
  ) {
    if (a <= 0 || b <= 0 || c <= 0) {
      throw new Error('Sides must be greater than 0');
    }

    const sides = [a, b, c].sort((x, y) => x - y);

    if (sides[2] >= sides[0] + sides[1]) {
      throw new Error('1 side cant be longer than 2 other');
    }
  }
```
You can avoid setting values manually, typescript will do it for you if you specify properties types. Fix in all places

src/figuresClasses.ts
```diff
@@ -1,11 +1,102 @@
...
+
+  getArea(): number {
+    const s = (this.a + this.b + this.c) / 2;
+    const area = Math.sqrt(s * (s - this.a) * (s - this.b) * (s - this.c));
```
BudnikOleksii, line 43:
```suggestion
    const { a, b, c } = this;
    const semiperimeter = (a + b + c) / 2;
    const area = Math.sqrt(semiperimeter * (semiperimeter -a) * (semiperimeter - b) * (semiperimeter - c));
```
Avoid one-letter variable names, and you can use destructing

BudnikOleksii (changes requested):
Good start 👍 
Left some suggestions, check them

BudnikOleksii (approved):
Nicely done 🚀

PR #2255, opened 2024-08-08
task solution

SashaPas05:


src/figuresClasses.ts
```diff
@@ -1,11 +1,97 @@
...
+export class Circle implements Figure {
+  shape: Shape = Shape.Circle;
+
+  constructor(
+    public color: Color,
+    protected radius: number,
+  ) {
+    if (radius <= 0) {
+      throw new Error('Circle radius must be greater than 0.');
+    }
```
VenomMB, line 67: you can create function like

```
function checkOnError(...args: number[]): void {
    throw new Error('Some side less 0');
  }
}
```

VenomMB (approved):
Great job, let's work on improvements

PR #2270, opened 2024-08-18
task solution

tonni004:


src/figuresClasses.ts
```diff
@@ -1,11 +1,94 @@
-export interface Figure {}
+export interface Figure {
+  shape: string;
+  color: string;
```
```suggestion
type Color = 'red' | 'green' | 'blue';
type Shape = 'triangle' | 'circle' | 'rectangle';

export interface Figure {
  shape: Shape;
  color: Color;
```
we can specify types more precisely

etojeDenys (changes requested):

src/figuresClasses.ts
etojeDenys: image
still not fixed

etojeDenys (changes requested):

src/figuresClasses.ts
```diff
@@ -1,11 +1,97 @@
...
+      throw new Error(
+        `The longest side can't be greater than the sum of the other two sides`,
+      );
+    }
+
+    this.color = color;
+    this.a = a;
+    this.b = b;
+    this.c = c;
+  }
```
etojeDenys, line 36: [Typescript Constructor Shorthand](https://dev.to/satansdeer/typescript-constructor-shorthand-3ibd)

you can simplify logic using it

etojeDenys (changes requested):

src/figuresClasses.ts
```diff
@@ -1,11 +1,82 @@
...
+type Shape = 'triangle' | 'circle' | 'rectangle';
+type Color = 'red' | 'green' | 'blue';
 
-export class Triangle implements Figure {}
+export interface Figure {
+  shape: Shape;
+  color: Color;
+  a?: number;
+  b?: number;
+  c?: number;
```
etojeDenys, line 9: all classes need those fields? you can remove it
```suggestion

```

etojeDenys (approved):
lgtm

PR #2276, opened 2024-08-23
first try

Fugas252mm:
https://fugas252mm.github.io/ts_figures_classes/

### DIFF
--- a/src/figuresClasses.ts
+++ b/src/figuresClasses.ts
@@ -17,11 +17,11 @@ export class Triangle implements Figure {
     public shape: string = 'triangle',
   ) {
     if (a <= 0 || b <= 0 || c <= 0) {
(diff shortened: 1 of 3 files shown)


src/figuresClasses.ts
```diff
@@ -1,11 +1,70 @@
-export interface Figure {}
+export interface Figure {
+  shape: string;
```

src/figuresClasses.ts
```diff
@@ -1,11 +1,70 @@
...
+  color: string;
```
anastasiiavorobiova, line 3: Consider creating a type for the shape

anastasiiavorobiova (changes requested):
Great work! Almost done!

volodymyr-soltys97 (approved):
Good job! 👍

PR #2277, opened 2024-08-23
Figures

under-a:


src/figuresClasses.ts
```diff
@@ -1,11 +1,82 @@
-export interface Figure {}
+export interface Figure {
+  shape: 'triangle' | 'circle' | 'rectangle';
+  color: 'red' | 'green' | 'blue';
```
etojeDenys, line 3: create separate types for color and you can create an enum for shape

and reuse them everywhere

src/figuresClasses.ts
```diff
@@ -1,11 +1,82 @@
...
+  getArea(): number;
+}
+
+export class Triangle implements Figure {
+  public shape: 'triangle';
+
+  constructor(
+    public color: 'red' | 'green' | 'blue',
```
etojeDenys, line 11: here

etojeDenys (changes requested):

etojeDenys (approved):
lgtm

PR #2280, opened 2024-08-25
Solution

joaoguilherme000:


pedro-ruas (approved):
Very good João!
Just remember to use better error messages next time. Whenever we throw an error, it's better to say exactly why we are doing it. `error sides` doesn't say much `The gives sides don't match a Triangle` would be something a little more intuitive

PR #2281, opened 2024-08-25
task figure

MatiushechkoMark:


src/figuresClasses.ts
```diff
@@ -1,11 +1,72 @@
-export interface Figure {}
+export interface Figure {
+  shape: 'triangle' | 'circle' | 'rectangle';
+  color: 'red' | 'green' | 'blue';
```
etojeDenys, line 3: move those type to some custom types(`type Color = ......`)

src/figuresClasses.ts
```diff
@@ -1,11 +1,72 @@
...
+  getArea(): number;
+}
+
+export class Triangle implements Figure {
+  shape: 'triangle' = 'triangle';
+
+  constructor(
+    public color: 'red' | 'green' | 'blue',
```
etojeDenys, line 11: and reuse it everywhere

etojeDenys (changes requested):

etojeDenys (approved):
lgtm

PR #2283, opened 2024-09-02
done

v-zagorovskii:


src/figuresClasses.ts
```diff
@@ -1,11 +1,79 @@
-export interface Figure {}
+export interface Figure {
+  shape: string;
+  color: string;
+  getArea(): number;
+}
```
```suggestion
type Color = 'red' | 'green' | 'blue';
type Shape = 'triangle' | 'circle' | 'rectangle';

export interface Figure {
  shape: Shape;
  color: Color;
  getArea(): number;
}
```
you can specify types more precisely

etojeDenys (changes requested):

anastasiiavorobiova (approved):
Great work!

PR #2285, opened 2024-09-05
solution

DanielYurchenco:


src/figuresClasses.ts
```diff
@@ -1,11 +1,95 @@
-export interface Figure {}
+export interface Figure {
+  shape: string;
```
anastasiiavorobiova: Consider creating an enum for the shape

src/figuresClasses.ts
```diff
@@ -1,11 +1,95 @@
...
+
+  b: number;
+
+  c: number;
+
+  constructor(
+    public color: 'red' | 'green' | 'blue',
+    a: number,
+    b: number,
+    c: number,
```
anastasiiavorobiova:
```suggestion
    public a: number,
    public b: number,
    public c: number,
```

src/figuresClasses.ts
```diff
@@ -1,11 +1,95 @@
...
+export class Circle implements Figure {
+  shape = 'circle';
 
-export class Circle implements Figure {}
+  radius: number;
 
-export class Rectangle implements Figure {}
+  constructor(
+    public color: 'red' | 'green' | 'blue',
+    radius: number,
```
anastasiiavorobiova:
```suggestion
    public radius: number,
```

src/figuresClasses.ts
```diff
@@ -1,11 +1,95 @@
...
+  shape = 'rectangle';
+
+  width: number;
+
+  height: number;
+
+  constructor(
+    public color: 'red' | 'green' | 'blue',
+    width: number,
+    height: number,
```
anastasiiavorobiova:
```suggestion
    public width: number,
    public height: number,
```

anastasiiavorobiova (changes requested):
Great work! Almost done!

src/figuresClasses.ts
DanielYurchenco: sorry for the spam, I just can’t run the test in vs code, this is the only way I can check the code

src/figuresClasses.ts
DanielYurchenco: Regarding this, I originally wrote this way, but ai decided to correct me

src/figuresClasses.ts
```diff
@@ -1,11 +1,83 @@
-export interface Figure {}
+export enum Shape {
+  Triangle = 'triangle',
+  Circle = 'circle',
+  Rectangle = 'rectangle',
+}
+
+export interface Figure {
+  shape: Shape;
+  color: 'red' | 'green' | 'blue';
```
etojeDenys, line 9: better to move this to some type and reuse it everywhere

etojeDenys (approved):

PR #2288, opened 2024-09-06
task solution

YuliiaKosenchuk:


src/figuresClasses.ts
```diff
@@ -1,11 +1,108 @@
...
+
+  b: number;
+
+  c: number;
+
+  constructor(color: ColorType, a: number, b: number, c: number) {
+    super(ShapeEnum.TRIANGLE, color);
+    this.a = a;
+    this.b = b;
+    this.c = c;
```
etojeDenys, line 39: we can simplify it using the [Typescript Constructor Shorthand](https://dev.to/satansdeer/typescript-constructor-shorthand-3ibd)

etojeDenys (approved):

PR #2291, opened 2024-09-08
add task solution

daryna-kopytko:


src/figuresClasses.ts
```diff
@@ -1,11 +1,80 @@
-export interface Figure {}
+export interface Figure {
+  shape: string;
```
anastasiiavorobiova: Consider creating an enum for the shape

src/figuresClasses.ts
```diff
@@ -1,11 +1,80 @@
...
+  color: string;
```
anastasiiavorobiova: Consider creating a type for the color

anastasiiavorobiova (changes requested):
Great work! Almost done! Consider fixing error messages to provide information about an error without typos etc.

vadiimvooo (approved):
LGTM

PR #2298, opened 2024-09-13
Solution

Bajkowsky:


src/figuresClasses.ts
```diff
@@ -1,11 +1,85 @@
...
+      throw new Error(
+        'The side of the figure cannot be less than or equal to zero',
+      );
+    }
+
+    const maxLength: number = Math.max(a, b, c);
+    const totalLength: number = a + b + c;
+
+    if (maxLength >= totalLength - maxLength) {
+      throw new Error("Sides 1, 2, and 3 can't form a triangle");
```
natalia-klonowska, line 29: you're using this message from instruction but numbers 1, 2, 3 are used because those were the values given as sides. It would be better to substitute the values of sides as a, b, c
```suggestion
      throw new Error(`Sides ${a}, ${b}, and ${c} can't form a triangle`);
```
image

natalia-klonowska (changes requested):
Great work 😃 just fix error message to be more precise ;)

zibi-dev (approved):
Well done!

PR #2301, opened 2024-09-16
solution

Daryna-Kukharets:


src/figuresClasses.ts
```diff
@@ -1,11 +1,78 @@
-export interface Figure {}
+export interface Figure {
+  shape: string;
+  color: 'red' | 'green' | 'blue';
```
etojeDenys, line 3: better to create some type for Color and reuse that
```suggestion
  color: Color;
```

etojeDenys (approved):

PR #2306, opened 2024-09-21
add solution

YuriiMykhailenko:
🤷‍♂️

### DIFF
--- a/src/figuresClasses.ts
+++ b/src/figuresClasses.ts
@@ -1,5 +1,3 @@
-// import { error } from 'console';
-
 type Shape = 'triangle' | 'circle' | 'rectangle';
 type Color = 'red' | 'green' | 'blue';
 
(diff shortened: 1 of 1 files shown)


src/figuresClasses.ts
```diff
@@ -1,11 +1,86 @@
-export interface Figure {}
+// import { error } from 'console';
```
VolodymyrKirichenko: delete all comments if u don't need them

src/figuresClasses.ts
```diff
@@ -1,11 +1,86 @@
...
+
+    if (a <= 0 || b <= 0 || c <= 0 || sumOfThoSmallest <= theLongestSide) {
+      throw new Error(
+        'Each side ought be >0. And the longest ought be less than others sum',
+      );
+    }
+  }
+
+  getArea(): number {
+    const semiPerimeter: number = 0.5 * (this.a + this.b + this.c);
```
VolodymyrKirichenko: not an error, but will be more readable
```suggestion
    const semiPerimeter: number = (this.a + this.b + this.c) / 2;
```

VolodymyrKirichenko (approved):
Well done 🔥

PR #2315, opened 2024-09-26
Solution

v1RnT:


src/figuresClasses.ts
```diff
@@ -1,11 +1,91 @@
-export interface Figure {}
+/* eslint-disable @typescript-eslint/lines-between-class-members */
 
-export class Triangle implements Figure {}
+export interface Figure {
+  shape: string;
```
VolodymyrKirichenko, line 5: use a `union type` for it

src/figuresClasses.ts
```diff
@@ -1,11 +1,91 @@
...
+  color: 'red' | 'green' | 'blue';
```
VolodymyrKirichenko: u always make duplicates. Move it to a separate `union type' - `Color'
```suggestion
  color: Color;
```

src/figuresClasses.ts
```diff
@@ -1,11 +1,91 @@
...
+
+export class Triangle implements Figure {
+  shape = 'triangle';
+  color: 'red' | 'green' | 'blue';
+  private a: number;
+  private b: number;
+  private c: number;
+
+  constructor(
+    color: 'red' | 'green' | 'blue',
```
VolodymyrKirichenko: change it for all the cases
```suggestion
  color: Color;
  private a: number;
  private b: number;
  private c: number;

  constructor(
    color: Color,
```

src/figuresClasses.ts
```diff
@@ -1,11 +1,91 @@
...
+
+    if (a <= 0 || b <= 0 || c <= 0) {
+      throw new Error('Side lengths must be greater than 0');
+    }
+
+    const sides = [a, b, c].sort((x, z) => z - x);
+
+    if (sides[0] >= sides[1] + sides[2]) {
+      throw new Error('This triangle cannot exist');
+    }
```
VolodymyrKirichenko: sort???
u can use `Math.max` method for this logic
```suggestion
    const maxNum = Math.max(a, b, c); // max size
    const sum = a + b + c; // all the sides

    if (maxNum >= sum - maxNum) { // invalid triangle
      throw new Error('This triangle cannot exist');
    }
```

src/figuresClasses.ts
```diff
@@ -1,11 +1,91 @@
...
+    this.a = a;
+    this.b = b;
+    this.c = c;
```
VolodymyrKirichenko: why?
i think it's extra

src/figuresClasses.ts
```diff
@@ -1,11 +1,91 @@
...
+  }
+}
+
+export class Rectangle implements Figure {
+  shape = 'rectangle';
+  color: 'red' | 'green' | 'blue';
+  private width: number;
+  private height: number;
+
+  constructor(color: 'red' | 'green' | 'blue', width: number, height: number) {
```
VolodymyrKirichenko: for better readability
```suggestion
  constructor(
    color: Color,
    width: number, 
    height: number
  ) {
```

src/figuresClasses.ts
```diff
@@ -1,11 +1,91 @@
...
+    this.color = color;
+
+    if (width <= 0 || height <= 0) {
+      throw new Error('Width and hegiht must be greater than 0');
+    }
```
VolodymyrKirichenko, line 78: i see that u do it in constructor, but why?

VolodymyrKirichenko (changes requested):
Looks good, but let's make some improvements 👍

src/figuresClasses.ts
v1RnT: nope, they are necessary, i wouldn't be able to pass the variables `a`, `b` and `c` to other class methods, such as `getArea()`.

VolodymyrKirichenko: > nope, they are necessary, i wouldn't be able to pass the variables `a`, `b` and `c` to other class methods, such as `getArea()`.
Ohhh, yes, I messed up.
In order to access variables without - `this`, u need to initialize the public variable in the constructor.
Then you won't need to create it at the beginning and assign them at the end. 
Iєll show how it will look in a separate comment

src/figuresClasses.ts
```diff
@@ -1,11 +1,89 @@
...
+    const sumOfSides = a + b + c;
+
+    if (maxSide >= sumOfSides - maxSide) {
+      throw new Error('This triangle cannot exist');
+    }
+
+    this.a = a;
+    this.b = b;
+    this.c = c;
+  }
```
VolodymyrKirichenko: u can use next syntax in all ur cases. but it's just another way
```suggestion
  constructor(
   public color: Color, 
   public a: number,
   public b: number,
   public c: number
  ) {
    if (a <= 0 || b <= 0 || c <= 0) {
      throw new Error('Side lengths must be greater than 0');
    }

    const maxSide = Math.max(a, b, c);
    const sumOfSides = a + b + c;

    if (maxSide >= sumOfSides - maxSide) {
      throw new Error('This triangle cannot exist');
    }
  }
```

VolodymyrKirichenko (approved):
Good, but check my comments

PR #2316, opened 2024-09-26
Solution

KrivorotkoYurii:


src/figuresClasses.ts
```diff
@@ -1,11 +1,105 @@
...
+        'The longest side of a triangle is greater than a sum of two others',
+      );
+    }
+
+    this.color = color;
+
+    this.a = a;
+    this.b = b;
+    this.c = c;
+  }
```
VolodymyrKirichenko, line 38: u can use public to declare and initialize ur variables in the constructor. This will make ur code smaller
```suggestion
  shape: Shape = 'triangle';

  constructor(
    public color: Color,
    public a: number, 
    public b: number,
    public c: number
  ) {
    if (a <= 0 || b <= 0 || c <= 0) {
      throw new Error('Any length must not be less than 0');
    }

    if (a + b <= c || a + c <= b || b + c <= a) {
      throw new Error(
        'The longest side of a triangle is greater than a sum of two others',
      );
    }
  }
```

src/figuresClasses.ts
```diff
@@ -1,11 +1,105 @@
...
+
+  getArea(): number {
+    const perimeter = this.a + this.b + this.c;
+    const semiperimeter = perimeter / 2;
```
VolodymyrKirichenko, line 42: use camelCase for ur variables.
i think - 1 variable is extra)
```suggestion
    const semiPerimeter = this.a + this.b + this.c / 2;
```

src/figuresClasses.ts
```diff
@@ -1,11 +1,105 @@
...
+export class Triangle implements Figure {
+  shape: Shape = 'triangle';
+
+  color: Color;
+
+  a: number;
+
+  b: number;
+
+  c: number;
```
VolodymyrKirichenko, line 20: if u want left this variant - fix indents
```suggestion
  shape: Shape = 'triangle';
  color: Color;
  a: number;
  b: number;
  c: number;
```

VolodymyrKirichenko (approved):
Good job, but plz, review my comments 🥹

PR #2318, opened 2024-09-26
Add task solution

WatarJoy:


src/figuresClasses.ts
```diff
@@ -1,11 +1,101 @@
-export interface Figure {}
+export interface Figure {
+  shape: 'triangle' | 'circle' | 'rectangle';
+  color: `red` | `green` | `blue`;
 
-export class Triangle implements Figure {}
+  getArea(): number;
+}
+export class Triangle implements Figure {
+  shape: 'triangle';
```
VolodymyrKirichenko: 1) u can assign the value u want right away instead of doing it at the end
2) use the `union type` u created at the beginning
```suggestion
  shape: Shape;
```

src/figuresClasses.ts
```diff
@@ -1,11 +1,101 @@
-export interface Figure {}
+export interface Figure {
+  shape: 'triangle' | 'circle' | 'rectangle';
+  color: `red` | `green` | `blue`;
```
VolodymyrKirichenko: create `union types` for both properties
```suggestion
  shape: Shape;
  color: Color;
```

src/figuresClasses.ts
```diff
@@ -1,11 +1,101 @@
...
+export class Triangle implements Figure {
+  shape: 'triangle';
+
+  color: 'red' | 'green' | 'blue';
+
+  private a: number;
+
+  private b: number;
+
+  private c: number;
```
VolodymyrKirichenko: fix indents for better radability
```suggestion
  shape: 'triangle';
  color: 'red' | 'green' | 'blue';
  private a: number;
  private b: number;
  private c: number;
```

src/figuresClasses.ts
```diff
@@ -1,11 +1,101 @@
...
+    b: number,
+    c: number,
+  ) {
+    if (a <= 0 || b <= 0 || c <= 0) {
+      throw new Error('All sides must be greater than 0');
+    }
+
+    const sides = [a, b, c].sort((x, y) => y - x);
+
+    if (sides[0] >= sides[1] + sides[2]) {
```
VolodymyrKirichenko: why sort?
u can use `Math.max` and `Mathematic`  for this logic
```suggestion
    const maxSize = Math.max(a, b, c);
    const sum = ............

    if (maxSize >= sum + maxSize) {
```

src/figuresClasses.ts
```diff
@@ -1,11 +1,101 @@
...
+    }
+    this.shape = 'triangle';
+    this.color = color;
+    this.a = a;
+    this.b = b;
+    this.c = c;
+  }
+
+  getArea(): number {
+    const s = (this.a + this.b + this.c) / 2;
```
VolodymyrKirichenko: `s`? What the `s`? idk...
Change it

src/figuresClasses.ts
```diff
@@ -1,11 +1,101 @@
...
+
+    return Math.floor(area * 100) / 100;
+  }
+}
+
+export class Circle implements Figure {
+  shape: 'circle';
 
-export class Circle implements Figure {}
+  color: 'red' | 'green' | 'blue';
```
VolodymyrKirichenko: Use ur `union` type in all such cases!
```suggestion
  color: Color;
```

VolodymyrKirichenko (changes requested):
Good, but let's improve ur code! 😮‍💨

src/figuresClasses.ts
WatarJoy: It's a linter requirement, so it won't even pass the github tests.

src/figuresClasses.ts
```diff
@@ -1,11 +1,105 @@
...
+
+  color: Color;
+
+  private a: number;
+
+  private b: number;
+
+  private c: number;
+
+  constructor(color: Color, a: number, b: number, c: number) {
```
BudnikOleksii, line 21:
```suggestion
  constructor(private color: Color, private a: number, private b: number, private c: number) {
```
You can do it directly in constructor

src/figuresClasses.ts
```diff
@@ -1,11 +1,105 @@
...
+    if (maxSide >= sumOfOtherSides) {
+      throw new Error(
+        'The longest side must be less than the sum of the other two',
+      );
+    }
+    this.shape = 'triangle';
+    this.color = color;
+    this.a = a;
+    this.b = b;
+    this.c = c;
```
BudnikOleksii, line 38: And in this case you don't need to assign them manually

src/figuresClasses.ts
```diff
@@ -1,11 +1,105 @@
...
+  shape: Shape;
+  color: Color;
 
-export class Circle implements Figure {}
+  getArea(): number;
+}
+export class Triangle implements Figure {
+  shape: Shape;
+
+  color: Color;
```
BudnikOleksii, line 13:
```suggestion
  shape: Shape = 'triangle';
```

BudnikOleksii (approved):
Good work 👍 
Approved, but left a few comments

PR #2325, opened 2024-09-27
add task solution

dokrod:
solution

### DIFF
--- a/src/figuresClasses.ts
+++ b/src/figuresClasses.ts
@@ -1,11 +1,78 @@
-export interface Figure {}
+type Shape = 'triangle' | 'circle' | 'rectangle';
+type Color = 'red' | 'green' | 'blue';
 
-export class Triangle implements Figure {}
+export interface Figure {
+  shape: Shape;
+  color: Color;
+  getArea: () => number;
+}
 
-export class Circle implements Figure {}
+export class Triangle implements Figure {
+  public shape: 'triangle' = 'triangle';
 
-export class Rectangle implements Figure {}
+  constructor(
+    public color: Color,
+    private a: number,
+    private b: number,
+    private c: number,
+  ) {
+    if (this.a <= 0 || this.b <= 0 || this.c <= 0) {
+      throw new Error('Invalid data, the value cannot be negative');
+    }
 
-export function getInfo(figure): string {
-  return typeof figure;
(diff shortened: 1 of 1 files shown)


src/figuresClasses.ts
```diff
@@ -1,11 +1,78 @@
...
+    private c: number,
+  ) {
+    if (this.a <= 0 || this.b <= 0 || this.c <= 0) {
+      throw new Error('Invalid data, the value cannot be negative');
+    }
+
+    if (
+      this.a + this.b <= this.c ||
+      this.a + this.c <= this.b ||
+      this.b + this.c <= this.a
```
VolodymyrKirichenko, line 26: u used `private` for initialization. u can use these variables without `this`.
Change it in all such cases!
```suggestion
    if (a <= 0 || b <= 0 || c <= 0) {
      throw new Error('Invalid data, the value cannot be negative');
    }

    if (
      a + b <= c ||
      a + c <= b ||
      b + c <= a
```

src/figuresClasses.ts
```diff
@@ -1,11 +1,78 @@
...
 
-export class Triangle implements Figure {}
+export interface Figure {
+  shape: Shape;
+  color: Color;
+  getArea: () => number;
+}
+
+export class Triangle implements Figure {
+  public shape: 'triangle' = 'triangle';
```
VolodymyrKirichenko, line 11: u have `union type` for it. Use it in all such cases!
```suggestion
  public shape: Shape = 'triangle';
```

src/figuresClasses.ts
```diff
@@ -1,11 +1,78 @@
...
+      this.a + this.b <= this.c ||
+      this.a + this.c <= this.b ||
+      this.b + this.c <= this.a
+    ) {
+      throw new Error('Incorrect value(s) for triangle');
+    }
+  }
+
+  getArea = (): number => {
+    const s = (this.a + this.b + this.c) / 2;
```
VolodymyrKirichenko, line 33: `s`? What the `s`?
`Sandwich`? are u hungry and decided to have a snack right here?))
Change it)

VolodymyrKirichenko (changes requested):
Good job, but we can make some improvements 🥹 
Let's do it!

PR #2330, opened 2024-09-28
Solution

vikshiyan:


src/figuresClasses.ts
```diff
@@ -1,11 +1,84 @@
...
+  shape: Shape = 'triangle';
+
+  constructor(
+    public color: Color,
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    if (a <= 0 || b <= 0 || c <= 0) {
+      throw new Error('your error message');
```
VolodymyrKirichenko, line 19: Why u don't explain the error?
Do it in all such cases!
```suggestion
      throw new Error('...........');
```

src/figuresClasses.ts
```diff
@@ -1,11 +1,84 @@
...
+
+    if (a >= b + c || b >= a + c || c >= a + b) {
+      throw new Error('your error message');
+    }
+  }
+
+  getArea(): number {
+    const half = 1 / 2;
+    const perimetr = this.a + this.b + this.c;
+    const halfPerimeter = half * perimetr;
```
VolodymyrKirichenko, line 30: two variables are redundant
```suggestion
    const semiPerimeter = (this.a + this.b + this.c) / 2;
```

VolodymyrKirichenko (changes requested):
Well done! 🔥 
But let's make some fixes

PR #2331, opened 2024-09-29
Solution

Hradivnyk:


src/figuresClasses.ts
```diff
@@ -1,11 +1,83 @@
...
+    public color: Color,
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    const maxSide = Math.max(this.a, this.b, this.c);
+
+    if (
+      [this.a, this.b, this.c].some((item) => item <= 0) ||
+      maxSide >= this.a + this.b + this.c - maxSide
```
VolodymyrKirichenko: You do not need to manually assign constructor parameters to class properties when using access modifiers. TypeScript does this automatically. You can remove the assignments for 'color', 'a', 'b', and 'c' parameters.

```suggestion
    const maxSide = Math.max(a, b, c);

    if (
      [a, b, c].some((item) => item <= 0) ||
      maxSide >= a + b + c - maxSide
```

src/figuresClasses.ts
```diff
@@ -1,11 +1,83 @@
...
+    ) {
+      throw new Error('your error message');
```
VolodymyrKirichenko: Why u don't explain the error?
Do it for all such cases!
```suggestion
      throw new Error('...............');
```

src/figuresClasses.ts
```diff
@@ -1,11 +1,83 @@
...
+    }
+  }
+
+  getArea(): number {
+    const s = (this.a + this.b + this.c) / 2;
```
VolodymyrKirichenko: `s`? What the `s`?
Change it

src/figuresClasses.ts
```diff
@@ -1,11 +1,83 @@
-export interface Figure {}
+type Shape = 'triangle' | 'circle' | 'rectangle';
+type Color = 'red' | 'green' | 'blue';
 
-export class Triangle implements Figure {}
+function deleteRedundantNumber(numberWithRedundantNumber: number): number {
```
VolodymyrKirichenko: delete?
the function rounds, not deletes

VolodymyrKirichenko (changes requested):
Looks good, but let's make ur code more better! 🥹

BudnikOleksii (approved):
Looks good 🔥

PR #2333, opened 2024-09-30
Solution

Valeraom:


src/figuresClasses.ts
```diff
@@ -1,11 +1,119 @@
-export interface Figure {}
+type ShapeType = `triangle` | `circle` | `rectangle`;
+type ColorType = `red` | `green` | `blue`;
 
-export class Triangle implements Figure {}
+function checkSidesLength(...sides: number[]): boolean {
```
VolodymyrKirichenko, line 4: Why?
u can do it without the function.
I think this is redundant.
a triangle always has 3 sides, the `||` operator will enough

src/figuresClasses.ts
```diff
@@ -1,11 +1,119 @@
...
+}
+
+export class Triangle implements Figure {
+  constructor(
+    public color: ColorType,
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    const validSidesLength: boolean = checkSidesLength(this.a, this.b, this.c);
```
VolodymyrKirichenko, line 32: do u need `this` for access? In all the cases. Why?

src/figuresClasses.ts
```diff
@@ -1,11 +1,119 @@
...
+      case this.b:
+        validSides = maxSide < this.a + this.c;
+        break;
+
+      case this.c:
+        validSides = maxSide < this.a + this.b;
+        break;
+    }
+
+    if (!validSidesLength) {
```
VolodymyrKirichenko, line 50: ?
```suggestion
    if (a <= 0 || b <= 0 || c <= 0) {
```

src/figuresClasses.ts
```diff
@@ -1,11 +1,119 @@
...
+        break;
+
+      case this.b:
+        validSides = maxSide < this.a + this.c;
+        break;
+
+      case this.c:
+        validSides = maxSide < this.a + this.b;
+        break;
+    }
```
VolodymyrKirichenko, line 48: think about it)
```suggestion
    const maxSide = Math.max(a, b, c);
    const sum = a + b + c;
    
    if (sum - maxNum) {
      throw new Error('The longest side of a triangle is >= than a sum of two others');
    }
```

src/figuresClasses.ts
```diff
@@ -1,11 +1,119 @@
...
+    );
+
+    return roundToHundr(area);
+  }
+}
+
+export class Circle implements Figure {
+  constructor(
+    public color: ColorType,
+    public r: number,
```
VolodymyrKirichenko, line 77: what the `r`???
`Rabbit`?
`Robot`?
Change it

VolodymyrKirichenko (changes requested):
Looks good 👍 
But let's make some improvements

PR #2336, opened 2024-09-30
solution

coldmelnyk:


src/figuresClasses.ts
```diff
@@ -1,11 +1,86 @@
...
+  public shape: ValidShapes = 'triangle';
+
+  constructor(
+    public color: ValidColors,
+    public a: number,
+    public b: number,
+    public c: number,
+  ) {
+    const sides = [a, b, c];
+    const longestSide = Math.max(...sides);
```
VolodymyrKirichenko, line 20: this is redundant
```suggestion
    const longestSide = Math.max(a, b, c);
```

src/figuresClasses.ts
```diff
@@ -1,11 +1,86 @@
...
+    let sumOfOtherSides = 0;
+
+    for (const side of sides) {
+      if (side < longestSide) {
+        sumOfOtherSides += side;
+      }
+    }
```
VolodymyrKirichenko, line 27: do u need `for loop` for this?
```suggestion
    const sumOfOtherSides = (a + b + c) - longestSide;
```

src/figuresClasses.ts
```diff
@@ -1,11 +1,86 @@
...
+
+    if (a <= 0 || b <= 0 || c <= 0 || sumOfOtherSides <= longestSide) {
+      throw new Error('is not valid length!');
+    }
+  }
+
+  getArea(): number {
+    const p = (this.a + this.b + this.c) / 2;
```
VolodymyrKirichenko, line 35: `p`? What the `p`?
`party`? Do u want to make a party?)
Change it)

VolodymyrKirichenko (changes requested):
Well done, but let's make some improvements) 🥹

PR #2412, opened 2024-11-28
solution

KostiantynBond:


src/figuresClasses.ts
```diff
@@ -1,11 +1,97 @@
-export interface Figure {}
+export interface Figure {
+  shape: string;
```
BudnikOleksii, line 2: It's no a string, it should be `triangle, circle or rectangle`
image

src/figuresClasses.ts
```diff
@@ -1,11 +1,97 @@
...
+  color: 'red' | 'green' | 'blue';
```
BudnikOleksii, line 3: Create a type and use it

src/figuresClasses.ts
```diff
@@ -1,11 +1,97 @@
...
+  getArea(): number;
+}
+
+export class Triangle implements Figure {
+  shape = 'triangle';
+
+  color: 'red' | 'green' | 'blue';
```
BudnikOleksii, line 10: `shape` should have a corresponding `Shape` type, the same as color

src/figuresClasses.ts
```diff
@@ -1,11 +1,97 @@
...
+
+    if (sides[2] >= sides[0] + sides[1]) {
+      throw new Error(`Sides ${a}, ${b}, and ${c} can't form a triangle`);
+    }
+
+    this.color = color;
+    this.a = a;
+    this.b = b;
+    this.c = c;
+  }
```
BudnikOleksii, line 38:
```suggestion
  constructor(
    private color: Color,
    private a: number,
    private b: number,
    private c: number,
  ) {
    if (a <= 0 || b <= 0 || c <= 0) {
      throw new Error('Sides must be greater then 0');
    }

    const sides = [a, b, c].sort((x, y) => x - y);

    if (sides[2] >= sides[0] + sides[1]) {
      throw new Error(`Sides ${a}, ${b}, and ${c} can't form a triangle`);
    }
  }
```
You can declare properties directly in the constructor, in this case you don't need to assign them, typescript will do it for you

src/figuresClasses.ts
```diff
@@ -1,11 +1,97 @@
...
+
+  getArea(): number {
+    const s = (this.a + this.b + this.c) / 2;
```
BudnikOleksii, line 41: What is `s` try to avoid one-letter names

src/figuresClasses.ts
```diff
@@ -1,11 +1,97 @@
...
+export class Circle implements Figure {
+  shape = 'circle';
 
-export class Circle implements Figure {}
+  color: 'red' | 'green' | 'blue';
 
-export class Rectangle implements Figure {}
+  private radius: number;
+
+  constructor(color: 'red' | 'green' | 'blue', radius: number) {
```
BudnikOleksii, line 55: Fix type in all places, declarations as well

BudnikOleksii (changes requested):
Good work 👍 
Let's improve your solution

PR #2418, opened 2024-11-28
Solution

nataliia2211:


src/figuresClasses.ts
```diff
@@ -1,11 +1,75 @@
-export interface Figure {}
+export interface Figure {
+  shape: 'triangle' | 'circle' | 'rectangle';
+  color: 'red' | 'green' | 'blue';
```
BudnikOleksii: Create types for shape and color and use them

src/figuresClasses.ts
```diff
@@ -1,11 +1,75 @@
...
+      throw new Error('All sides must be positive numbers');
+    }
+
+    if (a + b <= c || a + c <= b || b + c <= a) {
+      throw new Error('Sides do not form a valid triangle');
+    }
+  }
+
+  getArea(): number {
+    const s = (this.a + this.b + this.c) / 2;
```
BudnikOleksii: What is `s` try to avoid one-letter names

BudnikOleksii (changes requested):
Good work 👍 
Let's fix types a bit

BudnikOleksii (approved):
LGTM 🚀

PR #2420, opened 2024-11-29
solution

Fridrif27:


src/figuresClasses.ts
```diff
@@ -1,11 +1,100 @@
-export interface Figure {}
+export interface Figure {
+  shape: 'triangle' | 'circle' | 'rectangle';
+  color: 'red' | 'green' | 'blue';
```
BudnikOleksii, line 3: Create types for shape and color and use them

src/figuresClasses.ts
```diff
@@ -1,11 +1,100 @@
...
+    if (a >= b + c || b >= a + c || c >= a + b) {
+      throw new Error('Sides cannot form a triangle');
+    }
+
+    this.shape = 'triangle';
+    this.color = color;
+    this.a = a;
+    this.b = b;
+    this.c = c;
+  }
```
BudnikOleksii, line 37:
```suggestion
  constructor(
    private color: Color,
    private a: number,
    private b: number,
    private c: number,
  ) {
    if (a <= 0 || b <= 0 || c <= 0) {
      throw new Error('Sides must be greater than 0');
    }

    if (a >= b + c || b >= a + c || c >= a + b) {
      throw new Error('Sides cannot form a triangle');
    }
  }
```
You can declare properties directly in the constructor, in this case you don't need to assign them, typescript will do it for you

Fix in all places

src/figuresClasses.ts
```diff
@@ -1,11 +1,100 @@
...
+
+  getArea(): number {
+    const s = (this.a + this.b + this.c) / 2;
```
BudnikOleksii, line 40: What is `s`? Try to avoid one-letter names

BudnikOleksii (changes requested):
Good work 👍 
Left a few comments to improve your solution

PR #2434, opened 2024-12-01
Task solution

AnderZerfall:


src/figuresClasses.ts
```diff
@@ -1,11 +1,98 @@
...
+
+  public color: Color;
+
+  private pointA: number;
+
+  private pointB: number;
+
+  private pointC: number;
+
+  constructor(color: Color, pointA: number, pointB: number, pointC: number) {
```
BudnikOleksii, line 20:
```suggestion
  public color: Color;

  constructor(private color: Color, private pointA: number,private  pointB: number, private pointC: number) {
```
You can add modifiers directly in constructor and in this case you don't have to assign values, typescript will do it for you

src/figuresClasses.ts
```diff
@@ -1,11 +1,98 @@
...
+      pointA >= pointB + pointC ||
+      pointB >= pointA + pointC ||
+      pointC >= pointB + pointA
+    ) {
+      throw new Error('all sides should be shorter than the sum of others');
+    }
+    this.color = color;
+    this.pointA = pointA;
+    this.pointB = pointB;
+    this.pointC = pointC;
```
BudnikOleksii, line 31:
```suggestion
```
Like this

BudnikOleksii (approved):
Good work 👍

PR #2446, opened 2024-12-07
add solution

Kuxim:


src/figuresClasses.ts
```diff
@@ -1,11 +1,86 @@
-export interface Figure {}
+export interface Figure {
+  shape: 'triangle' | 'circle' | 'rectangle';
+  color: 'red' | 'green' | 'blue';
```
etojeDenys, line 3: create some type Color
```suggestion
  color: Color;
```

src/figuresClasses.ts
```diff
@@ -1,11 +1,86 @@
...
+  getArea(): number;
+}
+
+export class Triangle implements Figure {
+  shape: 'triangle' = 'triangle';
+
+  color: 'red' | 'green' | 'blue';
```
etojeDenys, line 10: and reuse it here

etojeDenys (approved):

PR #2463, opened 2024-12-18
ADD new testes comparar Master

amandaga0189:
class Animal {
  public name: string;
 
  public constructor(theName: string) {
    this.name = theName;
  }
 
  public move(distanceInMeters: number) {
    console.log(`${this.name} moved ${distanceInMeters}m.`);
  }
}


amandaga0189 (commented):
Oi

PR #2481, opened 2024-12-31
Solution

djtyrf312:


src/figuresClasses.ts
```diff
@@ -1,11 +1,71 @@
-export interface Figure {}
+export interface Figure {
+  shape: string;
```
naziamoff, line 2: Shape should be an enum with predefined values, so I know I did something wrong when I passed `horse` there.

src/figuresClasses.ts
```diff
@@ -1,11 +1,71 @@
...
+      throw new Error('Triangle with such sides does not exist');
+    }
+  }
+
+  getArea(): number {
+    const p = (this.a + this.b + this.c) / 2;
+
+    return +Math.sqrt(p * (p - this.a) * (p - this.b) * (p - this.c)).toFixed(
+      2,
+    );
```
naziamoff, line 30: I would format it something like that. It's not the best approach, you can probably find better. But we want to have this construction readable to understand exactly what's happening here

```suggestion
    return +Math.sqrt(
      p * (p - this.a) * (p - this.b) * (p - this.c)
    ) .toFixed(2);
```

naziamoff (approved):

PR #2559, opened 2025-02-13
The shapes task has been completed

Dmytro-Bashlai:


src/figuresClasses.ts
```diff
@@ -1,11 +1,102 @@
...
+
+    return this.roundToHundredths(shapeArea);
+  }
+}
+
+export class Circle extends BaseFigure {
+  public shape: 'circle' = 'circle';
+
+  constructor(
+    color: 'red' | 'green' | 'blue',
```
etojeDenys, line 68:
```suggestion
    color: Color
```
it's better to move it to some type

etojeDenys (approved):

PR #2645, opened 2025-03-18
add-solution

George-Buzhenitsa:


src/figuresClasses.ts
```diff
@@ -1,11 +1,76 @@
-export interface Figure {}
+export interface Figure {
+  shape: string;
```
etojeDenys, line 2: you can create the custom type for shape and color
```suggestion
  shape: Shape;
```

src/figuresClasses.ts
```diff
@@ -1,11 +1,76 @@
...
+  color: string;
+  getArea: () => number;
+}
+
+export class Triangle implements Figure {
+  public shape: string = 'triangle';
```
etojeDenys, line 8:
```suggestion
  public shape: Shape = 'triangle';
```

etojeDenys (approved):

PR #2674, opened 2025-04-03
add task solution

Uusuff:


src/figuresClasses.ts
```diff
@@ -1,11 +1,75 @@
-export interface Figure {}
+export interface Figure {
+  shape: 'triangle' | 'circle' | 'rectangle';
+  color: 'red' | 'green' | 'blue';
```
volodymyr-soltys97: Create a enum for these values and use it everywhere

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code

maxim2310 (approved):
Well done!!!

PR #2897, opened 2025-08-07
implement Figures classes functionality

YuliaProkipchuk:


src/figuresClasses.ts
```diff
@@ -1,11 +1,76 @@
...
+  shape: 'triangle' | 'circle' | 'rectangle';
+  color: 'red' | 'blue' | 'green';
+  getArea(): number;
+}
+
+export class Triangle implements Figure {
+  shape: 'triangle' | 'circle' | 'rectangle';
+
+  constructor(
+    public color: 'red' | 'blue' | 'green',
```
ArturErmolenko: DRY(Don't repeat yourself) - you repeat shape color 4 times , this tells us that we need to put them into a separate type.

src/figuresClasses.ts
```diff
@@ -1,11 +1,76 @@
...
+    public b: number,
+    public c: number,
+  ) {
+    const max = Math.max(a, b, c);
+
+    if (a <= 0 || b <= 0 || c <= 0 || max >= b + c + a - max) {
+      throw new Error('Data is not valid');
+    }
+
+    this.shape = 'triangle';
```
ArturErmolenko: if you create enum for shape u can use it instead of magic string

src/figuresClasses.ts
```diff
@@ -1,11 +1,76 @@
...
+  constructor(
+    public color: 'red' | 'blue' | 'green',
+    public radius: number,
+  ) {
+    if (radius <= 0) {
+      throw new Error('Data is not valid');
+    }
 
-export class Circle implements Figure {}
+    this.shape = 'circle';
```
ArturErmolenko: and here)

src/figuresClasses.ts
```diff
@@ -1,11 +1,76 @@
...
+export class Circle implements Figure {
+  shape: 'triangle' | 'circle' | 'rectangle';
 
-export class Triangle implements Figure {}
+  constructor(
+    public color: 'red' | 'blue' | 'green',
+    public radius: number,
+  ) {
+    if (radius <= 0) {
+      throw new Error('Data is not valid');
```
ArturErmolenko: move error message in const or enum, don't rely on unknown string

ArturErmolenko (changes requested):

YuliaProkipchuk:
@ArturErmolenko Thank you for review! I've added type for color and enums for error message and shape

alexander-ignatow (approved):
Very good! 👍

PR #2901, opened 2025-08-10
add task solution

isapq:


DanielCamposDevX (changes requested):
image
Tests not passing + suggested AI changes. AI Aproved just so it doenst break your learning process but activity it still not 100%

PR #2904, opened 2025-08-11
added task solution

horidor:
finished


horidor:
@mateacademy-ai-mentor There is no requirement to have error messages - they just should throw errors if any length is a zero or negative value, or a special case for triangle.

PR #3140, opened 2026-01-12
add solution

bohdan8098:
Added solution


bohdan8098:
changed error messages

PR #3146, opened 2026-01-12
ts_figures_classes

SenyaKir:


src/figuresClasses.ts
```diff
@@ -1,11 +1,75 @@
...
+  shape: 'triangle' | 'circle' | 'rectangle';
+  color: 'red' | 'green' | 'blue';
+  getArea(): number;
+}
+
+export class Triangle implements Figure {
+  shape: 'triangle' = 'triangle';
+
+  constructor(
+    public color: 'red' | 'green' | 'blue',
```
dothearts: You can create a separate type (Shape, Color). 
Then reuse it in all places where you see duplication.

DanilWeda (changes requested):
All good! 
Please, remove type duplication. 
For example:
` color: 'red' | 'green' | 'blue';`
Move this into a separate type and replace it in all places. 
Many thanks!

DanilWeda (approved):
Approved.
Many thanks for the changes!

PR #3158, opened 2026-01-16
added solution

y-ivsnko:


y-ivsnko:
The task was not confirmed on MA platform

PR #3300, opened 2026-05-06
Solution

Blitzzry:


Blitzzry:
That is not fair, checklist item #3 just say that the function have to throw that error, but that does not say anything about a detailed explanation or something like that.

The constructors should throw new Error('your error message') if:

any length is <= 0
the longest side of a triangle is >= than a sum of two others

That is what it says, nothing else.

